PR #2, opened 2015-08-05
add domain id to node creation

dirk-thomas:
This is necessary for the recent changes in ros2/rmw#28


richiprosima:
Thanks for the contribution

PR #3, opened 2015-08-11
A few small changes to get it building on my system.

wjwwood:
One thing I had to do was disable the find_package(fastrtps) call.

This is because the CMake module for Fast-RTPS is incomplete:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake#L48

This results in a syntax error from CMake.
It happens to compile even without the find_package rule, so I disabled it for now.
However, it should really be finished and added back to this CMake logic.


richiprosima:
Thanks for the contribution

PR #9, opened 2015-12-18
Refactor for rmw_wait/rmw_waitset changes

jacquelinekay:
Connects to ros2/rmw#49

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -1,6 +1,7 @@
 #include "rmw/allocators.h"
 #include <rmw/rmw.h>
 #include <rmw/error_handling.h>
+#include <rmw/impl/cpp/macros.hpp>
 #include <rosidl_typesupport_introspection_cpp/identifier.hpp>
 #include <rmw_fastrtps_cpp/MessageTypeSupport.h>
 #include <rmw_fastrtps_cpp/ServiceTypeSupport.h>
@@ -23,6 +24,12 @@
 using namespace eprosima::fastrtps;
 
 class ClientListener;
+
+typedef struct CustomWaitsetInfo
+{
+    std::condition_variable condition;
+    std::mutex condition_mutex;
+} CustomWaitsetInfo;
 
 typedef struct CustomClientInfo
 {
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1474,43 +1591,57 @@ extern "C"
             rmw_guard_conditions_t *guard_conditions,
             rmw_services_t *services,
             rmw_clients_t *clients,
+            rmw_waitset_t * waitset,
             const rmw_time_t *wait_timeout)
     {
-        std::mutex conditionMutex;
-        std::condition_variable conditionVariable;
+        if (!waitset) {
+            return RMW_RET_ERROR;
```
dirk-thomas: When returning an error code it must set the error message before using `RMW_SET_ERROR_MSG`.

Same below.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1595,17 +1737,20 @@ extern "C"
...
-                guard_conditions->guard_conditions[i] = 0;
+                void *data = guard_conditions->guard_conditions[i];
+                GuardCondition *guard_condition = (GuardCondition*)data;
+                if(!guard_condition->getHasTriggered())
+                {
+                    guard_conditions->guard_conditions[i] = 0;
+                }
+          lock.unlock();
+                guard_condition->dettachCondition();
+          lock.lock();
```
dirk-thomas: Inconsistent indentation.

dirk-thomas:
Some of the new code has a different indentation then the existing code.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -865,6 +872,116 @@ extern "C"
...
+          rtps_guard_cond->attachCondition(&waitset_info->condition_mutex, &waitset_info->condition);
+        }
+      }
+
+      return waitset;
+
+    fail:
+      if (waitset) {
+        if (waitset->data) {
+          if (waitset_info) {
```
dirk-thomas: This check seems to be unnecessary since it is essentially the same as `waitset->data`.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -865,6 +872,116 @@ extern "C"
...
+      if (!waitset) {
+        RMW_SET_ERROR_MSG("failed to allocate waitset");
+        goto fail;
+      }
+      waitset->implementation_identifier = eprosima_fastrtps_identifier;
+      waitset->data = rmw_allocate(sizeof(CustomWaitsetInfo));
+      // This should default-construct the fields of CustomWaitsetInfo
+      waitset_info = static_cast<CustomWaitsetInfo *>(waitset->data);
+      RMW_TRY_PLACEMENT_NEW(waitset_info, waitset_info, goto fail, CustomWaitsetInfo);
+      if (!waitset->data || !waitset_info) {
```
dirk-thomas: Checking both is redundant - they can never have different values.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1474,43 +1591,57 @@ extern "C"
...
+        if (!waitset) {
+            return RMW_RET_ERROR;
+        }
+        CustomWaitsetInfo * waitset_info = static_cast<CustomWaitsetInfo*>(waitset->data);
+        if (!waitset_info) {
+            return RMW_RET_ERROR;
+        }
+        std::mutex * conditionMutex = &waitset_info->condition_mutex;
+        std::condition_variable * conditionVariable = &waitset_info->condition;
+        if (!conditionMutex || !conditionVariable) {
```
dirk-thomas: Separate checking for different error messages.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1474,43 +1591,57 @@ extern "C"
...
+            }
         }
+        rmw_guard_conditions_t * fixed_guard_conditions = waitset->fixed_guard_conditions;
 
-        std::unique_lock<std::mutex> lock(conditionMutex);
+        std::unique_lock<std::mutex> lock(*conditionMutex);
 
         // First check variables.
-        bool hasToWait = true;
+        bool hasToWait = (wait_timeout && wait_timeout->sec > 0 && wait_timeout->nsec > 0) || !wait_timeout;
```
dirk-thomas: The logic seems wrong. A timeout with `sec = 0` but `nsec > 0` would result in `false`?

jacquelinekay:
Yes, I'm not sure what to do about indentation in this repo. Running uncrustify on `functions.cpp` on master branch gives a lot of code style divergences. I can try to replicate the existing style, for example, 4 spaces of indentation seems consistent throughout the file. But there are some cases where the existing style looks weird to me, like where tabs are used instead of spaces...

dirk-thomas:
I would suggest to stay as close as possible to the style of the file. Mostly 4 space indentation. If you touch a line with tabs change it to spaces - otherwise keep it. eProsima does not use uncrustify (or anything else) so don't bother running it :wink:

richiprosima:
Not, we're not use any indention checker or anything. In all my development environments I have configured my Vim with 4 space indention. But my recently Ubuntu distro to develop ROS2 doesn't. Sorry. You are free to use anything you want.

jacquelinekay:
No worries! 4 spaces is fine, and it's better to keep the tab sizes consistent throughout the file anyway.

PR #10, opened 2015-12-30
Match alloc and free calls

gerkey:
This test result:

http://ci.ros2.org/view/nightly/job/nightly_linux/143/testReport/%28root%29/test_services_cpp__rmw_fastrtps_cpp/test_services_cpp/

shows a service client crashing with a SIGABRT (-6). valgrind identified mismatched allocation and destruction calls. Here I'm matching them up. With this change, valgrind gives a zero-error report on the service client (there still seem to be situations in which the test can hang instead of passing, but that's a different matter).

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -280,7 +280,8 @@ extern "C"
             return NULL;
         }
 
-        rmw_node_t * node_handle = rmw_node_allocate();
+        rmw_node_t * node_handle =
+            static_cast<rmw_node_t *>(malloc(sizeof(rmw_node_t)));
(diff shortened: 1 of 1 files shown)


wjwwood:
lgtm, I guess was `malloc` being used with `delete` and not `free`?

gerkey:
The `node->name` was being `malloc`ed, then `delete`d, which was definitely wrong. The `node` itself was being `rmw_node_allocate`d, then `delete`d, which could be right or wrong, depending on what happens in the allocate function. So I normalized them both to use `malloc` and `free`.

gerkey:
@wjwwood, if you like this change, can you merge it? (I don't have commit permissions on this repo.)

PR #12, opened 2016-01-19
fix missing export dependency

dirk-thomas:
`rmw` is listed as an exported dependency in CMake (https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/7631ef13594cdd7badd40d3d5c4d5d478ec9bd48/rmw_fastrtps_cpp/CMakeLists.txt#L16) but not listed in the package.xml.

Failing job: http://ci.ros2.org/job/ci_windows/933/consoleFull#console-section-66

Fixed in http://ci.ros2.org/job/ci_windows/935/


wjwwood:
I added a commit to make it one exported dependency per call and alphabetize them. I would have never noticed it as it is.

PR #15, opened 2016-02-10
add missing dep on introspection c

dirk-thomas:
Fix regression of #13 unbreaking the CI build.
Before: http://ci.ros2.org/job/ci_linux/919/ and http://ci.ros2.org/view/nightly/job/nightly_linux/185/
After: http://ci.ros2.org/job/ci_linux/924/


jacquelinekay:
I don't understand why we didn't see this problem until now, but the fix looks good to me.

+1

dirk-thomas:
I won't answer your question - I don't know either :wink: In isolation other packages should not be able to hide that...

dirk-thomas:
I was also not able to reproduce the failing build locally.

Issue #16, opened 2016-02-11
Reset fixed guard conditions


gerkey:
In ros2/rmw_connext#125, we found that the fixed guard conditions were never being reset to false, leading to `rmw_wait()` in many situations waking up immediately, every call. We fixed it in [rmw_connext](https://github.com/ros2/rmw_connext/pull/129) and [rmw_opensplice](https://github.com/ros2/rmw_opensplice/pull/111) by resetting the fixed guard conditions to false before existing `rmw_wait()`.

For FastRTPS, it looks like you'll want to repeat [this block](https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/functions.cpp#L1751-L1765) for the `fixed_guard_conditions`. I would submit a PR with that fix, but I'm not familiar enough with the code to know that that's the right answer.

dirk-thomas:
Fixed guard conditions have been removed some time ago (e.g. ros2/rmw_connext#147). Closing...

PR #18, opened 2016-03-04
Added support for introspection type support in C

esteve:
This adds support for rosidl_introspection_c, which allows being used with rcl (and thus rclpy).


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -896,10 +1168,12 @@ extern "C"
 
         waitset->fixed_guard_conditions = fixed_guard_conditions;
         if (fixed_guard_conditions) {
+/*
```
jacquelinekay: Any reason why this was commented? It seems like it should either be uncommented or deleted. It seems like important error checking to me...

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -2,28 +2,245 @@
...
+                    alignment = alignof(uint32_t);
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT64:
+                    alignment = alignof(int64_t);
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT64:
+                    alignment = alignof(uint64_t);
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+                    alignment = alignof(std::string);
```
jacquelinekay: I'm not sure this will work in the case of C strings. I haven't thought about this carefully but I suspect that `rosidl_generator_c__String` will have a different alignment from `std::string`.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -2,28 +2,245 @@
...
+                size = sizeof(uint32_t);
+                break;
+            case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT64:
+                size = sizeof(int64_t);
+                break;
+            case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT64:
+                size = sizeof(uint64_t);
+                break;
+            case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+                size = sizeof(std::string);
```
jacquelinekay: see above note about C strings.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -35,25 +252,555 @@ namespace rmw_fastrtps_cpp
...
+                    SER_ARRAY_SIZE_AND_VALUES(uint32_t)
+                        break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT64:
+                    SER_ARRAY_SIZE_AND_VALUES(int64_t)
+                        break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT64:
+                    SER_ARRAY_SIZE_AND_VALUES(uint64_t)
+                        break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+                    SER_ARRAY_SIZE_AND_VALUES(std::string)
```
jacquelinekay: You probably need logic here to select between `std::string` and `rosidl_generator_c__String`.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -35,25 +252,555 @@ namespace rmw_fastrtps_cpp
...
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT64:
+                    deser >> *(int64_t*)field;
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT64:
+                    deser >> *(uint64_t*)field;
+                    break;
+                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+                    {
+                        std::string &str = *(std::string*)field;
```
jacquelinekay: if using C typesupport, casting the field as `std::string` won't work.

esteve: Yeah, it shouldn't work, but it worked with the rclpy listener example, so I'm a bit puzzled :-P

jacquelinekay: With https://github.com/ros2/rclpy/pull/5 and all the necessary changes to stop ignoring fastrtps?

Do the rcl pub/sub with strings tests pass with fastrtps?

esteve: Yeah, I haven't checked the tests, but the examples do run.

jacquelinekay: I built rcl with this branch of rmw_fastrtps_cpp and removed the ignore logic in CMakeLists.txt for the tests, then ran `test_subscription__rmw_fastrtps_cpp` and got a segfault in the test case `test_subscription_nominal_string`:

```
0x00007ffff6d4b7e3 in std::string::swap(std::string&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff6d4b7e3 in std::string::swap(std::string&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff6d4b829 in std::string::operator=(std::string&&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007ffff6488eed in eprosima::fastcdr::Cdr::deserialize(std::string&) ()
   from /home/jackie/code/ros2_ws/install/lib/librmw_fastrtps_cpp.so
#3  0x00007ffff6488c09 in eprosima::fastcdr::Cdr::operator>>(std::string&) ()
   from /home/jackie/code/ros2_ws/install/lib/librmw_fastrtps_cpp.so
#4  0x00007ffff64958eb in rmw_fastrtps_cpp::TypeSupport<rosidl_typesupport_introspection_c__MessageMembers>::deserializeROSmessage(eprosima::fastcdr::Cdr&, rosidl_typesupport_introspection_c__MessageMembers const*, void*, bool)
    () from /home/jackie/code/ros2_ws/install/lib/librmw_fastrtps_cpp.so
#5  0x00007ffff6490155 in rmw_fastrtps_cpp::TypeSupport<rosidl_typesupport_introspection_c__MessageMembers>::deserializeROSmessage(rmw_fastrtps_cpp::Buffer const*, void*) ()
```

jacquelinekay: I can debug this if you like; I think I know what needs to happen with C strings/C++ strings based on what I did for Connext Dynamic.

esteve: Yeah, that sounds more like it :-) I still don't know why the examples work, maybe I was running the listener for another implementation by mistake. It'd be great if you can work on this, I added a templated struct that returns the appropriate type depending on whether we're using the C or the C++ structures, it might be useful in this case:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/18/files#diff-a8e9397d37ed8274594931575b5c83e1R211

esteve: This is how I used the templated struct to pick between C and C++ strings:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/18/files#diff-a8e9397d37ed8274594931575b5c83e1R529

rmw_fastrtps_cpp/src/functions.cpp
esteve: The check was too strict in my opinion. @richiprosima, what do you think of removing this check? In rclpy, we pass 0 as the number of guard conditions right now, which made this fail, is that a valid value?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
esteve: I think it's not needed to specialize for `rosidl_generator_c__String` because all strings end up being serialized/deserialized to/from `std::string`. In the case of `rosidl_generator_c__String`, they are first converted to `std::string` before they are passed to `eprosima::fastcdr::Cdr`

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
esteve: I think I fixed this, but I'm running the tests and see if they pass.

esteve:
The rcl tests show now pass, for example:

http://ci.ros2.org/job/ci_linux/1081/testReport/%28root%29/TestSubscriptionFixture__rmw_fastrtps_cpp/

but unfortunately the rclcpp tests seem to be broken:

http://ci.ros2.org/job/ci_linux/1081/testReport/junit/projectroot.home.rosbuild.ci_scripts.ws.src.ros2.system_tests/test_rclcpp/gtest_multiple_service_calls__rmw_fastrtps_cpp/

esteve:
The rclcpp tests now seems to pass:

http://ci.ros2.org/job/ci_linux/1083/

rmw_fastrtps_cpp/src/functions.cpp
richiprosima: @esteve, if you think it is too strict, go ahead and remove that check

dirk-thomas: This check should definitely not be remove. But currently the check is wrong. It is fine that `guard_conditions` is `0` if and only if `guard_condition_count` is zero.

jacquelinekay:
There is an uninitialized variable warning on OSX due to CppCheck (I think):
http://ci.ros2.org/job/ci_osx/877/testReport/junit/rmw_fastrtps_cpp/cppcheck/error__uninitvar__include_rmw_fastrtps_cpp_TypeSupport_h_326_/

jacquelinekay:
hey @esteve, I'm going to take this over if that's ok with you.

esteve:
@jacquelinekay thanks! I've updated this PR with your changes and also added a few notes about `StringHelper`, `std::string` and `rosidl_generator_c__String` that I kept meaning to add. I've reassigned this one to myself, since I need this for Android.

jacquelinekay:
that's odd, the CI jobs I launched and linked to in the last pull request have failures for the rcl publisher and subscription test, but when I run them locally they pass...

esteve:
Did you pass this repo in the repo url parameter to the Ci job?

Or maybe these is a change in master and when the CI Joe is triggered,
Jenkins rebases this against master and the tests fail. I can try rebasing
this against master, though this passes locally for me too.

esteve:
I think I know what's happening here, I submitted a PR to rcl with the same
name (typesupport_c) in https://github.com/ros2/rcl/pull/32 and maybe the
fixed conditions check is still too strict, so the test fails.

An earlier run of the test passed, but with the check disabled
http://ci.ros2.org/job/ci_linux/1084/

jacquelinekay:
The pull request in rcl with the same name is required to test this pull request.

The branch on master with the same name is my branch which is now up to par with this branch.

It's possible that I didn't rebase locally which is why the test passes for me but passes on master (I'll check now).

It's clear that we disagree about the guard conditions check. I think the error seen in CI might actually be due to issue #16. I will investigate this by resetting the fixed guard conditions and keeping the check, which I still think is necessary.

esteve:
> It's clear that we disagree about the guard conditions check.

Nope. I don't think we disagree, I believe the check is correct, but too
strict with the current codebase. If there's a way to keep the check intact
and have the tests passing, that'd be fantastic.

jacquelinekay:
I rebased my branch and built with Debug symbols, and it appears that the segfault has nothing to do with the issue I originally suggested, or with guard conditions:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000418d2c in TestPublisherFixture__rmw_fastrtps_cpp_test_publisher_nominal_Test::TestBody (
    this=0x97c9d0) at /home/jackie/code/ros2_ws/src/ros2/rcl/rcl/test/rcl/test_publisher.cpp:100
100	  EXPECT_EQ(strcmp(rcl_publisher_get_topic_name(&publisher), topic_name), 0);
```

jacquelinekay:
I fixed the test by adding the missing field (`topic_name`) to publishers and subscriptions, as well as fixing a few warnings, on my branch.

esteve:
I've included @jacquelinekay 's latest changes, rebased this branch and squashed the commits. @richiprosima have you had a moment to have a quick look at this and see if it looks good? If it does, it might be included in Alpha 5 and would bring FastRTPS up to speed with the other RMW implementations (e.g. support for rclpy). Thanks!

esteve:
Just rebased this again and resolved the conflicts.

tfoote:
@wjwwood to review

esteve:
@richiprosima sent me an email and told me that he was ok with the changes, so feel free to merge this PR.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/MessageTypeSupport.h
```diff
@@ -3,19 +3,41 @@
 
 #include "TypeSupport.h"
 #include "rosidl_typesupport_introspection_cpp/message_introspection.hpp"
+#include "rosidl_typesupport_introspection_cpp/field_types.hpp"
+
+#include <fastcdr/FastBuffer.h>
+#include <fastcdr/Cdr.h>
```
dirk-thomas: The package needs to export the FastCDR include directories in order to make them available to downstream packages.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -2,28 +2,270 @@
...
+#include "rosidl_typesupport_introspection_cpp/service_introspection.hpp"
+#include "rosidl_typesupport_introspection_cpp/visibility_control.h"
+
+#include "rosidl_typesupport_introspection_c/field_types.h"
+#include "rosidl_typesupport_introspection_c/identifier.h"
+#include "rosidl_typesupport_introspection_c/message_introspection.h"
+#include "rosidl_typesupport_introspection_c/service_introspection.h"
+#include "rosidl_typesupport_introspection_c/visibility_control.h"
+
+#define SER_ARRAY_SIZE_AND_VALUES(TYPE) \
```
dirk-thomas: These macros might need to be namespaced.

dirk-thomas:
It is very difficult to read the diff since it is unclear what code blocks have just been moved and if / which parts have been modified.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -634,15 +919,16 @@ extern "C"
         subscription = new rmw_subscription_t;
         subscription->implementation_identifier = eprosima_fastrtps_identifier;
         subscription->data = info;
+        subscription->topic_name = reinterpret_cast<const char *>(new char [strlen(topic_name) + 1]);
```
dirk-thomas: Why is the `reinterpret_cast` necessary here?

esteve:
I've created #27 with the contents of this PR so you guys can take over it. Closing this PR.

PR #19, opened 2016-03-17
Remove fixed guard conditions

jacquelinekay:
Connects to ros2/rcl#33

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -873,7 +873,7 @@ fail:
     }
 
     rmw_waitset_t *
-    rmw_create_waitset(rmw_guard_conditions_t * fixed_guard_conditions, size_t max_conditions)
+    rmw_create_waitset(size_t max_conditions)
     {
         rmw_waitset_t * waitset = rmw_waitset_allocate();
         GuardCondition * rtps_guard_cond = nullptr;
@@ -892,25 +892,6 @@ fail:
         if (!waitset_info) {
             RMW_SET_ERROR_MSG("failed to construct waitset info struct");
             goto fail;
-        }
-
-        waitset->fixed_guard_conditions = fixed_guard_conditions;
-        if (fixed_guard_conditions) {
-            if (!fixed_guard_conditions->guard_conditions) {
-                RMW_SET_ERROR_MSG("Received invalid guard condition array");
-                goto fail;
-            }
-            // Attach the fixed guard conditions to the waitset (and detach them in destroy)
-            for (size_t i = 0; i < fixed_guard_conditions->guard_condition_count; ++i) {
-                void * guard_cond = fixed_guard_conditions->guard_conditions[i];
-                if (!guard_cond) {
-                    RMW_SET_ERROR_MSG("Received invalid guard condition");
-                    goto fail;
-                }
-                rtps_guard_cond = static_cast<GuardCondition *>(guard_cond);
-                rtps_guard_cond->attachCondition(
-                    &waitset_info->condition_mutex, &waitset_info->condition);
-            }
         }
 
         return waitset;
@@ -949,25 +930,6 @@ fail:
         if (!conditionMutex) {
             RMW_SET_ERROR_MSG("waitset mutex is null");
             return RMW_RET_ERROR;
-        }
-
-        rmw_guard_conditions_t * fixed_guard_conditions = waitset->fixed_guard_conditions;
-        if (fixed_guard_conditions && fixed_guard_conditions->guard_conditions) {
-            // Attach the fixed guard conditions to the waitset (and detach them in destroy)
-            std::unique_lock<std::mutex> lock(*conditionMutex);
-            for (size_t i = 0; i < fixed_guard_conditions->guard_condition_count; ++i) {
-                void * guard_cond = fixed_guard_conditions->guard_conditions[i];
-                if (!guard_cond) {
-                    continue;
-                }
-                GuardCondition * rtps_guard_cond = static_cast<GuardCondition *>(guard_cond);
-                if (!rtps_guard_cond) {
-                    continue;
-                }
-                lock.unlock();
-                rtps_guard_cond->dettachCondition();
-                lock.lock();
-            }
         }
 
         if (waitset) {
@@ -1640,7 +1602,6 @@ fail:
                 guard_condition->attachCondition(conditionMutex, conditionVariable);
             }
         }
-        rmw_guard_conditions_t * fixed_guard_conditions = waitset->fixed_guard_conditions;
 
         std::unique_lock<std::mutex> lock(*conditionMutex);
 
@@ -1680,15 +1641,6 @@ fail:
                 void *data = guard_conditions->guard_conditions[i];
                 GuardCondition *guard_condition = (GuardCondition*)data;
                 if(guard_condition->hasTriggered())
-                    hasToWait = false;
-            }
-        }
-        if (fixed_guard_conditions) {
-            for (unsigned long i = 0; hasToWait && i < fixed_guard_conditions->guard_condition_count; ++i)
-            {
-                void *data = fixed_guard_conditions->guard_conditions[i];
-                GuardCondition *guard_condition = (GuardCondition*)data;
-                if(guard_condition->getHasTriggered())
                     hasToWait = false;
             }
         }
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1750,10 +1702,9 @@ extern "C"
             {
                 void *data = guard_conditions->guard_conditions[i];
                 GuardCondition *guard_condition = (GuardCondition*)data;
-                if(!guard_condition->getHasTriggered())
-                {
-                    guard_conditions->guard_conditions[i] = 0;
-                }
+                // Guard conditions persist (don't set to null)
+                // We call getHasTriggered to reset the trigger state.
```
dirk-thomas: Setting it to null was used to signal to the caller that this condition has "triggered". How is that being detected now?

jacquelinekay: @wjwwood wrote a rather long post about guard conditions: https://github.com/ros2/rcl/pull/33#issuecomment-198531721

I don't think that it's necessary to signal which guard condition triggered to the caller of wait. We use this pattern for subscriptions, services, and clients that get triggered so that the executor knows which callback to call, but guard conditions have no corresponding callback (they simply wake up the blocking call to wait).

wjwwood: It might be worth preserving the ability to detect which guard conditions were triggered during a wait (even if only for debugging).

jacquelinekay: ok. I confess I don't really like the "setting the handles to null" pattern in rmw/rclcpp because it requires managing a lot of boilerplate in the Executor, but I'm not going to spend time trying to make up a better pattern right now because I have so many pull requests in flight. It's also a bit hard to motivate keeping the boilerplate to manage the null handles in Executor for debugging purposes when there isn't an existing use case for it--I can see how this would be useful if we didn't have a completely monolithic system where we can edit and rebuild all the packages, but right now if I wanted to debug things I would just go directly into the RMW implementations... Anyway, if you both think this is worth preserving, I will fix up this pull request (and all the other rmw implementations) accordingly.

wjwwood: Just so we're on the same page, this is how I currently imagine things, maybe I am missing some part, so I'll describe it and perhaps you could correct me if I have it wrong? Currently you have to:

- Add guard conditions to the wait set once per executor (rather than once per call to rmw_wait)

And to preserve the ability to tell which guard condition was triggered you would need to:

- Add guard conditions to a list of guard conditions once per executor
- Add a pointer for each guard condition in the list once per call to rmw_wait

Is there other boilerplate that you need? If I'm right, you should no longer need to check which guard conditions were set to null each time in the executor (unless we add code to do debugging later). There is more work to be done in each rmw implementation (which we're talking about here), but I believe they are already setup to do that anyways.

The slight benefit to debugging, that I was thinking about, was being able to put a debug line in which printed a message, for example, when rmw_wait had been woken up by a node adding something new, rather than due to shutdown or ctrl-c. This would not be as convenient within the rmw implementations since they don't have the context about what each which guard condition they were given is used for, but the executor does.

Hopefully we can come to a consensus that this a valuable thing to preserve, rather than just giving in to the will of others. Honestly, if you think it's not a good idea then I'd said we do what you want since you're working on it actively at the moment.

As for the pattern of setting non-ready things to null, that's the pattern used by the `select` system call. My gut reaction is that it's probably the best solution because I've never seen a better one, but if you find one I'd love to hear about it. It is definitely an inconvenient interface, but I had written that off as one of the costs of doing this in C.

jacquelinekay: I agree with you that we no longer need to check which guard conditions were set to null in the executor.

By making the change that Dirk commented on that prompted this conversation, I am taking this one step further, e.g. because we no longer need to check it then we no longer need to set the guard condition handles to null.

I guess I would advocate actually implementing the debugging system you proposed if we are going to also keep setting the guard condition handles to null in the Executor, rather than keeping the functionality that I want to eliminate and not implementing the system that depends on it (the debugging system).

Does that make sense?

wjwwood: Yeah, that would be great, but unfortunately we don't have console logging yet :stuck_out_tongue:. And this is the kind of thing you want wrapped in `#ifndef NDEBUG` and on a "super debug level" so it doesn't print all the time. Just printing to stderr probably wouldn't be appropriate in this case. Otherwise I would have just suggested we implement the debugging now.

It shouldn't be difficult (and shouldn't take much effort) to re-add this functionality in the future if/when we have logging and need to debug it. So based on that and you're opinion, I'd say it's safe to remove it.

dirk-thomas: By not having a way to detect which conditions triggered the caller must *always* "handle" all of them. If in the event of a condition the processing is significant the caller has no way of doing that overhead conditionally. Therefore I think it makes sense to *keep* the ability to signal the triggered conditions.

wjwwood: > By not having a way to detect which conditions triggered the caller must always "handle" all of them. If in the event of a condition the processing is significant the caller has no way of doing that overhead conditionally. Therefore I think it makes sense to keep the ability to signal the triggered conditions.

I think @jacquelinekay's point is that this is another example of a solution in search of a problem. Without being able to point to a scenario where we need to do this it might not be worth keeping. Another way to look at it is that if we're not using it (even just in a test) then it's very likely it will be broken at some point and no one will notice.

dirk-thomas: Without this we are not able to check that our code is sleeping correctly. Simply because of that I think it is crucial to keep.

Second: the only reason why our executor has to refill all the handles every time before wait is because it simply doesn't use the information if the interrupt condition was triggered (in order to use new/changes handles). I think that is a good example why using this information is useful to avoid overhead like this.

wjwwood: > Without this we are not able to check that our code is sleeping correctly. Simply because of that I think it is crucial to keep.

That's the debugging argument, which without addressing the fact that we don't have a good way to do the debugging yet is to just reiterate what has already been argued.

My opinion is that we leave it until we need it or we have the tools to responsibly log it, but I can't honestly say that opinion is any more correct than @jacquelinekay's which is that if we're not using it we should get rid of it.

> Second: the only reason why our executor has to refill all the handles every time before wait is because it simply doesn't use the information if the interrupt condition was triggered (in order to use new/changes handles). I think that is a good example why using this information is useful to avoid overhead like this.

I don't understand what you mean here. As I understand it, the reason we have to refill the handles each time is because the rmw layer will replace pointers to the guard conditions with null anytime they're not set after waking up. Also that the set of guard conditions being passed may change from loop to loop (like when nodes are added or removed). So I think the simplest thing to do is to clear the guard conditions and add them all each time regardless of what has been waited on in the past and what was triggered or not last loop. This isn't an expensive set of operations, it's basically a memset to 0 for the waitset storage and then assigning the pointers into the waitset's storage. You might have to resize the waitset storage which is more expensive but hopefully infrequent.

I suppose you could maintain additional state and only selectively "reinsert" pointers into the waitset storage when entering a new loop, but I think that would be more computation than just reassigning them all each time. Also reassigning them all each time couldn't be avoided when you need to resize the waitset or when the set of things being waited on has changed. So I think it would add a lot of complexity for relatively little benefit.

wjwwood: At this point, I don't know if we've reached a consensus or not. My vote is to keep setting non-triggered guard conditions to null (and refilling them each loop) and just ignore which ones were triggered until we need that information for whatever reason. But I'm also fine if we take it out now.

@jacquelinekay I don't know if you just want to make a decision and run with it or wait until the next ROS 2 meeting.

wjwwood:
Aside from the discussion inline, this lgtm.

rmw_fastrtps_cpp/src/functions.cpp
jacquelinekay: I can wait until the next meeting. I think if I focus on the rcl/rclcpp refactor for a bit it will help me understand what the rcl_wait/rmw_wait and guard conditions part of the code should look like. Merging C services will also make working on this easier, since I'll only have to rebase on one set of changes instead of 2.

jacquelinekay: I thought about it some more and I decided that I agree with Dirk for now.

f8e81aa

I also think that this will work with my current set of changes in rclcpp just fine. Launching CI just in case (I should also probably launch a repeated job).

http://ci.ros2.org/job/ci_osx/895/

jacquelinekay: Oops, the new build failed because I forgot to rebase rcl, so I retriggered. http://ci.ros2.org/job/ci_osx/896/

PR #21, opened 2016-03-25
Explicit downcast of type size

codebot:
Avoid Windows compiler warning by making the 64->32 bit downcast explicit.

### DIFF
--- a/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
+++ b/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
@@ -21,7 +21,7 @@ MessageTypeSupport::MessageTypeSupport(const rosidl_typesupport_introspection_cp
     setName(strdup(name.c_str()));
 
     if(members->member_count_ != 0)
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
```diff
@@ -21,7 +21,7 @@ MessageTypeSupport::MessageTypeSupport(const rosidl_typesupport_introspection_cp
     setName(strdup(name.c_str()));
 
     if(members->member_count_ != 0)
-        m_typeSize = calculateMaxSerializedSize(members, 0);
+        m_typeSize = (uint32_t)calculateMaxSerializedSize(members, 0);
```
dirk-thomas: Use C++ cast here instead of C cast.

jacquelinekay:
`static_cast<uint32_t>(...)` is C++-style

jacquelinekay:
Nice, +1 with a squash

PR #22, opened 2016-03-25
Fix memory leak on Windows

codebot:
Unfortunately Windows throws a compiler warning when it sees strdup(); it wants to see _strdup() instead.

### DIFF
--- a/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
+++ b/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
@@ -18,7 +18,7 @@ MessageTypeSupport::MessageTypeSupport(const rosidl_typesupport_introspection_cp
         typeTooLarge_ = true;
 
     std::string name = std::string(members->package_name_) + "::msg::dds_::" + members->message_name_ + "_";
-    setName(strdup(name.c_str()));
+    setName(name.c_str());
 
     if(members->member_count_ != 0)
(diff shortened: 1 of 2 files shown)


tfoote:
Might it be better to use malloc and strncpy instead to avoid hidden mallocs: http://stackoverflow.com/a/14021507/604099 It's only one more line of code to explicitly malloc and copy.

jacquelinekay:
I haven't mucked around with this part of the codebase very much. @esteve would probably know since he worked a bunch with the TypeSupport to implement C typesupport. (Also, that reminds me--Esteve, you should probably watch the changes we're making to this repo because any changes to files that your pull request moves into a different file will essentially be lost if you don't rebase carefully.)

tfoote:
Yeah, it's just turning it back into a string. Passing the `.c_str()` to it should be adaquate. There are several other instances following that pattern. At least one in parameters and one in QoS.

dirk-thomas:
The same warning is in the sibling file `ServiceTypeSupport.cpp`: http://ci.ros2.org/job/ci_windows/1205/warnings34Result/source.6589365479438003664/#14

dirk-thomas:
Since `ServiceTypeSupport.cpp` has the exact same code with the exact same warnings as this `MessageTypeSupport.cpp` I would say they can be fixed together in this single PR.

wjwwood:
lgtm

dirk-thomas:
Can this be squashed and merged?

PR #24, opened 2016-04-05
export pthread dependency

dirk-thomas:
With only FastRTPS several examples fail to link since it seems to use `pthread` but doesn't export it (at least on Xenial, maybe also with other RMW impl. but currently I only have FastRTPS).

* http://ci.ros2.org/job/ci_linux/1159/
* http://ci.ros2.org/job/ci_osx/937/
* http://ci.ros2.org/job/ci_windows/1215/

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -37,6 +37,9 @@ ament_target_dependencies(rmw_fastrtps_cpp
 configure_rmw_library(rmw_fastrtps_cpp)
 
 ament_export_libraries(rmw_fastrtps_cpp fastcdr fastrtps)
+if(NOT WIN32)
+  ament_export_libraries(pthread)
+endif()
 
 register_rmw_implementation("rosidl_typesupport_introspection_cpp")
 


wjwwood:
I doubt this will work on Windows and maybe not ever on OS X or Linux with clang.

wjwwood:
Maybe CMake or our code (ignore if missing) is filtering it out? I don't think pthread is a valid library target on Windows.

dirk-thomas:
I wrapped it in a conditional for `NOT WIN32` (despite CI passing).

Since we use `pthread` in various other places on all non-Windows platforms I will keep that part as-is:
* https://github.com/ros2/rmw_connext/blob/6f1113f38be1cbb388311a2bcd0bc658edf29c8e/connext_cmake_module/cmake/Modules/FindConnext.cmake#L303
* https://github.com/ros2/rmw_opensplice/blob/d63fed94064e111b7dff6904f4475faf9f56c9bb/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake#L85

wjwwood:
Sounds good to me +1

jacquelinekay:
interesting, I wonder what happens with a Windows FastRTPS-only build.

jacquelinekay:
I confirmed on a Xenial VM that pthread linkage is an issue and that this fixes the build. +1

PR #27, opened 2016-04-26
Added support for type support in C

esteve:
This is a PR with the contents of #18 but within the eProsima repository, so anyone can contribute.


esteve:
I've rewritten the commit history a little bit and shuffled things around so to make this PR easier for review. The specific changes for adding support for typesupport in C are in this diff:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/compare/12a997d...typesupport_c

I've also created ros2/rosidl#128 to enable FastRTPS, I just tested all this with the rclpy examples and they seem to work locally.

@wjwwood I saw that @tfoote tagged you to review #18 (https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/18#issuecomment-214889872) let me know if you have any questions. @richiprosima  and @ros2_team feel free to make any changes you think necessary, thanks!

esteve:
Updated against master again, the specific changes for C typesupport are in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/compare/7315fe5...typesupport_c

jacquelinekay:
Testing this branch with rcl, I was getting segfaults in client/service instantiation. 9b2852f fixed it.

esteve:
@jacquelinekay thanks!

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_linux/1412/)
[Build Status](http://ci.ros2.org/job/ci_osx/1096/)
[Build Status](http://ci.ros2.org/job/ci_windows/1427/)

I will approve if this looks ok.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -86,11 +92,13 @@ static size_t calculateMaxAlign(const rosidl_typesupport_introspection_cpp::Mess
...
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+                    // StringHelper exposes rosidl_generator_c__String as std::string
+                    // so we don't need differentiate between C and C++ introspection typesupport
                     alignment = alignof(std::string);
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
                     {
-                        const ::rosidl_typesupport_introspection_cpp::MessageMembers *sub_members = (const ::rosidl_typesupport_introspection_cpp::MessageMembers*)member.members_->data;
+                        const MembersType *sub_members = (const MembersType*)member.members_->data;
```
dirk-thomas: No C casts in C++.

esteve: @dirk-thomas given that this repository is owned by eProsima (who may have their own codestyle guidelines, different from ROS2's), and that the changes are significant already, I tried to abstain from making further changes in order to help the reviewing process. However, if you think the code needs any changes, feel free to make them. Thanks!

jacquelinekay:
well, it looks like in the CI jobs I launched, most of the rclcpp tests using FastRTPS are broken.

@esteve I can look into this a bit.

esteve:
@jacquelinekay there was a null dereference in the timeout code that was causing `rmw_wait` to segfault (https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/27/commits/6965217845329b0ad161b337ce4a2a52905bfce8 fixes it). `rclcpp` tests now pass again (locally), except for `test_services_cpp__rmw_fastrtps_cpp` which timeouts (it timeouts in master as well).

I seem to have trouble to understand the timeout logic in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/typesupport_c/rmw_fastrtps_cpp/src/functions.cpp#L2064-L2127 Shouldn't it be something like the following?

```
        bool hasData = false;
        for(unsigned long i = 0; i < subscriptions->subscriber_count; ++i)
        {
            void *data = subscriptions->subscribers[i];
            CustomSubscriberInfo *custom_subscriber_info = (CustomSubscriberInfo*)data;
            hasData |= custom_subscriber_info->listener_->hasData();
            if(!custom_subscriber_info->listener_->hasData())
        ....
```

otherwise `hasData` would be false as soon as a subscriber, a client, etc. does not have data. The `test_services` test still timeouts with that change, though, but I wondered if the logic was correct.

In any case, with the fix for the null dereference this PR does not make fail any existing `rclcpp` test, at least on my local machine, so it's the same as master.

jacquelinekay:
Thanks for catching that issue; yeah, the logic I added to make the zero-wait timeout behavior consistent with other rmw implementations was quite incorrect.

[Build Status](http://ci.ros2.org/job/ci_linux/1425/)
[Build Status](http://ci.ros2.org/job/ci_osx/1109/)
[Build Status](http://ci.ros2.org/job/ci_windows/1441/)

jacquelinekay:
I've fixed the compiler warnings in dff2499 and 0515632.

There are some new Python test failures in that set of CI.

We don't have unit tests for C typesupport, but I strongly believe that this is because the typesupport code in this PR doesn't properly deserialize arrays. I will attempt to comment places in the code where I think this is wrong. I'll try to take care of it tomorrow.

jacquelinekay:
Btw, I use the "rcl" branch of system_tests for some standalone tests for C typesupport:

https://github.com/ros2/system_tests/tree/rcl

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -243,7 +255,7 @@ bool TypeSupport::serializeROSmessage(eprosima::fastcdr::Cdr &ser, const rosidl_
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
                     {
-                        std::string &str = *(std::string*)field;
+                        std::string && str = StringHelper<MembersType>::convert_to_std_string(field);
```
jacquelinekay: why a universal reference here?

jacquelinekay:
OK, looks like I can't comment on the code inline because the buggy functions are not in the diff of this PR.

The SER_ARRAY and DESER_ARRAY macros will lead to invalid casts from C types to C++ casts in C typesupport:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/27/files#diff-dfb392020946c298dc12127f8ef52c0dR195

This needs to be fixed before this PR is merged.

jacquelinekay:
Did a first pass at C-style array serialization in 2949f60 and 43b2aef. The tests I referred to earlier still don't pass though. Not sure about Python yet.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
esteve: This was a leftover from a bad rebase I think. Anyway, I've updated this to use the universal reference and `auto`, and `StringHelper::convert_to_std_string` to return a reference. The idea is that `StringHelper` does not create an extra copy of `std::string` for C++ typesupport, and by using `auto &&` it can be bound as an lvalue as well.

esteve:
@jacquelinekay thanks for looking into arrays, I didn't add support for them. I wrote this when the Python binding didn't support arrays and I've only tested this with single values. I just moved this PR to `in progress`

jacquelinekay:
I think I've fixed the issues with array types. I've started another set of jobs (this time with FastRTPS only, to save a bit of time).

[Build Status](http://ci.ros2.org/job/ci_linux/1434/)
[Build Status](http://ci.ros2.org/job/ci_osx/1118/)
[Build Status](http://ci.ros2.org/job/ci_windows/1451/)

jacquelinekay:
Looks like the linter errors are pre-existing, and the service failures are also pre-existing will be fixed once wait_for_service is implemented. There are some new Windows failures, however, in rcl, posibly due to changes in wait made in this PR.

jacquelinekay:
OK, the failure on Windows is in the "negative_timeout" case in `test_wait.c` in rcl:

http://ci.ros2.org/job/ci_windows/1452/testReport/(root)/WaitSetTestFixture__rmw_fastrtps_cpp/negative_timeout/

```
C:\J\workspace\ci_windows\ws\src\ros2\rcl\rcl\test\rcl\test_wait.cpp:107
Value of: ret
  Actual: 0
Expected: 2
error not set
```
For reference: https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_wait.cpp#L107

The test expects rcl_wait to time out, but it actually returns RCL_RET_OK.

My current theory is that this is because FastRTPS uses condition variables to implement guard conditions, which can wake up spuriously on Windows.

I will see if I can make a quick fix to the FastRTPS flavor of rmw_wait to make its behavior more consistent with the other implementations.

jacquelinekay:
well, I thought 	4ac24ed would fix the wait test on Windows but it didn't:

http://ci.ros2.org/job/ci_windows/1480/testReport/(root)/projectroot/test_wait__rmw_fastrtps_cpp/

@wjwwood We discussed this approach, does the way I implemented it look wrong to you?

wjwwood:
I think the commit looks ok. @jacquelinekay and I discussed the test failure offline and she's looking into the root cause.

jacquelinekay:
Alright, I expect this to work (no Opensplice): http://ci.ros2.org/job/ci_windows/1483/

jacquelinekay:
whoops, the Windows build failed spuriously while installing vcstool; trying again.

http://ci.ros2.org/job/ci_windows/1484/

dirk-thomas:
We had the same problem the last time a new version was released (which happened again today). Last time we only installed the latest version manually and I already mentioned that this wouldn't solve the problem but only the symptom. We need to modify the job to either:
* Provision and update the slave manually and disable the attempt to install the Python packages with pip within the job which didn't have admin access
* Allow the job to write to these files (admin access) which I wouldn't recommend since there job can then do anything to the system

jacquelinekay:
I had to fix a segfault that occurs in some cases with the changes to wait: http://ci.ros2.org/job/ci_windows/1488/

jacquelinekay:
had to deal with Windshield going offline:

[Build Status](http://ci.ros2.org/job/ci_windows/1490/)

jacquelinekay:
I think this is ready to be merged. The CI has only FastRTPS-services related failures. Anyone want to weigh in here?

wjwwood:
+1 to merge, if you're happy with the CI (I haven't looked at it closely). The code review was good for me before and the new commits also look ok to me.

jacquelinekay:
This also contains bbe90179554b6c986e6a4921daa62a71f8fbba79 which adds support for Android. I'm fine with adding it although I remember @dirk-thomas mentioned he wanted to implement Android support differently. Regardless I'll squash this into 2-3 commits instead of 1 to preserve the history a bit.

jacquelinekay:
Eh, maybe more like 5 commits.

dirk-thomas:
I think this being merged is making builds with Connext and FastRTPS at the same time fail. Because now two implementations providing C typesupport for Python both use introspection:
* http://ci.ros2.org/job/ci_linux/1499/
* http://ci.ros2.org/job/ci_osx/1163/
* likely same on Windows

jacquelinekay:
I'll look into it

PR #29, opened 2016-04-27
Explicit downcast to fix warning on Windows

codebot:
This time it only has one commit. Thanks Tully :smile:


wjwwood:
+1, sorry for causing you extra work 🍧.

PR #31, opened 2016-05-27
stub functions for wait_for_service

wjwwood:
This is needed so that rcl can compile. All the code which uses this downstream is configured to not test this code (which just raises 'not implemented').

Connects to ros2/ros2#215


wjwwood:
This is also in review. It's just the bare minimum for getting the rmw and rcl `wait_for_services` merged.

tfoote:
+1

PR #32, opened 2016-06-01
fix rmw_get_graph_guard_condition

wjwwood:
Previously it was just returning junk, which was left over from my `wait_for_service` work.

This doesn't implement it, but at least it returns a valid guard condition. I wouldn't return anything at all, but `rcl_node_init` calls it, so I have to return something valid or else all fastrtps programs would fail.

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -144,6 +144,11 @@ class ClientListener : public SubscriberListener
         std::condition_variable *conditionVariable_;
 };
 
+struct FastRTPSNodeImpl {
+    Participant * participant;
+    rmw_guard_condition_t * graph_guard_condition;
+};
+
 extern "C"
 {
     const char* const eprosima_fastrtps_identifier = "rmw_fastrtps_cpp";
@@ -272,7 +277,7 @@ extern "C"
             return NULL;
         }
 
-	eprosima::Log::setVerbosity(eprosima::VERB_ERROR);
+        eprosima::Log::setVerbosity(eprosima::VERB_ERROR);
 
         ParticipantAttributes participantParam;
         participantParam.rtps.builtin.domainId = static_cast<uint32_t>(domain_id);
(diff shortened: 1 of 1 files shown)


wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1369/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1067/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1374/)

tfoote:
lgtm, there are 6 instances of the 7 lines to get the participant from the impl that could be turned into a function. If there's expected to be more instances the helper function might be worth it. Otherwise we can leave it as is.

wjwwood:
Yeah, there's a lot of improvements that could be made, but with respect to deduplicating the code, in about half the cases you have to return `NULL` and the other half return `RMW_RET_ERROR` if there is a problem. This is why we have common macros to encapsulate this in rmw:

https://github.com/ros2/rmw/blob/master/rmw/include/rmw/impl/cpp/macros.hpp

And in `rcl` as well:

https://github.com/ros2/rcl/blob/master/rcl/src/rcl/common.h#L26-L33

I'd leave it as-is for now and try to deduplicate code between this and the other rmw implementations in the future.

wjwwood:
CI looks good (existing warnings on Windows and flaky tests on Linux).

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -287,14 +292,30 @@ extern "C"
...
+        if (!graph_guard_condition) {
+            // error already set
+            return NULL;
+        }
+
+        FastRTPSNodeImpl * node_impl = nullptr;
+        try {
+            node_impl = new FastRTPSNodeImpl();
+        } catch(std::bad_alloc & exc) {
+            RMW_SET_ERROR_MSG("failed to allocate node impl struct");
```
dirk-thomas: Doesn't this leak `graph_guard_condition`?

wjwwood: Yes. The function was not setup to handle unrolling. None of the functions in this file are except one. So I didn't want to change that in these changes. In general I think this time implementation needs some clean up.

wjwwood: https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/33

PR #35, opened 2016-06-03
Remove hardcoded rules about rcl_interfaces message types

jacquelinekay:
There are some hardcoded rules that disable parameter-related rcl_interfaces types by setting `typeTooLarge_`, which should cause an exception in serialization:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/TypeSupport.cpp#L180-L182
https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/TypeSupport.cpp#L327-L329
https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/TypeSupport.cpp#L589

I would expect that with these rules the FastRTPS ROS 2 parameter examples should never work, which contradicts what we see on CI.

I'm currently doing some testing to see how this affects the behavior of the parameters examples on Linux with Release build type.

### DIFF
--- a/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
+++ b/rmw_fastrtps_cpp/src/MessageTypeSupport.cpp
@@ -14,9 +14,6 @@ MessageTypeSupport::MessageTypeSupport(const rosidl_typesupport_introspection_cp
     assert(members);
     members_ = members;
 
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Closing this that has been resolved in https://github.com/ros2/rmw_fastrtps_cpp/commit/e336a8b0f93c8179cc2b4a62d02272403ed356bb

PR #37, opened 2016-06-04
make serialization of bool values more robust against uninitialized values

dirk-thomas:
Linux:
* before: http://ci.ros2.org/job/ci_linux/1390/testReport/
* after: http://ci.ros2.org/job/ci_linux/1391/testReport/
* -16 failing tests

OS X:
* before: http://ci.ros2.org/job/ci_osx/1085/testReport/
* after: http://ci.ros2.org/job/ci_osx/1086/testReport/
* -13 failing tests

The uninitialized values should be addressed too but they are coming from our code, e.g. `rclcpp::ParameterVariant`.

### DIFF
--- a/rmw_fastrtps_cpp/src/TypeSupport.cpp
+++ b/rmw_fastrtps_cpp/src/TypeSupport.cpp
@@ -202,8 +202,10 @@ bool TypeSupport::serializeROSmessage(eprosima::fastcdr::Cdr &ser, const rosidl_
             {
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_BOOL:
                     {
-                        bool sb = false, &b = *(bool*)field;
-                        if(b) sb = true; 
+                        bool sb = false;
+                        // don't cast to bool here because if the bool is
+                        // uninitialized the random value can't be deserialized
+                        if(*(uint8_t*)field) sb = true;
                         ser << sb;
                     }
                     break;


wjwwood:
Would it also help to give all message fields a valid default value in the constructor? I think we've discussed this before: https://github.com/ros2/rosidl/issues/77

rmw_fastrtps_cpp/src/TypeSupport.cpp
```diff
@@ -202,8 +202,10 @@ bool TypeSupport::serializeROSmessage(eprosima::fastcdr::Cdr &ser, const rosidl_
             {
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_BOOL:
                     {
-                        bool sb = false, &b = *(bool*)field;
-                        if(b) sb = true; 
+                        bool sb = false;
+                        // don't cast to bool here because if the bool is
+                        // uninitialized the random value can't be deserialized
+                        if(*(uint8_t*)field) sb = true;
```
wjwwood: I believe you that this change fixes the issue, but for my own edification, I'm having a hard time understanding this change.

Why does casting this to a `uint8_t` rather than a `bool` make any difference? In both cases `sb` is the thing being serialized and it is always either `true` or `false` and is never uninitialized.

Which cast are you expecting to happen here? Maybe it would be clearer if you used `reinterpret_cast<bool>` or `reinterpret_cast<uint8_t>` it would be clearer what's happening. Since `field` is only being read here, I don't see how casting it to a different type changes anything other than the branch the `if` statement takes. I feel like I'm missing something obvious.

wjwwood: I could imagine that `*(bool *)field` could result in a bad access if `field` points to memory outside of the process. I could also imagine that copying the memory contents at `field` into a bool on the stack could mess up the stack, but I wouldn't think that's possible since the space reserved for `b` on the stack would be the same size as the memory copied from `field` and therefore the bogus contents of memory at `field` should not "leak" into the stack. I guess the compiler could be generating slightly different code here as an optimization which would make that possible.

Maybe I'm completely off-base, I'm just thinking out loud.

dirk-thomas: Basically it comes down to uninitialized values result in undefined behavior. One example see http://markshroyer.com/2012/06/c-both-true-and-false/ In this specific case it looks like that the compiler generated assembler code makes assumptions about the value being 0 or 1 (since it is supported to be bool) which leads to sb actually taking the value of field (since the compiler reuses a register which is supposed to contain the intended value.

dirk-thomas:
Commonly our code already sets all fields values. The biggest exception is `rclcpp::ParameterVariant` since currently the class only sets the fields for specified type and keeps all other fields uninitialized. I will probably update those as well as the constructors for messages in Monday.

It would be good to get this merged before the next nightly jobs to see the results across all build types and platforms.

wjwwood:
So, would it be fair to say that an unspoken rule is: "don't publish a message unless you've explicitly initialized all of the fields yourself"? (except for fields with explicitly called out default values)

If that's the case then I think it would be better to address that by properly initializing all fields (regardless of whether or not they have a default value set in the .msg file) in the message constructor for C++ and with a generated function for C. For two reasons: it's inconvenient and error-prone to ask the users to always remember to do this and it opens us up to subtle bugs which get introduced when a field is added to a message that's already in use.

Consider the current case of `ParameterVariant`. If you update the code to explicitly initialize the bool field that's giving us trouble here it will fix the problem. However, if we later add another bool field to the message and forget to update the code then we've reintroduced the bug. But if we properly initialize all fields in the constructor then we avoid this issue completely and it allows us to write more concise code when we're only interested in filling out a few of a message's fields.

As for merging this fix, I'm fine with it. +1

rmw_fastrtps_cpp/src/TypeSupport.cpp
dirk-thomas: What the compiler does is basically optimize this line to look like this:

```
// if(b) sb = true;
// since the compiler "knows" that b is true
// it generates assembler code reusing the already loaded value
if(b) sb = b;
```

PR #38, opened 2016-06-06
fix compiler warning on windows

codebot:
unary minus on an unsigned int gives a compiler warning in windows.


dirk-thomas:
Can you please run a Windows CI job with this patch and then link the new warning results.

tfoote:
does that not need to be a 1u like earlier in the line? otherwise lgtm

PR #39, opened 2016-06-07
Add C++11 flag

esteve:
Added C++11 flag, like in https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/CMakeLists.txt#L5

Connects to https://github.com/ros2/rmw/pull/69


esteve:
@wjwwood thanks.

PR #42, opened 2016-06-27
Attributes are now correctly passed by reference in rmw_fastrtps_cpp

adamdbrw:
- Fixes issue with qos settings not applying in rmw_fastrtps_cpp (at all)


adamdbrw:
@wjwwood @jacquelinekay

dirk-thomas:
+1

dirk-thomas:
Thank you for the patch.

PR #45, opened 2016-06-29
Serialize and deserialize arrays

esteve:
This PR serializes and deserializes unbounded arrays. Tested locally with two Python scripts (listener and talker), using different permutations of OpenSplice and FastRTPS.

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
@@ -287,7 +287,6 @@ size_t get_array_size_and_assign_field(
     void *ptr = (void*)sub_members_size;
     size_t vsize = vector->size() / (size_t)align_(max_align, 0, ptr, space);
     if (member->is_upper_bound_ && vsize > member->array_size_) {
-        printf("vector overcomes the maximum length\n");
         throw std::runtime_error("vector overcomes the maximum length");
     }
     subros_message = reinterpret_cast<void*>(vector->data());
@@ -370,7 +369,6 @@ bool TypeSupport<MembersType>::serializeROSmessage(
                         // Control maximum length.
                         if((member->string_upper_bound_ && str.length() > member->string_upper_bound_ + 1) || str.length() > 256)
                         {
-                            printf("string overcomes the maximum length with length %zu\n", str.length());
                             throw std::runtime_error("string overcomes the maximum length");
                         }
                         ser << str;
@@ -383,7 +381,6 @@ bool TypeSupport<MembersType>::serializeROSmessage(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
@@ -462,7 +459,6 @@ bool TypeSupport<MembersType>::serializeROSmessage(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
@@ -637,7 +633,6 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
@@ -716,7 +711,6 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
@@ -772,7 +766,6 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
@@ -825,7 +818,6 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
                     }
                     break;
                 default:
-                    printf("unknown type id %u\n", member->type_id_);
                     throw std::runtime_error("unknown type");
             }
         }
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
@@ -249,7 +249,7 @@ void serialize_array(
         ser.serializeArray((T*)field, member->array_size_);
     } else {
         auto & data = *reinterpret_cast<typename GenericCArray<T>::type *>(field);
-        ser.serializeArray((T*)data.data, data.size);
+        ser.serializeSequence((T*)data.data, data.size);
     }
 }
 
@@ -498,6 +498,12 @@ void deserialize_array(
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
         deser.deserializeArray((T*)field, member->array_size_);
+    } else {
+        auto & data = *reinterpret_cast<typename GenericCArray<T>::type *>(field);
+        int32_t dsize = 0;
+        deser >> dsize;
+        GenericCArray<T>::init(&data, dsize);
+        deser.deserializeArray((T*)data.data, dsize);
     }
 }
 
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -259,6 +259,7 @@ void serialize_array(
             printf("vector overcomes the maximum length\n");
             throw std::runtime_error("vector overcomes the maximum length");
         }
+        ser << (int32_t)data.size;
```
dirk-thomas: Instead of replicating the wire format here this should use the API of Cdr to perform the serialization. I guess in this case it should call `serializeSequence` instead of serializing the size and calling `serializeArray`.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -514,6 +515,12 @@ void deserialize_array(
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
         deser.deserializeArray((T*)field, member->array_size_);
+    } else {
+        auto & data = *reinterpret_cast<typename GenericCArray<T>::type *>(field);
+        int32_t dsize;
+        deser >> dsize;
+        GenericCArray<T>::init(&data, dsize);
```
dirk-thomas: Since `field` is a  ROS 2 array the corresponding `Array__init` function in `rosidl_generator_c` must be used.

esteve: Unless I'm mistaken, that's what it already does, according to https://github.com/esteve/ROS-RMW-Fast-RTPS-cpp/blob/array-serialization/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/macros.hpp#L31

dirk-thomas: I didn't see that. It looked like a FastRTPS specific macro.

All good then.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
esteve: Done.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -528,6 +534,22 @@ void deserialize_array<std::string>(
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
         deser.deserializeArray(((rosidl_generator_c__String*)field)->data, member->array_size_);
+    } else {
+        std::vector<std::string> cpp_string_vector;
+        deser >> cpp_string_vector;
+
+        auto & string_array_field = *reinterpret_cast<rosidl_generator_c__String__Array *>(field);
+        if(!rosidl_generator_c__String__Array__init(&string_array_field, cpp_string_vector.size())) {
+            printf("unable to initialize rosidl_generator_c__String array\n");
```
dirk-thomas: All error messages should go to `stderr`. But since the exception is thrown anyway I would say the error message is redundant and can be removed.

wjwwood:
Thanks for the patch @esteve! I agree that the `printf` statements, while useful when debugging, are probably better left out.

With that comment addressed the changes lgtm. I'll just start some CI (assuming that removing the `printf`'s won't impact the result):

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1605/) / [Build Status](http://ci.ros2.org/job/ci_linux/1606/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1218/) / [Build Status](http://ci.ros2.org/job/ci_osx/1219/)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -514,6 +514,12 @@ void deserialize_array(
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
         deser.deserializeArray((T*)field, member->array_size_);
+    } else {
+        auto & data = *reinterpret_cast<typename GenericCArray<T>::type *>(field);
+        int32_t dsize;
```
wjwwood: Looks like cppcheck is concerned about this not be initialized:

http://ci.ros2.org/job/ci_linux/1606/testReport/(root)/projectroot/cppcheck/

You could either give it a default value or use the no lint comment to squelch cppcheck.

esteve: Fixed.

esteve:
@wjwwood the `printf` statements next to throwing exceptions seem to be a recurring pattern in eProsima's code:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h#L402

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h#L481

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h#L239

so I'm going to leave the `printf` statements in this PR, just for consistency, though I agree they all should be removed since they don't add much value.

esteve:
@richiprosima I've rebased this and removed the `printf` calls, this should be mergeable now.

dirk-thomas:
Can you please trigger the CI builds for these changes.

wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1712/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1274/)

dirk-thomas:
* Windows:
  * [Build Status](http://ci.ros2.org/job/ci_windows/1629/)

firesurfer:
I tested this PR with my .Net wrapper and it seems to work fine, apart from passing c-typesupport messages to cpp.

What works:
* Sending and recieving messages in program(s) that use c-typesupport
* Recieving messages sent from a cpp program in a program that uses the same  messages but with c-typesupport. 

What doesn't work:
* Sending messages from a program that uses messages with c-typesupport and recieving them in a cpp program.

esteve:
@firesurfer thanks for testing this PR. That's weird, once the messages go through the wire, the typesupport should matter. Do you have a concrete example which I can test? I recall I tested this between rclpy (c-typesupport) and rclpp (c++-typesupport), and it seemed to work, but I'll check again.

firesurfer:
@esteve I tested it only with my C# wrapper so far. In my testing workspace ( https://github.com/firesurfer/rclcs_testing_ws ) I got one program that uses in C#  - test_cs.exe (c-typesupport) and one program that uses  c++ -  test_cpp (c++-typesupport). Sending a message from the C# program to another program that uses the C# wrapper works. But sending a message from the C# wrapper to the C++ program doesn't work.  

On the other side a program that uses the C# wrapper receives the message published from a C++ program. I haven't had the time yet to investigate this issue further but I think I can find some time to have a deeper look at it tomorrow.

esteve:
@firesurfer do you have the message definition you used? I can test it out and see if I can find the problem.

firesurfer:
@esteve  this is my dummy message: https://github.com/firesurfer/rclcs_testing_ws/blob/master/src/test_msgs/msg/Dummy.msg

esteve:
@firesurfer thanks!

mikaelarguedas:
CI with fastrtps whitelisted in `test_communication`
Master:
 [Build Status](http://ci.ros2.org/job/ci_linux/1722)
This PR:
 [Build Status](http://ci.ros2.org/job/ci_linux/1721)

mikaelarguedas:
TD;DR this PR does fix the serialisation/deserialisation of dynamic primitive array messages (`DynamicArrayPrimitives` and `BoundedArrayPrimitives`) but doesn't fix the other arrays test cases, it can be merged as is without resulting in any regression.

@esteve do you prefer:
1) to merge this as is and we'll create a follow-up PR to address the remaining array test failures
2) allow us to [push commits on your branch](https://github.com/blog/2247-improving-collaboration-with-forks) and merge this PR once it fixes all the array serialization problems ?

Here is the results of my investigation for the remaining test failures:
 * All non array test failures (`Nested`/`Primitives`) are due to `rclpy.spin_once()` behaving differently with fastrtps and other rmw implementations, need more investigation to figure out why, quick fix: relax test timeout
 * `StaticArrayPrimitives` fails because fixed size string arrays are not serialized/deserialized properly (the function `serializeArray` is only applied the first `rosidl_generator_c__String` element and the rest of the array is dropped (c.f. https://github.com/esteve/ROS-RMW-Fast-RTPS-cpp/blob/a4d829308e18d3c3d24b58f903f803bd8f1ada18/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h#L267)). It uses the size of the String array instead of the size of the char array so the communication works only if you have an array longer than the individual strings in that array
 * `StaticArrayNested` IMO same string (de)serialization problem as `StaticArrayPrimitives`, works well if we remove strings from the equation
 * `DynamicArrayNested` feature not implemented (c.f. [this todo](https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h#L305))
 * `BoundedArrayNested` I assume it is the same as `DynamicArrayNested` but we may uncover more modifications to be done once the previous test case is fixed

esteve:
Thank you so much for looking into this @mikaelarguedas!

I think it'd be fine to just merge this PR and continue any remaining bits
in a follow-up PR, so that @firesurfer and I can keep working on our stuff.
What's in this PR, while not complete, is still some progress.

dirk-thomas:
After talking to @mikaelarguedas we will merge this as-is and go from there. Thanks.

mikaelarguedas:
@esteve @firesurfer #58 may solve some of the issues you encountered. The tests are passing in both rclcpp -> rclpy and rclpy -> rclcpp for all message types

PR #49, opened 2016-07-06
Use a universal reference to avoid extra std::string copies

esteve:
@jacquelinekay I noticed that the universal reference was removed when https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/27 was merged, I don't know if it was causing any issues on the buildfarm, I checked locally and it seems to work fine.

Alternatively, I also pushed #50, which accomplishes the same but without using universal references, in case the universal reference were causing trouble.

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
@@ -381,7 +381,7 @@ bool TypeSupport<MembersType>::serializeROSmessage(
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
                     {
-                        auto str = StringHelper<MembersType>::convert_to_std_string(field);
+                        auto && str = StringHelper<MembersType>::convert_to_std_string(field);
 
                         // Control maximum length.
                         if((member->string_upper_bound_ && str.length() > member->string_upper_bound_ + 1) || str.length() > 256)


jacquelinekay:
Sorry for changing this line in in #27, it wasn't intentional.

However, I still don't see why this is necessary. I think that a modern compiler will know to implement return value optimization to avoid the extra copy from returning std::string by value.

https://blog.knatten.org/2011/08/26/dont-be-afraid-of-returning-by-value-know-the-return-value-optimization/

I also made this quick test that I think replicates the situation you're talking about but instruments the copy and move constructors of a struct to see if RVO is used on my compiler. When I run it, I see that neither the copy or move constructor is invoked.

https://gist.github.com/jacquelinekay/b6098b211beaf7d1cc53b2f0f12f70b7

But this doesn't necessarily prove anything about this particular line of code. Did you observe an extra string copy originating from this line when returning the string by value? If you can show that this code produces an extra string copy, then that would change my mind.

esteve:
@jacquelinekay here's a synthetic example of what happens when a universal reference is not used:

https://gist.github.com/esteve/e03870d027965798d5ed37d2ff5e68e8

It'll produce a new copy of `std::string`, since the address of the original string (s) is the same as the one from the universal reference (s2). This also shows that the alternate approach in #50 works, as it'll print the same address too when using `std::string &`:

esteve:
Here's the output of that program:

```
Address of string s: 0x7fff69d85360
Address of string s1: 0x7fff69d85370
Address of string s2: 0x7fff69d85360
Address of string s2_copy: 0x7fff69d85380
Address of string str: 0x7fff69d85330
Address of string str: 0x7fff69d85360
```

jacquelinekay:
I'm not convinced that your example is consistent with what what's happening in this library. It could still be valid, it just has a few differences.

Both specializations of convert_to_std_string return a std::string by value, whereas in your example, the specialization for `this_will_not_create_a_copy` returns a string by references. Also, in your example you cast a void * to a string type, which gets passed to the constructor of std::string, while in `convert_to_std_string`, the void * is cast to a `rosidl_generator_c__string`, which then passes its underlying data (a `char *`) to the constructor of std::string.

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/187964ce080632fd53bbf0569fe986fb20713753/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h#L54-L69

esteve:
`convert_to_std_string` does return a reference:

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/187964ce080632fd53bbf0569fe986fb20713753/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h#L84

the point of this PR is that when the C++ specialization is used (`rosidl_typesupport_introspection_cpp::MessageMembers`), `StringHelper` just casts `field` as a `std::string`, avoiding a new copy entirely. This is in the same spirit as the original code, which used a `std::string &` https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/20d43ef13bf668eb75dd61e226730ec26960d93e/rmw_fastrtps_cpp/src/TypeSupport.cpp#L246

jacquelinekay:
Oh, sorry, I was only looking at the C specialization of that function, where avoiding a copy of the underlying C array is not possible.

Looking at the C++ specialization for `convert_to_std_string`, it would probably be good to check for null before dereferencing the the string--but that's a side note.

I guess I've learned my lesson that auto doesn't deduce reference types.

+1 to this solution since it's more concise and allows for deducing either kind of reference.

esteve:
@jacquelinekay good point about avoiding dereferencing null. Thanks for taking the time to review this PR.

PR #50, opened 2016-07-06
Avoid creating a new copy of std::string

esteve:
Alternate approach to avoid extra `std::string` copies, if this PR is merged, #49 should be closed.


jacquelinekay:
I'm not sure if taking the reference to the returned string as done in this pull request would work. Wouldn't that involve taking an lvalue reference of a temporary?

esteve:
@jacquelinekay it just works, see https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/49#issuecomment-230828190

jacquelinekay:
Going to close this in favor of #49

PR #53, opened 2016-08-04
Format fixes

vmayoral:
It gave us a few headaches so in case it's useful, here's our code formatting contribution


dirk-thomas:
I know that style is a very subjective topic. Should we follow the style guide of ROS 2 instead (which will allow us to enable automatic linting for it)?

vmayoral:
@dirk-thomas sure. For the record, leaving here the ROS 2 style:
https://github.com/ros2/ros2/wiki/Developer-Guide#c-1

PR #51, opened 2016-07-08
Large message pub sub revision

wjwwood:
Replacement for #47 and closes #36.

I tested this and it seems to work as expected, with the caveats about OS X mentioned in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/36#issuecomment-229718748.

This requires my FastCDR branch (https://github.com/eProsima/Fast-CDR/pull/7) and the feature branch of Fast-RTPS (https://github.com/eProsima/Fast-RTPS/tree/feature/dynamicmemory_revision) for which there is no pull request yet.

PR #54, opened 2016-08-09
Use new api

dirk-thomas:
The Log API changed in eProsima/Fast-RTPS@f94ba23f40a8d669961786f540af09edd1ff1212.

The new method `TypeSupport::` was introduced in eProsima/Fast-RTPS#59. I am just not sure if calling `calculateMaxSerializedSize` is appropriate here.

While it makes it compile again (http://ci.ros2.org/job/ci_linux/1691/) it fails several tests due to the new size provider. @richiprosima Maybe you can advise what the correct way of implementing this is.

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
@@ -112,6 +112,8 @@ namespace rmw_fastrtps_cpp
 
             bool deserialize(SerializedPayload_t *payload, void *data);
 
+            std::function<uint32_t()> getSerializedSizeProvider(void* data);
+
             void* createData();
 
             void deleteData(void* data);
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/MessageTypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/MessageTypeSupport_impl.h
@@ -41,7 +41,7 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType *members)
     if(members->member_count_ != 0)
         this->m_typeSize = static_cast<uint32_t>(this->calculateMaxSerializedSize(members, 0));
     else
-        this->m_typeSize = 1;
+        this->m_typeSize = 4;
 }
 
 #endif // _RMW_FASTRTPS_CPP_MESSAGETYPESUPPORT_IMPL_H_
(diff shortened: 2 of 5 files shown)


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
```diff
@@ -116,6 +116,13 @@ namespace rmw_fastrtps_cpp
 
             void deleteData(void* data);
 
+            std::function<uint32_t()> getSerializedSizeProvider(void* data) {
+              (void)data;
+              return [&]() -> uint32_t {
```
wjwwood: I don't think it needs to capture `data`? Maybe it should just be `[]() -> uint32_t`.

richiprosima:
+1 in my Linux machine

dirk-thomas:
* http://ci.ros2.org/job/ci_linux/1696/
* http://ci.ros2.org/job/ci_osx/1258/
* http://ci.ros2.org/job/ci_windows/1614/

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
dirk-thomas: The `&` is not there to capture `data` but to capture `this` which is necessary in order to access `members_`.

dirk-thomas:
@richiprosima I don't think the patch works correctly as-is. The test failures are still the same as in my original comment and seem to be clearly related to the new message size limit. Can you please suggest how we can address them?

dirk-thomas:
Even with the latest changes this still fails several tests:

* http://ci.ros2.org/job/ci_linux/1701/
* http://ci.ros2.org/job/ci_osx/1263/
* http://ci.ros2.org/job/ci_windows/1620/

wjwwood:
#51 may overlap with this a little bit, I'll look into that.

dirk-thomas:
I think we should first merge this PR to get the current state working again (once it passes all the tests). And then integrate the new features form #51.

wjwwood:
This has been open for 16 days, so 1 more day isn't going to make a big difference. I am going to see what the state of #54 is by fixing the branch and then doing CI. If it needs more work we can merge this one and rebase that one, but if it works then it's probably easier to merge #54 and close this one. Especially if #54 fixes the remaining failing tests that this pr has.

wjwwood:
The latest nightly builds for Linux are green. The OS X and Windows warnings and test failures are not related to these problems, I think this can be closed since #51 was merged.

PR #58, opened 2016-09-21
serialize deserialize arrays

mikaelarguedas:
follow-up of #45 
The goal of this PR is to provide support for all ROS2 message types in rmw_fastrtps_cpp

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
@@ -49,6 +49,7 @@ SPECIALIZE_GENERIC_C_ARRAY(uint32, uint32_t)
 SPECIALIZE_GENERIC_C_ARRAY(int64, int64_t)
 SPECIALIZE_GENERIC_C_ARRAY(uint64, uint64_t)
 
+// TODO(mikaelarguedas) find right spot to define these
 typedef struct rosidl_generator_c__void__Array
 {
   void * data;
@@ -57,6 +58,51 @@ typedef struct rosidl_generator_c__void__Array
   /// The number of allocated items in data
   size_t capacity;
 } rosidl_generator_c__void__Array;
+// SPECIALIZE_GENERIC_C_ARRAY(void, void)
+
+bool
+rosidl_generator_c__void__Array__init(rosidl_generator_c__void__Array * array, size_t size, size_t member_size)
+
+{
+  if (!array) {
+    return false;
+  }
+  void * data = NULL;
+  if (size) {
+    data = (void *)calloc(size, member_size);
+    if (!data) {
+      return false;
+    }
+  }
+  array->data = data;
+  array->size = size;
+  array->capacity = size;
+  return true;
+}
+
+void
+rosidl_generator_c__void__Array__fini(rosidl_generator_c__void__Array * array)
+{
+  if (!array) {
+    return;
+  }
+  if (array->data) {
+    // ensure that data and capacity values are consistent
+    assert(array->capacity > 0);
+    // finalize all array elements
+    free(array->data);
+    array->data = NULL;
+    array->size = 0;
+    array->capacity = 0;
+  } else {
+    // ensure that data, size, and capacity values are consistent
+    assert(0 == array->size);
+    assert(0 == array->capacity);
+  }
+}
+
+
+
 
 
 template <typename MembersType>
@@ -381,7 +427,6 @@ size_t get_array_size_and_assign_field(
     size_t, size_t, size_t)
 {
     rosidl_generator_c__void__Array * tmparray = (rosidl_generator_c__void__Array *) field;
-    void * ptr = (void *)(tmparray->size);
     if (member->is_upper_bound_ &&  tmparray->size > member->array_size_) {
         throw std::runtime_error("vector overcomes the maximum length");
     }
@@ -654,13 +699,18 @@ size_t get_submessage_array_deserialize(
     eprosima::fastcdr::Cdr & deser,
     void * field,
     void *& subros_message,
-    bool, size_t, size_t, size_t)
+    bool,
+    size_t sub_members_size,
+    size_t, size_t)
 {
-  (void)member;
-  uint32_t size = 0;
-  deser >> size;
-  subros_message = field;
-  return size;
+    (void)member;
+    // Deserialize length
+    uint32_t vsize = 0;
+    deser >> vsize;
+    rosidl_generator_c__void__Array * tmparray = (rosidl_generator_c__void__Array *)field;
+    rosidl_generator_c__void__Array__init(tmparray, vsize, sub_members_size);
+    subros_message = reinterpret_cast<void*>(tmparray->data);
+    return vsize;
 }
 
 template <typename MembersType>
@@ -680,9 +730,7 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
             switch(member->type_id_)
             {
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_BOOL:
-                    {
-                        deser >> *(bool*)field;
-                    }
+                    deser >> *(bool*)field;
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_BYTE:
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT8:
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
@@ -182,9 +182,12 @@ static size_t calculateMaxAlign(const MembersType *members)
                     alignment = alignof(uint64_t);
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
-                    // StringHelper exposes rosidl_generator_c__String as std::string
-                    // so we don't need differentiate between C and C++ introspection typesupport
-                    alignment = alignof(std::string);
+                    // Note: specialization needed because calculateMaxAlign is called before
+                    // casting submembers as std::string, returned value is the same on i386
+                    if(std::is_same<MembersType, rosidl_typesupport_introspection_c__MessageMembers>::value)
+                        alignment = alignof(rosidl_generator_c__String);
+                    else
+                        alignment = alignof(std::string);
                     break;
                 case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
                     {
@@ -202,8 +205,8 @@ static size_t calculateMaxAlign(const MembersType *members)
     return max_align;
 }
 
-// C++ specialization
-static size_t size_of(const rosidl_typesupport_introspection_cpp::MessageMembers *members)
+template <typename MembersType>
+static size_t size_of(const MembersType *members)
 {
     size_t size = 0;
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
This is WIP and not ready for merge

mikaelarguedas:
Status:
 * `StaticArrayPrimitives`:  static string array fixed in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/87d7c887b10a4267960a5fa2e424b37a2bb644f6
 * `StaticNestedArray`: fixed by specializing `size_of` function for c arrays https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/1d97127e239b83c4596fa3269d4a9a706eb527e5
 * `BoundedNestedArray` and `DynamicNestedArray`:
   * serialization fixed by introducing a generic nested C array structure in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/ad46b7a4743740237cb878e916c560d7630047a3 and adapting `get_array_size_and_assign_field` accordingly https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/2a880f01f93ca9746a436629035ce83d1fb0e28d
   * deserialization not yet implemented

mikaelarguedas:
CI: master
[Build Status](http://ci.ros2.org/job/ci_linux/1752)
this PR:
[Build Status](http://ci.ros2.org/job/ci_linux/1754)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -49,6 +49,56 @@ SPECIALIZE_GENERIC_C_ARRAY(uint32, uint32_t)
 SPECIALIZE_GENERIC_C_ARRAY(int64, int64_t)
 SPECIALIZE_GENERIC_C_ARRAY(uint64, uint64_t)
 
+// TODO(mikaelarguedas) find right spot to define these
```
mikaelarguedas: NOTE: I wanted to implement these arrays through the PRIMITIVE_ARRAY macros [here](https://github.com/ros2/rosidl/blob/master/rosidl_generator_c/include/rosidl_generator_c/primitives_array_functions.h#L29) and [here](https://github.com/ros2/rosidl/blob/master/rosidl_generator_c/include/rosidl_generator_c/primitives_array.h#L22) and the associated c files. Problem is that these macros rely on the result of `sizeof` for memory allocation where here we would like to specify the size of the custom array members via another argument (c.f. l64 parameter `member_size`).
I don't think that adding an optional parameter to all C array definition macros just to cover this specific use case is the way to go.  So I'm considering keeping this `void__Array` definition as a separate piece of code.

I'm open to suggestions regarding which approach makes more sense and, if the latter, where to store these definitions.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -203,13 +253,83 @@ static size_t size_of(const MembersType *members)
...
+        }
+
+        if(last_member.is_array_)
+            size *= last_member.array_size_;
+    }
+
+    return last_member.offset_ + size;
+}
+
+// C specialization
```
mikaelarguedas: It surprisingly seems to work well without having to differentiate the `__align` function as well.
I'm a bit confused about this, am I missing something here ?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
dirk-thomas: I agree that making the existing work for this use case is a good idea.

I think it is ok to keep the new struct and functions where they are. If we ever need them somewhere else we can easily refactor / move them to a more generic place.

mikaelarguedas:
@dirk-thomas @wjwwood I can't label this as in review, if any of you has time to have a look and comment on the uncertainties I pointed out with inline comments that'd be amazing,
Thanks!

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: `alignof` does return the same value for both the `rosidl_generator_c__String`  representation and `std::string`. It could be worth it to check if it's the case on other architecture (e.g. ARM)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: ok, then I will keep it here from now without modifying existing macros in rosidl

mikaelarguedas:
C/C++ differentiation for `size_of` and `calculateMaxAlign` updated to avoid code duplication.
Will rerun CI on all platforms once review is done and comments addressed

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
dirk-thomas: Then the todo can be removed.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -513,7 +575,18 @@ void deserialize_array<std::string>(
 {
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
-        deser.deserializeArray(((rosidl_generator_c__String*)field)->data, member->array_size_);
+        auto deser_field = (rosidl_generator_c__String*)field;
+        std::string stringseq[member->array_size_];
+        deser.deserializeArray(stringseq, member->array_size_);
+        for (size_t i = 0; i < member->array_size_; ++i) {
+          // NOTE: Do we need to call init ?
```
dirk-thomas: This should be answered / addressed before merging. On a first look I am not sure what the answer is though.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -263,13 +318,18 @@ void serialize_array<std::string>(
     // First, cast field to rosidl_generator_c
     // Then convert to a std::string using StringHelper and serialize the std::string
     if (member->array_size_ && !member->is_upper_bound_) {
-        auto string_field = (rosidl_generator_c__String *) field;
-        ser.serializeArray(string_field->data, member->array_size_);
+      std::string stringseq[member->array_size_];
```
mikaelarguedas: Just remembered that MSVC doesnt like VLAs. This won't compile on windows

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: Done in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/07757844e03acd018a4abe54005b271240cd001c

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: It seems to me that it's not needed given that assign will call `rosidl_generator_c__String__fini` on the data structure before trying to assign elements. But I wasn't sure so I kept it commented hoping that someone would confirm it during review

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: fixed in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/8b844a55182d24e343573a87ff3bd0d0fb06b9e4

mikaelarguedas:
windows job afer https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/8b844a55182d24e343573a87ff3bd0d0fb06b9e4
[Build Status](http://ci.ros2.org/job/ci_windows/1676)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
```diff
@@ -513,7 +574,18 @@ void deserialize_array<std::string>(
 {
     (void)call_new;
     if (member->array_size_ && !member->is_upper_bound_) {
-        deser.deserializeArray(((rosidl_generator_c__String*)field)->data, member->array_size_);
+        auto deser_field = (rosidl_generator_c__String*)field;
+        std::vector<std::string> stringseq(member->array_size_);
```
dirk-thomas: This looks weird to me. Later you treat all items in this vector as c strings. Shouldn't the vector be of that type then?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
dirk-thomas: I don't think the conclusion is correct. The invocation of `rosidl_generator_c__String__fini` will read the `data` member and based on its content do conditional logic. But that variable has never been initialized. So I expect this to break in release code / corrupt memory. So imo the init call is required.

mikaelarguedas: Good point. added back in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/c15cfc342e6bcaa09a638a9a71a331134dd540c2

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
dirk-thomas: If the two types `std::string` and `rosidl_generator_c__String` have different sized (which I would expect) the vector allocates the wrong amount of memory. In case the first type is bigger its not fatal but in case the second type is bigger the code will write into not owned memory. Imo this must use the type it is being used for. rmw has access to both FastRTPS internal data structures as well as ROS 2 internal data structures since it is bridging between the two.

dirk-thomas: Forget my comments about this line. I was thinking about `field` and not `stringseq`. Their code really deserializes into `std::string` and then we copy it again from there into a c string. Not efficient but ok for now.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
dirk-thomas: After discussing this I think we have to assume that the passed in `deser_field[i]` has been correctly initialized before calling this. (We can't even check for that since uninitialized is not distinguishable.) It is also not coming from any user land code but our own rmw code. So if we don't screw it up there is will be initialized correctly. So the init call needs to be removed again. Sorry for the back and forth.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: updated using a single temporary std::string for storage https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/58/commits/b04928bc0fe9b9b4b789cf7bc54fd9b6eef01b7f

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
mikaelarguedas: commit reverted

mikaelarguedas:
new set of CI with latest changes: these jobs will test cross rmw and cross rcl communication.
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1757)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1306/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1678/)

mikaelarguedas:
new round of CI using Release build type:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1758)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1307/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1679/)

mikaelarguedas:
All tests from `test_communication` passed despite the weird test report on windows (repeated tests appear as failed even if they passed on a later iteration, we can discuss this issue elsewhere).
So I think this is ready to be squashed/merged

PR #59, opened 2016-10-02
Error checking get names and types

mikaelarguedas:
This checks that the structure `topic_names_and_types` is properly initialized
needed for https://github.com/ros2/rclpy/pull/40

connects to https://github.com/ros2/rclpy/pull/40

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -2286,10 +2286,6 @@ fail:
 	    RMW_SET_ERROR_MSG("node handle not from this implementation");
 	    return RMW_RET_ERROR;
 	}
-	if (!topic_names_and_types) {
-	  RMW_SET_ERROR_MSG("topics handle is null");
-	  return RMW_RET_ERROR;
-	}
 	if (topic_names_and_types->topic_count) {
 	  RMW_SET_ERROR_MSG("topic count is not zero");
 	  return RMW_RET_ERROR;
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -2281,11 +2281,27 @@ fail:
 		return RMW_RET_ERROR;
 	}
 	//Get participant pointer from node
-        if(node->implementation_identifier != eprosima_fastrtps_identifier)
-        {
-            RMW_SET_ERROR_MSG("node handle not from this implementation");
-            return RMW_RET_ERROR;
-        }
+  if(node->implementation_identifier != eprosima_fastrtps_identifier)
+  {
+      RMW_SET_ERROR_MSG("node handle not from this implementation");
+      return RMW_RET_ERROR;
+  }
+  if (!topic_names_and_types) {
+    RMW_SET_ERROR_MSG("topics handle is null");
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -2281,11 +2281,27 @@ extern "C"
...
-        }
+	if(node->implementation_identifier != eprosima_fastrtps_identifier)
+	{
+	    RMW_SET_ERROR_MSG("node handle not from this implementation");
+	    return RMW_RET_ERROR;
+	}
+	if (!topic_names_and_types) {
+	  RMW_SET_ERROR_MSG("topics handle is null");
+	  return RMW_RET_ERROR;
+	}
```
dirk-thomas: This seems to duplicate the check from line 2279-2282.

mikaelarguedas: Sorry about that: fixed in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/59/commits/546ffa52cf8c99867e3d130a93545d1732d27fac

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -2281,11 +2281,23 @@ extern "C"
...
-        {
-            RMW_SET_ERROR_MSG("node handle not from this implementation");
-            return RMW_RET_ERROR;
-        }
+	if(node->implementation_identifier != eprosima_fastrtps_identifier)
+	{
+	    RMW_SET_ERROR_MSG("node handle not from this implementation");
+	    return RMW_RET_ERROR;
+	}
+	if (topic_names_and_types->topic_count) {
```
dirk-thomas: Why is it important that this value is zero?

mikaelarguedas: if there is not topic advertised, the structure will be returned unmodified. So the topic_count will be whatever value is stored at that address. Results in segfault when we try to iterate over the returned lists

dirk-thomas: Should the function set the count to zero in that case instead?

dirk-thomas: I guess the patch is doing the same as the other rmw impl. atm. So it can stay like this I guess. (Still not convinced that this is a nice API...)

mikaelarguedas: Yeah I replicated the behavior of the other rmw impl. we can always improve it in a follow-up PR if we settle on a better API for this

wjwwood: This could probably be addressed in rcl. Currently the docs don't mentioned "zero initialized" and only mention that the structure pointer should point to allocated space:

https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h#L35-L58

We could push this check up to rcl and/or ensure it is setup properly before passing to rmw.

The whole point of the "zero initialized" requirement I have for structure pointers throughout rcl is to catch exactly this case, i.e. create_x fails and so the memory is still uninitialized and the user passes the allocated, but uninitialized, structure to destroy which assumes it is not pointing to bogus data.

PR #65, opened 2016-10-31
update style and enable linters

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/1862/)
* OS X [Build Status](http://ci.ros2.org/job/ci_osx/1397/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/1780/)

### DIFF
--- a/fastrtps_cmake_module/CMakeLists.txt
+++ b/fastrtps_cmake_module/CMakeLists.txt
@@ -12,14 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.5)
 
 project(fastrtps_cmake_module)
 
 find_package(ament_cmake REQUIRED)
 
-#list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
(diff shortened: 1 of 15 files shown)


dirk-thomas:
Ready for review.

wjwwood:
LGTM too, I'd suggest making code best practice changes separately (other than maybe use of `{}` blocks), so that they are easier to review. For example, which C++ style cast you use instead of the catch all C style cast should be reviewed carefully in each case. And it's easier to do that without the whitespace changes.

dirk-thomas:
@mikaelarguedas I haven't seen `(T * t)` in the diff anywhere. I only touched the code which is now being linted and needed update. Please feel free to create a follow up PR to add linting to additional parts.

@wjwwood I agree that it would be nicer to have smaller easier to review pieces. In this case the code was so different from our style that the changes are significant. Splitting those up would have required more time. The diff can easily be searched for e.g. `_cast<` in order to review those cases. I used `static_cast` when going from a `void *` to `Something *` and `reinterpret_cast` in the other remaining cases.

mikaelarguedas:
@dirk-thomas agreed it can definitely be addressed on a separate PR, hence the approval. I was just curious when I saw all these `(T *)` while reviewing and wasn't sure if I should act on it (and if the linters were supposed to catch it), hence the question. I'll open a follow-up PR to address them

PR #69, opened 2016-11-29
(fix) init c_str correctly

Karsten1987:
@mikaelarguedas As discussed, this has to be reviewed whether it's only applicable for fastrtps or all other vendors as well.


mikaelarguedas:
After some investigation with @Karsten1987 we concluded that this is not needed as long as people are passing the right typesupport structures (merging this will not solve the problem for more complex data structures/use cases). This can thus be closed.

Issue #36, opened 2016-06-03
Magic numbers in serialization


jacquelinekay:
https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/TypeSupport.cpp#L180\

        if(vector.size() > (member->is_upper_bound_ ? member->array_size_ : (typeByDefaultLarge() ? 30 : 101))) { \
            printf("vector overcomes the maximum length\n"); \
            throw std::runtime_error("vector overcomes the maximum length"); \
        } \

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/TypeSupport.cpp#L567

                case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
                    current_alignment += 4 + eprosima::fastcdr::Cdr::alignment(current_alignment, 4) + (member->string_upper_bound_ ? member->string_upper_bound_ + 1 : 257);
                    break;

I'm confused about the reasoning behind some of the "magic numbers" in these conditionals (30, 101, 257).

richiware:
Those "magic numbers" were a consequence of not having fragmentation in
FastRTPS. There are some meta-topics used in ROS2 whose types have a maximum
serialized payload larger than 65536. An examples is _ParameterEvent_. In IDL
if a sequence hasn't got a maximum length, it was set to 100. Those types has
several sequences and the calculated maximum serialized length is too large. I
made some calculations and for those types I set the default maximum length for
its sequences to 30.

Now FastRTPS has fragmentation and you can remove that limitation.

wjwwood:
@richiware I've been trying to remove these restrictions and test out our demos that send images. I am trying to do this by using the FastBuffer with internally managed memory (the default constructor) on publish/serialize:

```diff
diff --git a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
index eeac5f8..7e1bfdf 100644
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.h
[...]

PR #79, opened 2016-12-13
Default TTL to 0 and allow setting with env variable

wjwwood:
This pr will detect if Fast-RTPS has the feature that allows us to set the TTL for multicast UDP packets, and if it does it will default it to 0 (rather than 1 as it is now) and will allow users to control that setting using the `RMW_FASTRTPS_TTL_LEVEL` environment variable.

The idea behind making the default 0 instead of 1 is to avoid accidentally taking down WiFi networks when people run the camera demo. This will prevent people from using fast rtps across two computers unless they set the TTL to 1 on both machines using the environment variable.

In order for this pr to have any effect we need to either get this pr merged or carry the patch into a fork of Fast-RTPS which we use until it is merged: https://github.com/eProsima/Fast-RTPS/pull/68

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -42,6 +42,29 @@ ament_export_dependencies(rosidl_typesupport_introspection_cpp)
 
 include_directories(include)
 
+# try to compile some test code which uses the TTL field in the UDP
+# transport descriptions so that we can know if we can use it in this package
+try_compile(FAST_RTPS_HAS_TTL_IN_UDP_DESC
+  "${CMAKE_BINARY_DIR}/temp"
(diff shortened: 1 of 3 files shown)


wjwwood:
This comment is relevant:

>By default a subscriber uses a multicast address and a unicast address. Right now writer writes samples to both addresses. Next release will be more clever and if it is possible only write to one. I configured the subscriber not to use the default addresses, and only use a multicast address. Then the writer will write samples only to that address.
> -- https://github.com/ros2/rmw_fastrtps/issues/36#issuecomment-228753919

I'll try to adapt the logic he provided there to do the opposite and disable multicast all together in favor of just unicast until we can better control it in a new pr.

wjwwood:
Since the current version of Fast-RTPS has multicast disabled by default, I'll close this for now. We can do a better fix in the future when Fast-RTPS can automatically select between unicast and multicast.

PR #75, opened 2016-12-08
Export FastRTPS libraries

esteve:
This PR ensures that the FastRTPS and FastCDR libraries are exported to downstream packages.

Tested on OSX, before:

```sh
$ otool -L install_isolated/rclcpp_examples/bin/talker__rmw_fastrtps_cpp
install_isolated/rclcpp_examples/bin/talker__rmw_fastrtps_cpp:
        @rpath/librclcpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librcl.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librcl_interfaces__rosidl_typesupport_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librcl_interfaces__rosidl_typesupport_introspection_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librcl_interfaces__rosidl_generator_c.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librcl_interfaces__rosidl_typesupport_introspection_c.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librmw.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/librmw_fastrtps_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)
        @rpath/libstd_msgs__rosidl_typesupport_introspection_c.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libstd_msgs__rosidl_generator_c.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libstd_msgs__rosidl_typesupport_introspection_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libstd_msgs__rosidl_typesupport_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libbuiltin_interfaces__rosidl_typesupport_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libbuiltin_interfaces__rosidl_typesupport_introspection_cpp.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libbuiltin_interfaces__rosidl_generator_c.dylib (compatibility version 0.0.0, current version 0.0.0)
[...]

PR #82, opened 2016-12-20
rmw_fastrtps with C-only typesupport

esteve:
This RMW spins off the C-only typesupport so that it's easier to build ROS2 workspaces that only need typesupport for C, such language bindings, which are exclusively built on top of rcl (e.g. Android, see https://travis-ci.org/esteve/ros2_android/builds/185252405)

I also aggregated the C and C++ specializations code into separate structs (see https://github.com/esteve/rmw_fastrtps/blob/rmw-fastrtps-c/rmw_fastrtps_c/include/rmw_fastrtps_c/TypeSupport_c_impl.h and https://github.com/esteve/rmw_fastrtps/blob/rmw-fastrtps-c/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_cpp_impl.h)

The C-only RMW implementation is disabled by default, though.

Depends on https://github.com/ros2/rmw_fastrtps/pull/75


dirk-thomas:
We have recently suspended the support of two rmw implementations (Connext dynamic as well as OpenSplice). I am not sure if we want to add a new rmw implementation at this point.

esteve:
@dirk-thomas it's not a new implementation as much as a code reorganization, to be honest. In this PR, `rmw_fastrtps_cpp` piggybacks on `rmw_fastrtps_c`

dirk-thomas:
I do understand. But it is still one more rmw implementation to test as well as maintain in the future. I am just commenting with the rational why we reduced the number of supported rmw implementation.

esteve:
@dirk-thomas and that's why I clarified that this is not a new implementation, but just reorganizes existing code in a way so that it can be used in isolation.

wjwwood:
This is a bit confusing to me, since I assumed the `_cpp` in our rmw implementations referred to the language of the API we were using for that middleware, rather than the language of the messages being sent (ros messages or DDS message structures). For example, the only difference I would have imagined between `rmw_connext_cpp` and the theoretical `rmw_connext_c` would be that the former used the Connext C++ API (e.g. `DDS::Participant`) and the latter would use the Connext C API (e.g. `DDS_Participant_var`), but the ROS message type would be unchanged. So for me the `_cpp` in `rmw_fastrtps_cpp` indicated that we were using a C++ api from Fast-RTPS to implement `rmw`.

This idea of mine (probably incomplete or outdated) breaks down when you introduce type support introspection, as is the case for fastrtps. So there is message-type agnostic code for both C++ and C in `rmw_fastrtps_cpp`.

The diff is pretty hard to read, simply because it is so big, so it would help me to know/have:

- A summary of the new packages (and their purpose)
- A summary of the packages removed or repurposed
- A summary of the benefits of not building C++ type specialization in the Fast-RTPS rmw impl

----

I think what @dirk-thomas is talking about is that we build a copy of many things for each rmw implementation, for example `librcl`. We have a `librcl` (the default), `librcl__rmw_connext_cpp`, and `librcl__rmw_fastrtps_cpp`. We also have rmw impl specific builds for most of the tests, i.e. `test_foo__rmw_{fastrtps, connext}_cpp`. If we add another, say `rmw_fastrtps_c`, then we add a lot of new build artifacts and tests and therefore build time. So despite this just being a refactoring of the existing code, by creating a new rmw implementation there is a pretty big cost for our building time and testing time.

That being said, it should be weighed against the benefits of splitting it out (code size, build time, etc.), which is why I'd like a summary of the benefits from your perspective (beyond just "don't have to build c++ support").

Through all of this I'm assuming the end result is two rmw implementations where before there was just one: `rmw_fastrtps_cpp` and `rmw_fastrtps_c`.

I think you mentioned that `rmw_fastrtps_c` would be off by default, does that mean that when `rmw_fastrtps_c` is disabled, `rmw_fastrtps_cpp` still has support for both C and C++ ROS messages?

esteve:
@wjwwood thanks for taking the time to actually look at the diff.

> This is a bit confusing to me, since I assumed the _cpp in our rmw implementations referred to the language of the API we were using for that middleware, rather than the language of the messages being sent (ros messages or DDS message structures). For example, the only difference I would have imagined between rmw_connext_cpp and the theoretical rmw_connext_c would be that the former used the Connext C++ API (e.g. DDS::Participant) and the latter would use the Connext C API (e.g. DDS_Participant_var), but the ROS message type would be unchanged. So for me the _cpp in rmw_fastrtps_cpp indicated that we were using a C++ api from Fast-RTPS to implement rmw.
> 
> This idea of mine (probably incomplete or outdated) breaks down when you introduce type support introspection, as is the case for fastrtps. So there is message-type agnostic code for both C++ and C in rmw_fastrtps_cpp.

Yeah, I agree, the name wasn't the best choice, but couldn't think of anything better.

> The diff is pretty hard to read, simply because it is so big, so it would help me to know/have:
> 
>     A summary of the new packages (and their purpose)

- `rmw_fastrtps_common` contains all the code that does not interact with typesupport directly, so that there isn't anything specific to C or C++ typesupport
- `rmw_fastrtps_c` implements the `rmw` interface, but only supports C typesupport and exports the header with the C specialization.

>     A summary of the packages removed or repurposed

- `rmw_fastrtps_cpp` implements the `rmw` interface, builds on top of `rmw_fastrtps_c` to support C typesupport and additionally supports C++ typesupport, so it's functionally equivalent to the existing `rmw_fastrtps_cpp`

>     A summary of the benefits of not building C++ type specialization in the Fast-RTPS rmw impl

This basically boils down to workspaces / systems / robots that are built on top of `rcl` exclusively (e.g. Android, iOS, etc.), and which don't require C++ typesupport. In the case of Android, for example, adding C++ typesupport also requires a bunch of dependencies (e.g. Poco, pcre, etc.) that are either not easy to build or simply not available.

> I think you mentioned that rmw_fastrtps_c would be off by default, does that mean that when rmw_fastrtps_c is disabled, rmw_fastrtps_cpp still has support for both C and C++ ROS messages?

Yes, the `rmw_fastrtps_cpp` in this PR supports both C and C++, I kept it functionally the same so that these changes do not break anything. Additionally, I disabled compiling and registering `rmw_fastrtps_c` as an available RMW implementation (it only exports the C typesupport specialization) so that it does not interfere in any way with the existing code.

I've also submitted https://github.com/ros2/rmw_fastrtps/pull/83, which only does the code reorganization. I think the diff is more readable than this one, and if this PR can't be merged it makes it easier for me to just distribute patches that comment out the support for C++ typesupport.

wjwwood:
Ok, so it sounds like the main reason is to shed dependencies (I think Poco and pcre were mentioned).

So to recap, this pr:

- adds a new rmw implementation, but in the disabled state by default
- reuses the code in the new rmw implementation in the existing one so it is still tested
- avoids extra dependencies for users who only need C

Based on that I think this should be ok to pursue further. I still think we need to look at:

- can the dependencies be avoided in general (it's hard for me to imagine how poco and pcre are used in C++ but are not needed in C)
- is the mechanism for disabling and enabling each of the rmw implementations convenient and sufficient
- if there is a way to break this into stages, sortof like what you did in #83, to make it easier to review and merge
- the priority of this given the schedule conflicts with the related shared vs static library stuff
  - i.e. does it make sense to push this forward if the static library support is stalled waiting on the rmw impl selection refactoring

dirk-thomas:
As discussed in ament/ament_tools#128 `poco` will soon be required for the C typesupport too. At least until the optimization can be implemented to only support a single rmw implementation at compile time.

esteve:
@dirk-thomas thanks for tackling the `typesupport_c` refactor so soon, with that out of the way, I can look into getting it to work with Android and iOS.

Given that Poco is here to stay, I've submitted https://github.com/ros2/poco_vendor/pull/3 to enable building it for Android.

dirk-thomas:
@esteve Maybe you can respond to @wjwwood last comment and rebase the PR to resolve the conflicts.

dirk-thomas:
I will close this for now since there was no activity for a while and no response to the questions. Also as of #135 this patch is significantly diverged and would need a redo before it becomes mergable again.

esteve:
I forgot to close this one when I closed https://github.com/ros2/rmw_fastrtps/pull/83 Thanks for keeping an eye on it @dirk-thomas

PR #83, opened 2016-12-22
Refactor to isolate C and C++ specializations

esteve:
This moves the C and C++ specializations each to separate units, keeping the common code in one place.

The changes in this PR improve readability, so the general design of the code is abstracted from the C and C++ specializations.

Depends on #75 
Connect to #75

### DIFF
--- a/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
+++ b/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
@@ -64,13 +64,6 @@ endif()
 find_path(FastRTPS_INCLUDE_DIR
     NAMES fastrtps/)
 
-include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(FastRTPS
-  FOUND_VAR FastRTPS_FOUND
-  REQUIRED_VARS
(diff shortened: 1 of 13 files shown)


esteve:
@ros2/team I've rebased this and fixed he `-Wpedantic` warnings. I think the changes here are an improvement because they keep the C and C++-specific bits separate (this makes the code easier to read and more expandable) and keep the common code in one coherent place. These changes don't change the behavior of `rmw_fastrtps`, it's just a refactor.

dirk-thomas:
The diff is currently very hard to read. Can you please structure the changes into the following separate commits:

* copy existing file without modification to new filenames
* remove code from copied files
* update code to match moved code
* other refactorings
* apply non-essential changes (e.g. replace static string with variable name)

Please avoid any unnecessary whitespace related changes.

esteve:
@dirk-thomas that seems a bit of too much extra work in vain to be honest, especially because many of those changes happen at the same time, I can try though.

esteve:
I'm withdrawing this, I don't think I'll be able to work on it in the short term, so best not to pollute the queue with outdated PRs :-)

PR #88, opened 2017-01-26
expose list of nodes

Karsten1987:
Connect to ros2/cli_tools#1

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -2368,13 +2368,13 @@ rmw_destroy_topic_names_and_types(
 rmw_ret_t
 rmw_get_node_names(
   const rmw_node_t * node,
-  rmw_node_names_t * node_names)
+  rmw_string_array_t * node_names)
 {
   if (!node) {
     RMW_SET_ERROR_MSG("null node handle");
     return RMW_RET_ERROR;
   }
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -2366,6 +2366,49 @@ rmw_destroy_topic_names_and_types(
...
+  }
+
+  CustomParticipantInfo * impl = static_cast<CustomParticipantInfo *>(node->data);
+  Participant * participant = impl->participant;
+
+  auto participant_names = participant->getParticipantNames();
+  node_names->node_count = participant_names.size();
+  node_names->names = static_cast<char **>(rmw_allocate(node_names->node_count * sizeof(char *)));
+  for (size_t i = 0; i < participant_names.size(); ++i) {
+    node_names->names[i] = __local_strdup(participant_names[i].c_str());
```
Karsten1987: not sure I understand. 
Should we replace `strdup` with `rmw_allocate...` and `memcpy` ?

dirk-thomas: The assumption is that `rmw_allocate` should be able to provide a custom strategy (e.g. use memory from a preallocated pool to avoid any allocation at runtime). If any code directly allocates memory (like `strdup` does) it undermines that approach. E.g. this would immediately break any attempt to achieve realtime.

Therefore, yes, I think this should use `rmw_allocate` and `memcpy`.

PR #91, opened 2017-02-14
Fixed pointing to FASTRPCHOME instead of FASTRTPSHOME

### LINKED ISSUE #90: FindFastRTPS.cmake uses the env var FASTRPCHOME instead of FASTRTPSHOME
In FindFastRTPS.cmake:

if($ENV{FASTRPCHOME})
    set(FastRTPS_HOME $ENV{FASTRPCHOME})
endif()

This should of course be

if($ENV{FASTRTPSHOME})
    set(FastRTPS_HOME $ENV{FASTRTPSHOME})
endif()

jespersmith:
Fixes #90.

### DIFF
--- a/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
+++ b/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
@@ -48,8 +48,8 @@ set(FastRTPS_FOUND FALSE)
 # -----------------------------------------
 # Detect environment variable
 # -----------------------------------------
-if($ENV{FASTRPCHOME})
-    set(FastRTPS_HOME $ENV{FASTRPCHOME})
+if($ENV{FASTRTPSHOME})
+    set(FastRTPS_HOME $ENV{FASTRTPSHOME})
 endif()
 
 # -----------------------------------------


dirk-thomas:
@richiprosima Can you please comment on this since you wrote the snippet originally.

dirk-thomas:
@richiprosima Ping.

richiprosima:
When I started rmw_fastrtps_cpp I used rmw_opensplice_cpp as a guide. So I started to write the file FindFastRTPS.cmake. But then I decided to migrate FastRTPS to cmake build environment.

As I know, ament should use the configuration cmake files that FastRTPS generates, and the file FindFastRTPS.cmake should not be used, because it's an unfinished draft.

dirk-thomas:
@richiprosima Can you please also comment on the purpose of these two environment variables.

richiprosima:
Makefile files generated by *fastrtpsgen* use the environment variable FASTRTPSHOME to find the root directory of Fast RTPS installation on Windows OS.
We have another product called RPCDDS and it also has a code generator. This product uses the environment variable FASTRPCHOME.

dirk-thomas:
@richiprosima So just to clarify we don't need either of the two environment variables for ROS 2, correct? And if we would set them for convenience we should set both, right?

richiprosima:
You don't need to set any of the two environment variables. FASTRTPSHOME is used in Makefiles and Visual Studio solution generated by ```fastrtpsgen```. And as you are not using RPCDDS, forget the environment variable FASTRPCHOME.

dirk-thomas:
@richiprosima Thank you for clarifying.

@jespersmith Can you please update the PR to remove the usage of the environment variable.

jespersmith:
I think that's outside my scope and better be handled by people that know what they are doing.

dirk-thomas:
Closing in favor of #110.

PR #94, opened 2017-03-01
Added CDR encapsulation

richiware:
In future version of Fast RTPS, CDR encapsulation has to be done outside of the fastrtps library. This PR implements the CDR encapsulation on rmw_fastrtps_cpp.


mikaelarguedas:
thanks @richiware, I'll give it a try today

PR #97, opened 2017-03-29
update to support node namespace in rmw API

wjwwood:
Connects to ros2/rmw#95

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -632,48 +632,54 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t 
 
   eprosima::fastrtps::Log::SetVerbosity(eprosima::fastrtps::Log::Error);
 
+  // Declare everything before beginning to create things.
+  Participant * participant = nullptr;
+  rmw_guard_condition_t * graph_guard_condition = nullptr;
+  CustomParticipantInfo * node_impl = nullptr;
+  rmw_node_t * node_handle = nullptr;
+  topicnamesandtypesReaderListener * tnat_1 = nullptr;
+  topicnamesandtypesReaderListener * tnat_2 = nullptr;
+  std::pair<StatefulReader *, StatefulReader *> EDPReaders;
+
   ParticipantAttributes participantParam;
   participantParam.rtps.builtin.domainId = static_cast<uint32_t>(domain_id);
   participantParam.rtps.setName(name);
 
-  Participant * participant = Domain::createParticipant(participantParam);
+  participant = Domain::createParticipant(participantParam);
   if (!participant) {
     RMW_SET_ERROR_MSG("create_node() could not create participant");
     return NULL;
   }
 
-  rmw_guard_condition_t * graph_guard_condition = rmw_create_guard_condition();
+  graph_guard_condition = rmw_create_guard_condition();
   if (!graph_guard_condition) {
     // error already set
-    return NULL;
+    goto fail;
   }
 
-  CustomParticipantInfo * node_impl = nullptr;
   try {
     node_impl = new CustomParticipantInfo();
   } catch (std::bad_alloc) {
     RMW_SET_ERROR_MSG("failed to allocate node impl struct");
-    return NULL;
+    goto fail;
   }
 
-  rmw_node_t * node_handle =
-    static_cast<rmw_node_t *>(malloc(sizeof(rmw_node_t)));
+  node_handle = static_cast<rmw_node_t *>(malloc(sizeof(rmw_node_t)));
   if (!node_handle) {
     RMW_SET_ERROR_MSG("failed to allocate rmw_node_t");
-    return NULL;
+    goto fail;
   }
   node_handle->implementation_identifier = eprosima_fastrtps_identifier;
   node_impl->participant = participant;
   node_impl->graph_guard_condition = graph_guard_condition;
   node_handle->data = node_impl;
 
-
   node_handle->name =
     static_cast<const char *>(malloc(sizeof(char) * strlen(name) + 1));
   if (!node_handle->name) {
     RMW_SET_ERROR_MSG("failed to allocate memory");
-    free(static_cast<void *>(node_handle));
-    return NULL;
+    node_handle->namespace_ = nullptr;  // to avoid free on uninitialized memory
+    goto fail;
   }
   memcpy(const_cast<char *>(node_handle->name), name, strlen(name) + 1);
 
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -618,13 +618,18 @@ rmw_ret_t rmw_init()
...
 
-rmw_node_t * rmw_create_node(const char * name, size_t domain_id)
+rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t domain_id)
 {
   if (!name) {
     RMW_SET_ERROR_MSG("name is null");
     return NULL;
   }
 
+  if (!namespace_) {
```
Karsten1987: do we actively enforce users to call this function with `""` instead of NULL?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -672,6 +677,16 @@ rmw_node_t * rmw_create_node(const char * name, size_t domain_id)
   }
   memcpy(const_cast<char *>(node_handle->name), name, strlen(name) + 1);
 
+  node_handle->namespace_ =
+    static_cast<const char *>(malloc(sizeof(char) * strlen(namespace_) + 1));
```
Karsten1987: should this be `rmw_allocate`?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -672,6 +677,16 @@ rmw_node_t * rmw_create_node(const char * name, size_t domain_id)
...
+  if (!node_handle->namespace_) {
+    RMW_SET_ERROR_MSG("failed to allocate memory");
+    free(const_cast<char *>(node_handle->name));
```
Karsten1987: linked to above comment: `rmw_free` ?

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: Yeah, my assumption is that it is a valid string at this point. See a test for this in `rcl`:

https://github.com/ros2/rcl/pull/112/files#diff-56a2b9e98ab73d202a5f841102b7c54bR326

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: Yeah, I'll change it. However, `rmw_fastrtps` uses malloc in lots of places (since we didn't write it originally). Also the idea behind `rmw_allocate` is sort of a dead end in my mind. We should instead update `rmw` to use the allocator structure like `rcl` does.

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: Also, `node_handle->namespace_` 😛, thanks for finding two bugs lol.

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: Actually, since the rest of the code around it uses `malloc`, I'm going to leave it for consistency. At some point we should audit memory allocation in Fast-RTPS.

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: Also, this is not a bug, it is just undoing `->name` since we need to exit here. Since `->name` was allocated with `malloc` I'll continue to use `free` here.

wjwwood:
After discussing with @Karsten1987 in person, I refactored some of the code I was touching in `rmw_create_node` and ended up fixing several memory management issues in `rmw_destroy_node` as well: 983ea01

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -618,81 +618,111 @@ rmw_ret_t rmw_init()
...
   eprosima::fastrtps::Log::SetVerbosity(eprosima::fastrtps::Log::Error);
 
+  // Declare everything before beginning to create things.
+  Participant * participant = nullptr;
+  rmw_guard_condition_t * graph_guard_condition = nullptr;
+  CustomParticipantInfo * node_impl = nullptr;
+  rmw_node_t * node_handle = nullptr;
+  topicnamesandtypesReaderListener * tnat_1 = nullptr;
+  topicnamesandtypesReaderListener * tnat_2 = nullptr;
+  std::pair<StatefulReader *, StatefulReader *> EDPReaders;
```
dirk-thomas: Variable names shouldn't contain upper case characters: `edp_readers`.

Same below.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -716,12 +746,26 @@ rmw_ret_t rmw_destroy_node(rmw_node_t * node)
...
+  delete impl->secondarySubListener;
+  if (!EDPReaders.second->setListener(nullptr)) {
+    RMW_SET_ERROR_MSG("failed to unset EDPReader listener");
+    return RMW_RET_ERROR;
+  }
+  delete impl->secondaryPubListener;
 
+  if (node) {
+    free(const_cast<char *>(node->name));
+    free(const_cast<char *>(node->namespace_));
```
dirk-thomas: Should these two lines continue to set the pointer to `nullptr`? It often helps when debugging when they are actually reset before the parent is deallocated.

wjwwood: But `node` is deallocated on the next line, so I didn't. I don't know what debugging will happen between here and there.

Either way: But `node` is deallocated on the next line, so I didn't. I don't know what debugging will happen between here and there.

Either way: 79c4860

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: 6f5a012

dirk-thomas: Still the same below: https://github.com/ros2/rmw_fastrtps/pull/97/files#diff-8225b098f1ec4debcdcee03d65ae95d5R752

rmw_fastrtps_cpp/src/functions.cpp
dirk-thomas: We do this in existing code for debugging purposed. If the caller passes a pointer and you watch it in the debugger you will see that the member are actually reset to a `nullptr` rather than still contain a memory address.

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: ef810bc

PR #100, opened 2017-04-07
namespace implementation

Karsten1987:
connects to ros2/ros2#327

working implementation.

What's missing:
- Cleanup memory management in the style of: https://github.com/ros2/rmw_fastrtps/pull/97/commits/983ea01a2f488a31711a0b6dfefa1d358ecdc2ee
- Figure out why the two postfixes "Reply" and "Request" are necessary. Without these, the services won't match.

### DIFF
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -22,9 +22,9 @@
   <build_depend>rosidl_typesupport_introspection_c</build_depend>
   <build_depend>rosidl_typesupport_introspection_cpp</build_depend>
 
-  <build_export_depend>rcutils</build_export_depend>
   <build_export_depend>fastcdr</build_export_depend>
   <build_export_depend>fastrtps</build_export_depend>
+  <build_export_depend>rcutils</build_export_depend>
   <build_export_depend>rmw</build_export_depend>
   <build_export_depend>rosidl_generator_c</build_export_depend>
   <build_export_depend>rosidl_generator_cpp</build_export_depend>
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -22,7 +22,7 @@
   <build_depend>rosidl_typesupport_introspection_c</build_depend>
   <build_depend>rosidl_typesupport_introspection_cpp</build_depend>
 
-  <build_export_depend>c_utilities</build_export_depend>
+  <build_export_depend>rcutils</build_export_depend>
   <build_export_depend>fastcdr</build_export_depend>
   <build_export_depend>fastrtps</build_export_depend>
   <build_export_depend>rmw</build_export_depend>
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -22,6 +22,7 @@
   <build_depend>rosidl_typesupport_introspection_c</build_depend>
   <build_depend>rosidl_typesupport_introspection_cpp</build_depend>
 
+  <build_export_depend>c_utilities</build_export_depend>
   <build_export_depend>fastcdr</build_export_depend>
   <build_export_depend>fastrtps</build_export_depend>
   <build_export_depend>rmw</build_export_depend>
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -878,11 +878,14 @@ rmw_publisher_t * rmw_create_publisher(const rmw_node_t * node,
     RMW_SET_ERROR_MSG("Illformated topic name");
     goto fail;
   }
-  utilities_string_array_fini(&name_tokens);
 
-  fprintf(stderr, "ros topic: %s\n", topic_name);
-  fprintf(stderr, "partition name: %s\n", publisherParam.qos.m_partition.getNames()[0].c_str());
-  fprintf(stderr, "dds topic: %s\n", publisherParam.topic.topicName.c_str());
+  if (utilities_string_array_fini(&name_tokens) != UTILITIES_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
+
+  // fprintf(stderr, "ros topic: %s\n", topic_name);
+  // fprintf(stderr, "partition name: %s\n", publisherParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "dds topic: %s\n", publisherParam.topic.topicName.c_str());
 
   // 1 Heartbeat every 10ms
   // publisherParam.times.heartbeatPeriod.seconds = 0;
@@ -939,7 +942,9 @@ fail:
     rmw_publisher_free(rmw_publisher);
   }
 
-  utilities_string_array_fini(&name_tokens);
+  if (utilities_string_array_fini(&name_tokens) != UTILITIES_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
 
   return NULL;
 }
@@ -1183,11 +1188,13 @@ rmw_subscription_t * rmw_create_subscription(const rmw_node_t * node,
     RMW_SET_ERROR_MSG("Illformated topic name");
     goto fail;
   }
-  utilities_string_array_fini(&name_tokens);
+  if (utilities_string_array_fini(&name_tokens) != UTILITIES_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
 
-  fprintf(stderr, "ros topic: %s\n", topic_name);
-  fprintf(stderr, "partition name: %s\n", subscriberParam.qos.m_partition.getNames()[0].c_str());
-  fprintf(stderr, "dds topic: %s\n", subscriberParam.topic.topicName.c_str());
+  // fprintf(stderr, "ros topic: %s\n", topic_name);
+  // fprintf(stderr, "partition name: %s\n", subscriberParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "dds topic: %s\n", subscriberParam.topic.topicName.c_str());
 
   if (!get_datareader_qos(*qos_policies, subscriberParam)) {
     RMW_SET_ERROR_MSG("failed to get datareader qos");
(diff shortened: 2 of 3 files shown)


Karsten1987:
So I couldn't really figure out why `Reply` and `Request` has to be there.
I tried to replace it with `tacoReplybell` and `tacoRequestBell` and that seems to work as well, however only `taco` and `bell` doesn't. I'll ignore this change for now as it is not a stopper for this PR.

rmw_fastrtps_cpp/package.xml
```diff
@@ -22,6 +22,7 @@
   <build_depend>rosidl_typesupport_introspection_c</build_depend>
   <build_depend>rosidl_typesupport_introspection_cpp</build_depend>
 
+  <build_export_depend>rcutils</build_export_depend>
```
wjwwood: nitpick: alphabetical ordering

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -837,12 +861,31 @@ rmw_publisher_t * rmw_create_publisher(const rmw_node_t * node,
...
+    goto fail;
+  }
+
+  if (rcutils_string_array_fini(&name_tokens) != RCUTILS_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
+
+  // fprintf(stderr, "ros topic: %s\n", topic_name);
+  // fprintf(stderr, "partition name: %s\n", publisherParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "dds topic: %s\n", publisherParam.topic.topicName.c_str());
```
wjwwood: Please remove these or put a TODO describing what ought to be done with them in the future.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -837,12 +861,31 @@ rmw_publisher_t * rmw_create_publisher(const rmw_node_t * node,
...
+  name_tokens = rcutils_split_last(topic_name, '/');
+  if (name_tokens.size == 1) {
+    publisherParam.qos.m_partition.push_back(ros_topics_prefix);
+    publisherParam.topic.topicName = name_tokens.data[0];
+  } else if (name_tokens.size == 2) {
+    publisherParam.qos.m_partition.push_back(
+      (std::string(ros_topics_prefix) + "/" + name_tokens.data[0]).c_str());
+    publisherParam.topic.topicName = name_tokens.data[1];
+  } else {
+    RMW_SET_ERROR_MSG("Illformated topic name");
```
wjwwood: Either `Ill-formed` or `malformed` is more proper English I think.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1096,13 +1172,32 @@ rmw_subscription_t * rmw_create_subscription(const rmw_node_t * node,
...
+  name_tokens = rcutils_split_last(topic_name, '/');
+  if (name_tokens.size == 1) {
+    subscriberParam.qos.m_partition.push_back(ros_topics_prefix);
+    subscriberParam.topic.topicName = name_tokens.data[0];
+  } else if (name_tokens.size == 2) {
+    subscriberParam.qos.m_partition.push_back(
+      (std::string(ros_topics_prefix) + "/" + name_tokens.data[0]).c_str());
+    subscriberParam.topic.topicName = name_tokens.data[1];
+  } else {
+    RMW_SET_ERROR_MSG("Illformated topic name");
```
wjwwood: Same comment as above.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1096,13 +1172,32 @@ rmw_subscription_t * rmw_create_subscription(const rmw_node_t * node,
...
+    goto fail;
+  }
+  if (rcutils_string_array_fini(&name_tokens) != RCUTILS_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
```
wjwwood: This whole block seems to be shared between create publisher and create subscriber. Perhaps you should make a shared function which could be improved and tested in one place.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1096,13 +1172,32 @@ rmw_subscription_t * rmw_create_subscription(const rmw_node_t * node,
...
+
+  // fprintf(stderr, "ros topic: %s\n", topic_name);
+  // fprintf(stderr, "partition name: %s\n", subscriberParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "dds topic: %s\n", subscriberParam.topic.topicName.c_str());
```
wjwwood: Same here, TODO or remove.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1596,57 +1719,86 @@ rmw_client_t * rmw_create_client(const rmw_node_t * node,
...
+    publisherParam.topic.topicName = name_tokens.data[0];
+  } else if (name_tokens.size == 2) {
+    subscriberParam.qos.m_partition.push_back(
+      (std::string(ros_service_response_prefix) + "/" + name_tokens.data[0]).c_str());
+    subscriberParam.topic.topicName = name_tokens.data[1];
+    publisherParam.qos.m_partition.push_back(
+      (std::string(ros_service_requester_prefix) + "/" + name_tokens.data[0]).c_str());
+    publisherParam.topic.topicName = name_tokens.data[1];
+  } else {
+    RMW_SET_ERROR_MSG("Illformated service name");
```
wjwwood: Same spelling nitpick.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1596,57 +1719,86 @@ rmw_client_t * rmw_create_client(const rmw_node_t * node,
...
-  info->listener_ = new ClientListener(info);
-  info->response_subscriber_ = Domain::createSubscriber(participant, subscriberParam,
-      info->listener_);
+  // fprintf(stderr, "Original topic: %s\n", service_name);
+  // fprintf(stderr, "Pub Partition name: %s\n",
+  //   publisherParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "Sub Partition name: %s\n",
+  //   subscriberParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "Pub New topic: %s\n", publisherParam.topic.topicName.c_str());
+  // fprintf(stderr, "Sub New topic: %s\n", subscriberParam.topic.topicName.c_str());
```
wjwwood: Remove or TODO.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1909,83 +2084,117 @@ rmw_service_t * rmw_create_service(const rmw_node_t * node,
...
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
+
+  // fprintf(stderr, "Original topic: %s\n", service_name);
+  // fprintf(stderr, "Pub Partition name: %s\n",
+  //   publisherParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "Sub Partition name: %s\n",
+  //   subscriberParam.qos.m_partition.getNames()[0].c_str());
+  // fprintf(stderr, "Pub New topic: %s\n", publisherParam.topic.topicName.c_str());
+  // fprintf(stderr, "Sub New topic: %s\n", subscriberParam.topic.topicName.c_str());
```
wjwwood: Remove or TODO.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1909,83 +2084,117 @@ rmw_service_t * rmw_create_service(const rmw_node_t * node,
...
+    publisherParam.topic.topicName = name_tokens.data[0];
+  } else if (name_tokens.size == 2) {
+    subscriberParam.qos.m_partition.push_back(
+      (std::string(ros_service_requester_prefix) + "/" + name_tokens.data[0]).c_str());
+    subscriberParam.topic.topicName = name_tokens.data[1];
+    publisherParam.qos.m_partition.push_back(
+      (std::string(ros_service_response_prefix) + "/" + name_tokens.data[0]).c_str());
+    publisherParam.topic.topicName = name_tokens.data[1];
+  } else {
+    RMW_SET_ERROR_MSG("Illformated service name");
```
wjwwood: spelling nitpick

wjwwood:
I fixed a few in my own commits. What's left for me is the comments that should be removed or have a TODO added to them and the potential to refactor out the topic name splitting into a function.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -365,8 +400,14 @@ class topicnamesandtypesReaderListener : public ReaderListener
       tempMsg.length = change->serializedPayload.length;
       memcpy(tempMsg.buffer, change->serializedPayload.data, tempMsg.length);
       if (proxyData.readFromCDRMessage(&tempMsg)) {
+        auto partition_str = std::string("");
+        // don't use std::accumulate - schlemiel O(n2)
+        for (const auto & partition : proxyData.m_qos.m_partition.getNames()) {
```
wjwwood: @Karsten1987 this should be removed/replace with checking for exactly one partition which contains the ros specific prefix, right?

I'll add a TODO while rebasing (https://github.com/ros2/rmw_fastrtps/pull/109 caused a conflict here)

Karsten1987: I implemented the same logic in `rmw_connext` given that the same code should be used for non-ros dds topic which may use multiple partitions. 
https://github.com/ros2/rmw_connext/blob/namespace_impl/rmw_connext_shared_cpp/src/shared_functions.cpp#L114-L117

The check for partition size 1 is implemented before that call:
https://github.com/ros2/rmw_fastrtps/pull/100/files#diff-8225b098f1ec4debcdcee03d65ae95d5R2779

wjwwood:
I think we're sorting this out already with @clalancette's reports about the bridge, but the linux packaging job has been flakey since this was merged:

http://ci.ros2.org/job/packaging_linux-aarch64/48/ (starting May 19th)

We'll see if https://github.com/ros2/rmw_fastrtps/pull/114 and the related pr's fixes the packaging job tonight or tomorrow night.

PR #101, opened 2017-04-11
Cherry pick sros2 update for Fast-RTPS master support

nuclearsandwich:
This cherry picks a330d0e from the sros2 branch (with a style fixup) at the suggestion of @mikaelarguedas in order to support the current master branch of eProsima/Fast-RTPS.

Combined with https://github.com/ros2/ros2/pull/332 this will enable the update of Fast-RTPS to upstream's master once some local CMake changes are accepted upstream.


mikaelarguedas:
I think this should be merged along the PR tracking  Fast-RTPS master branch. As pointed out by @dirk-thomas I'm not sure this change will work with the current commit hash we are using (trying it locally right now and will post results here)

mikaelarguedas:
cool, thanks for checking

PR #103, opened 2017-04-18
DDS Security support

mikaelarguedas:
connects to ros2/ros2#339

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -985,11 +985,9 @@ rmw_create_node(
       property_policy.properties().emplace_back(
         Property("rtps.participant.rtps_protection_kind", "ENCRYPT"));
       participantAttrs.rtps.properties = property_policy;
-    } else {
-      if (options->enforce_security) {
-        RMW_SET_ERROR_MSG("couldn't find all security files!");
-        return NULL;
-      }
+    } else if (options->enforce_security) {
+      RMW_SET_ERROR_MSG("couldn't find all security files!");
+      return NULL;
     }
 #else
     RMW_SET_ERROR_MSG(
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -928,7 +928,7 @@ fail:
   return NULL;
 }
 
-bool rmw_get_security_file_paths(
+bool get_security_file_paths(
   std::array<std::string, 3> & security_files_paths, const char * node_secure_root)
 {
   // here assume only 3 files for security
@@ -973,7 +973,7 @@ rmw_create_node(
 #if HAVE_SECURITY
     std::array<std::string, 3> security_files_paths;
 
-    if (rmw_get_security_file_paths(security_files_paths, security_options->security_root_path)) {
+    if (get_security_file_paths(security_files_paths, security_options->security_root_path)) {
       PropertyPolicy property_policy;
       property_policy.properties().emplace_back(
         Property("dds.sec.auth.plugin", "builtin.PKI-DH"));
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -959,6 +959,10 @@ rmw_create_node(
     RMW_SET_ERROR_MSG("name is null");
     return NULL;
   }
+  if (!security_options) {
+    RMW_SET_ERROR_MSG("security_options is null");
+    return nullptr;
+  }
 
   ParticipantAttributes participantAttrs;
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -686,7 +688,10 @@ rmw_ret_t rmw_init()
   return RMW_RET_OK;
 }
 
-rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t domain_id)
+rmw_node_t * create_node(
+  const char * name,
+  const char * namespace_,
+  ParticipantAttributes participantParam)
```
dirk-thomas: Variable naming.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -796,6 +794,81 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t
...
+      property_policy.properties().emplace_back(
+        Property("dds.sec.auth.builtin.PKI-DH.private_key",
+        security_files_paths[2]));
+      property_policy.properties().emplace_back(
+        Property("dds.sec.crypto.plugin", "builtin.AES-GCM-GMAC"));
+      property_policy.properties().emplace_back(
+        Property("rtps.participant.rtps_protection_kind", "ENCRYPT"));
+      participantParam.rtps.properties = property_policy;
+    } else {
+      if (options->enforce_security) {
```
dirk-thomas: `elseif`?

mikaelarguedas:
> What is the HAVE_SECURITY define needed for ? How is it being set?

This is defined by fast-RTPS in [config.h](https://github.com/ros2/rmw_fastrtps/pull/103/files#diff-8225b098f1ec4debcdcee03d65ae95d5R37) at compile time. We need to check it because people not building fastrtps with `SECURITY=ON` will not have the security headers and symbols available causing rmw_fastrtps_cpp to not compile.

rmw_fastrtps_cpp/src/functions.cpp
mikaelarguedas: good catch, just copy-n-paste the original variable without renaming it. Will fix

rmw_fastrtps_cpp/src/functions.cpp
mikaelarguedas: thanks, another simplification I forgot to refactor, will fix

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -821,22 +826,15 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t
     return NULL;
   }
 
-  eprosima::fastrtps::Log::SetVerbosity(eprosima::fastrtps::Log::Error);
```
dirk-thomas: Why has this been removed?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -919,6 +917,79 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t
   return NULL;
 }
 
+bool rmw_get_security_file_paths(
+  std::array<std::string, 3> & security_files_paths, const char * node_secure_root)
+{
+  // here assume only 3 files for security
+  const char * file_names[3] = {"ca.cert.pem", "cert.pem", "key.pem"};
+  size_t num_files = 3;
```
dirk-thomas: Could this be `sizeof(...)` instead?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -919,6 +917,79 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t
   return NULL;
 }
 
+bool rmw_get_security_file_paths(
```
dirk-thomas: This function is not part of the rmw interface and should therefore not have the `rmw_` prefix.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -919,6 +917,79 @@ rmw_node_t * rmw_create_node(const char * name, const char * namespace_, size_t
...
+  if (!name) {
+    RMW_SET_ERROR_MSG("name is null");
+    return NULL;
+  }
+
+  ParticipantAttributes participantAttrs;
+  participantAttrs.rtps.builtin.domainId = static_cast<uint32_t>(domain_id);
+  participantAttrs.rtps.setName(name);
+
+  if (security_options->security_root_path) {
```
dirk-thomas: You must check that `security_options` is not null before accessing its members.

mikaelarguedas: done in https://github.com/ros2/rmw_fastrtps/pull/103/commits/4d4d8ff054e1d0c8036ea62ce34d016807f78d08

PR #108, opened 2017-05-15
fix unnecessary wake ups

dirk-thomas:
As a result of #51 the guard condition was constantly been triggered since `onNewCacheChangeAdded` is being invoked repeatedly for the same topic / type.

[Build Status](http://ci.ros2.org/job/ci_linux/2598/) using `--retest-until-fail 20` (all failing tests are also present in the corresponding nighlty)

Ready for review.

PR #109, opened 2017-05-15
support removing topics types

dirk-thomas:
This builds on top of #108.

While this patch implements the necessary logic in `rmw_fastrtps_cpp` I never saw that the code path actually got triggered. @richiprosima Maybe you can shed some light into this. When should `onNewCacheChangeAdded` be invoked with a `kind` not equal to `ALIVE`? Currently I don't see it being invoked (hence topics / types are never removed despite calling `rmw_destroy_publisher`).

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -357,29 +357,41 @@ public:
   void onNewCacheChangeAdded(RTPSReader * reader, const CacheChange_t * const change)
   {
     (void)reader;
-    if (change->kind == ALIVE) {
-      WriterProxyData proxyData;
-      CDRMessage_t tempMsg;
-      tempMsg.msg_endian = change->serializedPayload.encapsulation ==
-        PL_CDR_BE ? BIGEND : LITTLEEND;
-      tempMsg.length = change->serializedPayload.length;
-      memcpy(tempMsg.buffer, change->serializedPayload.data, tempMsg.length);
-      if (proxyData.readFromCDRMessage(&tempMsg)) {
-        mapmutex.lock();
-        auto it = topicNtypes.find(proxyData.topicName());
+    WriterProxyData proxyData;
+    CDRMessage_t tempMsg;
+    tempMsg.msg_endian = change->serializedPayload.encapsulation ==
+      PL_CDR_BE ? BIGEND : LITTLEEND;
+    tempMsg.length = change->serializedPayload.length;
+    memcpy(tempMsg.buffer, change->serializedPayload.data, tempMsg.length);
+    if (proxyData.readFromCDRMessage(&tempMsg)) {
+      bool mapModified = false;
+      mapmutex.lock();
+      auto it = topicNtypes.find(proxyData.topicName());
+      if (change->kind == ALIVE) {
         if (
           it == topicNtypes.end() ||
           it->second.find(proxyData.typeName()) == it->second.end())
         {
           topicNtypes[proxyData.topicName()].insert(proxyData.typeName());
-          rmw_ret_t ret = rmw_trigger_guard_condition(graph_guard_condition_);
-          if (ret != RMW_RET_OK) {
-            fprintf(stderr, "failed to trigger graph guard condition: %s\n",
-              rmw_get_error_string_safe());
-          }
+          mapModified = true;
         }
-        mapmutex.unlock();
+      } else {
+        if (
+          it != topicNtypes.end() &&
+          it->second.find(proxyData.typeName()) != it->second.end())
+        {
+          topicNtypes[proxyData.topicName()].erase(proxyData.typeName());
+          mapModified = true;
+        }
       }
+      if (mapModified) {
+        rmw_ret_t ret = rmw_trigger_guard_condition(graph_guard_condition_);
+        if (ret != RMW_RET_OK) {
+          fprintf(stderr, "failed to trigger graph guard condition: %s\n",
+            rmw_get_error_string_safe());
+        }
+      }
+      mapmutex.unlock();
     }
   }
   std::map<std::string, std::set<std::string>> topicNtypes;


dirk-thomas:
While the patch doesn't "make it work" yet I will merge it as is since we need this logic anyway.

@richiprosima If you could still comment on this to figure out how we get it working that we receive notifications of removed entities that would be good.

richiprosima:
User callback was not called when the information is about a removed entity. I made some changes to fix this and testing in our CI.

richiprosima:
I've fixed the problem. See eProsima/Fast-RTPS#105.

dirk-thomas:
I have tried the latest version but it still fails: http://ci.ros2.org/job/ci_linux/2605/testReport/junit/(root)/projectroot/test_graph__rmw_fastrtps_cpp/ Locally I added some debug statements and `onNewCacheChangeAdded` is still not called with anything but `ALIVE`.

richiware:
This [test](https://github.com/eProsima/Fast-RTPS/blob/6322cb9e875f685e9f68619143ded9374765ecb9/test/blackbox/BlackboxTests.cpp#L1728-L1752) is testing that. Tomorrow I will try to execute your ```test_graph__rmw_fastrtps_cpp```.

richiprosima:
I've added a message when receiving a not ```ALIVE``` ```CacheChange_t```:

```
  void onNewCacheChangeAdded(RTPSReader * reader, const CacheChange_t * const change)
  { 
    (void)reader;
    if (change->kind == ALIVE) {
      WriterProxyData proxyData;
      CDRMessage_t tempMsg;
      tempMsg.msg_endian = change->serializedPayload.encapsulation ==
        PL_CDR_BE ? BIGEND : LITTLEEND;
      tempMsg.length = change->serializedPayload.length;
      memcpy(tempMsg.buffer, change->serializedPayload.data, tempMsg.length);
      if (proxyData.readFromCDRMessage(&tempMsg)) {
        mapmutex.lock();
        topicNtypes[proxyData.topicName()].insert(proxyData.typeName());
        rmw_ret_t ret = rmw_trigger_guard_condition(graph_guard_condition_);
        if (ret != RMW_RET_OK) {
          fprintf(stderr, "failed to trigger graph guard condition: %s\n",
            rmw_get_error_string_safe());
        }
        mapmutex.unlock();
      }
    }
    else
        std::cout << "CALLED WHEN A ENTITY DESTROYED" << std::endl;
  }
```

The output of ```test_graph__rmw_fastrtps_cpp``` shows the message several times:

```
[ RUN      ] TestGraphFixture__rmw_fastrtps_cpp.test_graph_guard_condition_topics
starting memory checking...
stopping memory checking...
waiting up to '200000000' nanoseconds for graph changes
waiting up to '200000000' nanoseconds for graph changes
waiting up to '200000000' nanoseconds for graph changes
CALLED WHEN A ENTITY DESTROYED
waiting up to '200000000' nanoseconds for graph changes
CALLED WHEN A ENTITY DESTROYED
/home/ricardo/ros2_ws/src/ros2/rcl/rcl/test/rcl/test_graph.cpp:473: Failure
Expected: (graph_changes_count) >= (4ul), actual: 2 vs 4
[  FAILED  ] TestGraphFixture__rmw_fastrtps_cpp.test_graph_guard_condition_topics (622 ms)
```

dirk-thomas:
Hm, this is weird :worried: I have tried the same thing (see https://github.com/ros2/rmw_fastrtps/commit/4fd3232777dfa382bf90011236da56bb547cdf21). Yesterday I ran it locally and didn't see any output but `ALIVE`. I just ran it again but this time on the buildfarm to ensure I didn't miss anything locally but the output is the same: http://ci.ros2.org/job/ci_linux/2615/testReport/junit/(root)/projectroot/test_graph__rmw_fastrtps_cpp/ (and the build was using https://github.com/eProsima/Fast-RTPS/commit/6322cb9e875f685e9f68619143ded9374765ecb9 which should contain the changes you made in FastRTPS).

Can you please clarify what exact steps you ran when it worked for you.

dirk-thomas:
I moved the debug output above the `readFromCDRMessage` condition and `onNewCacheChangeAdded` **is** being called with `kind = 3` (see https://github.com/ros2/rmw_fastrtps/commit/ad4760c14e326885de793b86c9fff8c590ddca3f). Maybe you have an idea why `readFromCDRMessage` fails (which we need to pass in order to access the topic and type names.

richiprosima:
```readFromCDRMessage``` is failing because it has to only be called when ```ALIVE``` I suggest you to check this [code](https://github.com/eProsima/Fast-RTPS/blob/6322cb9e875f685e9f68619143ded9374765ecb9/test/blackbox/PubSubWriter.hpp#L93-L181) used in our tests.

dirk-thomas:
I am sorry @richiprosima I didn't notice `change_in->instanceHandle` when you pointed me to the test before.

* I updated our code to use that rather than trying to deserialize the not alive message. The updated implementation is on this branch: https://github.com/ros2/rmw_fastrtps/commit/e6924a0712e1ec60cd8a4fb2b048d01e6e0b2b9f
* I have also updated ros2/rcl#120 so that it continues to work with the latest master branches (since the namespace PRs have been merged yesterday).

When running the test now one of the nodes receives the callback for the not-alive topic/type and removes the information accordingly. But the second node never gets that information (and therefore still fails the test). Please see the [current test output](http://ci.ros2.org/job/ci_linux/2623/testReport/junit/(root)/projectroot/test_graph__rmw_fastrtps_cpp/) and look for "WriterInfo" with "insert(" vs. "erase(". It is being inserted twice (once for each node/participant) but only erased once. Therefore one of the nodes doesn't have the information and still thinks one publisher is still around. I am not sure if I have to do something differently? Any hint would be highly appreciated :bowing_man:

richiprosima:
Can you check again using commit eProsima/Fast-RTPS@5907c42a268fe35586d8822347356ff72718c76c?

dirk-thomas:
I just ran CI for this and it passes on Linux: http://ci.ros2.org/job/ci_linux/2646/testReport/(root)/TestGraphFixture__rmw_fastrtps_cpp/

I will follow up on this tomorrow to check if all platforms pass now. Thank you for the fix in FastRTPS.

PR #111, opened 2017-05-19
fix removal of topics / types

dirk-thomas:
This is WIP.

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -46,6 +46,7 @@
 #include "fastrtps/attributes/SubscriberAttributes.h"
 
 #include "fastrtps/rtps/RTPSDomain.h"
+#include "fastrtps/rtps/builtin/data/ReaderProxyData.h"
 #include "fastrtps/rtps/builtin/data/WriterProxyData.h"
 #include "fastrtps/rtps/common/CDRMessage_t.h"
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Current CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/204/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2105/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2768/)

Please see the [TODO in the patch](https://github.com/ros2/rmw_fastrtps/pull/111/files#diff-8225b098f1ec4debcdcee03d65ae95d5R446) about the partition information not being available for the deregistration. How should we address that?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -381,66 +382,177 @@ typedef struct CustomClientResponse
...
+      if (participant_->get_remote_reader_info(readerGuid, proxyData)) {
+        // TODO(wjwwood): remove this logic and replace with check for
+        // a single partition which is prefixed with the ROS specific
+        // prefix.
+        auto partition_str = std::string("");
+        // don't use std::accumulate - schlemiel O(n2)
+        for (const auto & partition : proxyData.m_qos.m_partition.getNames()) {
+          partition_str += partition;
+        }
+        partition_str = "rt";  // TODO(dirk-thomas): we don't have the partition info here
```
wjwwood: Why don't we have the information, is it just something that Fast-RTPS doesn't give us right now but could, or is technically/conceptually not possible?

An alternative would be to key the storage by the GUID of the datareader/datawriter on the other side if that information is available in both places.

dirk-thomas: In the `if` case the partition information is passed within the `serializedPayload`. In the `else` case that part is not available. (I don't know if FastRTPS has the information internally and just doesn't pass it along.) I guess we have to build a map mapping the guid to the partitions.

richiware: Since eProsima/FastRTPS@63ea163ddade9576cbf41c829c2494b30192925f ```get_remote_reader_info``` and ```get_remote_writer_info`` return the partition information.

dirk-thomas:
I updated the patch to use the partition information provided by the updated `get_remote_reader_info` / `get_remote_writer_info`. @richiware Thank you for implementing this!

CI looks good:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2665/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/215/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2783/)

Please review.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -391,66 +392,159 @@ typedef struct CustomClientResponse
   : buffer_(nullptr) {}
 } CustomClientResponse;
 
-class topicnamesandtypesReaderListener : public ReaderListener
+class ReaderInfo : public ReaderListener
 {
 public:
-  explicit topicnamesandtypesReaderListener(
+  explicit ReaderInfo(
```
wjwwood: No longer needs to be `explicit`, but probably doesn't matter either.

wjwwood:
@dirk-thomas is this good to merge with the other pr that reenables the tests? If so I'd like to do it tonight, so the nightlies can test it and so I can use it for my next pr.

rmw_fastrtps_cpp/src/functions.cpp
dirk-thomas: I removed it from both the `ReaderInfo` as well as the `WriterInfo`.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -391,66 +392,159 @@ typedef struct CustomClientResponse
...
+    graph_guard_condition_(graph_guard_condition)
   {}
 
-  void onNewCacheChangeAdded(RTPSReader * reader, const CacheChange_t * const change)
+  void onNewCacheChangeAdded(RTPSReader *, const CacheChange_t * const change)
   {
-    (void)reader;
+    ReaderProxyData proxyData;
+    if (change->kind == ALIVE) {
+      CDRMessage_t tempMsg;
```
dirk-thomas: See #124.

PR #113, opened 2017-05-31
implement avoid_ros_namespace_conventions QoS option

wjwwood:
Connects to ros2/rmw#106

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -342,14 +342,15 @@ _assign_partitions_to_attributes(
     attributes->topic.topicName = name_tokens.data[0];
     ret = RCUTILS_RET_OK;
   } else if (name_tokens.size == 2) {
+    std::string partition;
     if (avoid_ros_namespace_conventions) {
       // no prefix to be used, just assign the user's namespace
-      attributes->qos.m_partition.push_back(name_tokens.data[0]);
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -326,19 +326,30 @@ template<typename AttributeT>
...
-    attributes->qos.m_partition.push_back(
-      (std::string(prefix) + "/" + name_tokens.data[0]).c_str());
+    if (avoid_ros_namespace_conventions) {
+      // no prefix to be used, just assign the user's namespace
+      attributes->qos.m_partition.push_back(name_tokens.data[0]);
+    } else {
+      // concat the prefix with the user's namespace
+      attributes->qos.m_partition.push_back(
+        (std::string(prefix) + "/" + name_tokens.data[0]).c_str());
+    }
```
dirk-thomas: I would suggest restructuring the logic to not repeat the `push_back` call, like:

```
std::string partition = name_tokens.data[0];
if (!avoid_ros_namespace_conventions) {
  partition = std::string(prefix) + "/" + partition;
}
attributes->qos.m_partition.push_back(partition.c_str());
```

wjwwood: I updated the logic in bf1f981, but I left the two blocks in the if statement, because I think it's easier to read. But I factored the `push_back` out of the conditional.

PR #114, opened 2017-05-31
Topic names

Karsten1987:
this should fix the inconsistency with `rclcpp::Node::count_subscribers` and `rclcpp::Node::count_publishers` as well as `get_topic_names_and_types`.

The function truncates the internal topic fqdn in the form of `rt/my_namespace/my_topic` into `/my_namespace/my_topic`

@clalancette Could you please review this and report whether this solves your issue?

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <array>
 #include <cassert>
 #include <condition_variable>
 #include <limits>
@@ -551,21 +552,13 @@ inline
 std::string
 _filter_ros_prefix(const std::string & topic_name)
 {
-  if (topic_name.find(std::string(ros_topic_prefix) + "/") == 0) {
-    return topic_name.substr(
-      strlen(ros_topic_prefix),
-      topic_name.size() - strlen(ros_topic_prefix));
-  } else if (topic_name.find(std::string(ros_service_requester_prefix) + "/") == 0) {
-    return topic_name.substr(
-      strlen(ros_service_requester_prefix),
-      topic_name.size() - strlen(ros_service_requester_prefix));
-  } else if (topic_name.find(std::string(ros_service_response_prefix) + "/") == 0) {
-    return topic_name.substr(
-      strlen(ros_service_response_prefix),
-      topic_name.size() - strlen(ros_service_response_prefix));
-  } else {
-    return topic_name;
+  auto prefixes = {ros_topic_prefix, ros_service_requester_prefix, ros_service_response_prefix};
+  for (auto prefix : prefixes) {
+    if (topic_name.rfind(std::string(prefix) + "/") == 0) {
+      return topic_name.substr(strlen(ros_topic_prefix));
+    }
   }
+  return topic_name;
 }
 
 extern "C"
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -2522,7 +2522,15 @@ rmw_get_topic_names_and_types(
   slave_target->mapmutex.lock();
   for (auto it : slave_target->topicNtypes) {
     for (auto & itt : it.second) {
-      unfiltered_topics[it.first].insert(itt);
+      // truncate the ROS specific prefix
+      auto topic_fqdn = it.first;
+      if ( (it.first.find(std::string(ros_topic_prefix) + "/") == 0)
+        || (it.first.find(std::string(ros_service_requester_prefix) + "/") == 0)
+        || (it.first.find(std::string(ros_service_response_prefix) + "/") == 0)) {
+        topic_fqdn = it.first.substr(2, it.first.size() - 2);
+      }
+      //topics[topic_fqdn] = *it.second.begin();}
+      unfiltered_topics[topic_fqdn].insert(itt);
     }
   }
   slave_target->mapmutex.unlock();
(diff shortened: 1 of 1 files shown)


clalancette:
So this isn't currently working for my use case.  Looking at it, the reason is that I am calling count_subscribers with the string "map", but this PR removes all of the prefixes, but not the leading "/", and thus it is comparing "map" to "/map", and failing.  How should we handle that?

Karsten1987:
please rebase to the current head of  rcl and rclcpp as Williams topic name
validation got merged. This will prevent you from having this exact
mismatch.

clalancette:
Ah, OK, will do.

clalancette:
Still the same problem with @wjwwood 's code.

clalancette:
I ended up changing _filter_ros_prefix() to look like this:

```
inline
std::string
_filter_ros_prefix(const std::string & topic_name)
{
  std::list<std::string> prefixes{ros_topic_prefix,
      ros_service_requester_prefix, ros_service_response_prefix};

  for (auto it : prefixes) {
    std::string withslash = it + "/";
    if (topic_name.rfind(withslash, 0) == 0) {
      return topic_name.substr(withslash.length());
    }
  }
  return topic_name;
}
```

This has a few benefits:
* Doesn't repeat the substr code for each string we want to filter
* Only finds the prefix at the beginning of the string, not anywhere in the string
* Removes the slash along with the prefix

The last point there is mostly what makes it work for me.  This code still has the problem that if the user does pass in "/map", it still won't match.  I'm not sure exactly what we want to do about that.  It seems to me that we want to write some kind of "canonicalization" function that takes in a topic, and produces a canonical, fully-qualified version based on the rules on http://design.ros2.org/articles/topic_and_service_names.html, but I'm not sure.

wjwwood:
@Karsten1987 you will still need to patch rclcpp to call `expand_topic_or_service_name` within `count_subscriptions` so that @clalancette doesn't have the issue with `map` vs `/map`.

Karsten1987:
@clalancette I see two potential problems with your proposal:

1.) rfind does only find the last occurrence, that means a valid topic such as `rt/rt/rt` does not get truncated as the last occurrence is not 0.
2.) the list is nice, but it copies the values (nitpick here, but still)

@wjwwood Good point. Will do that straight ahead.

clalancette:
@Karsten1987 Actually, I don't think point 1) is a problem.  The reason is that we are calling rfind() with a pos of 0, which, according to http://www.cplusplus.com/reference/string/string/rfind/, means:

```
Position of the last character in the string to be considered as the beginning of a match
```

Thus, only the very beginning of the string will be considered for matching.  In a test program locally, the `rt/rt/rt` topic does get properly truncated to `rt/rt` when using my method above.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -536,14 +548,21 @@ typedef struct CustomParticipantInfo
...
+_filter_ros_prefix(const std::string & topic_name)
 {
-// static for internal linkage
-static const char * const eprosima_fastrtps_identifier = "rmw_fastrtps_cpp";
-static const char * const ros_topic_prefix = "rt";
-static const char * const ros_service_requester_prefix = "rq";
-static const char * const ros_service_response_prefix = "rr";
+  auto prefixes = {ros_topic_prefix, ros_service_requester_prefix, ros_service_response_prefix};
+  for (auto prefix : prefixes) {
+    if (topic_name.rfind(std::string(prefix) + "/") == 0) {
```
clalancette: As we discussed on chat, this should be:

`if (topic_name.rfind(std::string(prefix) + "/", 0) == 0) {`

clalancette:
Other than the one bug I pointed out in the code, this looks pretty good to me (and, in combination with the changes on the topic_names branch in rclcpp, works for me).  @wjwwood , I'd like to have your opinion on it before we approve.

wjwwood:
@Karsten1987 when do you think you'll be ready to merge this?

I started this CI to see if it helps with the linux packaging job:

[Build Status](http://ci.ros2.org/job/ci_packaging_linux/3/)

clalancette:
@wjwwood That job ran afoul of the setuptools issue, I kicked it off again here:


[Build Status](http://ci.ros2.org/job/ci_packaging_linux/5/)

Karsten1987:
this one seems to compile:
[Build Status](http://ci.ros2.org/job/ci_packaging_linux/6/)

Karsten1987:
The dynamic bridge seems to constantly timeout. Is that a known flaky test?

dhood:
the tests broke when the /rt prefixes were added

Karsten1987:
Then I think this PR is unrelated. Fixing the dynamic bridge tests should be in a separate PR. Thoughts?

clalancette:
@Karsten1987 Yeah, I agree.  If you look at the previous builds (like http://ci.ros2.org/view/packaging/job/packaging_linux/695/#showFailuresLink), that same test fails.  So this PR doesn't make things worse, I think it should be good to merge.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -1799,6 +1816,7 @@ rmw_client_t * rmw_create_client(const rmw_node_t * node,
 
   info->writer_guid_ = info->request_publisher_->getGuid();
 
+  fprintf(stderr, "Original service lcient topic %s\n", service_name);
```
wjwwood: These should not be merged in, here and below.

Karsten1987:
[Build Status](http://ci.ros2.org/job/ci_windows/2780/)
[Build Status](http://ci.ros2.org/job/ci_linux/2662/)
[Build Status](http://ci.ros2.org/job/ci_osx/2115/)

PR #116, opened 2017-06-07
Look for debug versions of fastcdr and fastrtps as well

esteve:
This should address the remaining issue of #75

### DIFF
--- a/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
+++ b/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
@@ -51,11 +51,13 @@ find_library(FastRTPS_LIBRARY_DEBUG
 find_library(FastCDR_LIBRARY_RELEASE
     NAMES fastcdr-${fastcdr_MAJOR_MINOR_VERSION} fastcdr)
 
-find_library(FastCDR_LIBRARY_RELEASE
+find_library(FastCDR_LIBRARY_DEBUG
     NAMES fastcdr${CMAKE_DEBUG_POSTFIX}-${fastcdr_MAJOR_MINOR_VERSION} fastcdr)
 
-set(FastRTPS_LIBRARIES_RELEASE ${FastRTPS_LIBRARY_RELEASE} ${FastCDR_LIBRARY_RELEASE})
-set(FastRTPS_LIBRARIES_DEBUG ${FastRTPS_LIBRARY_DEBUG} ${FastCDR_LIBRARY_DEBUG})
+set(FastRTPS_LIBRARIES
+    optimized ${FastRTPS_LIBRARY_RELEASE} ${FastCDR_LIBRARY_RELEASE}
+    debug ${FastRTPS_LIBRARY_DEBUG} ${FastCDR_LIBRARY_DEBUG}
+)
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(FastRTPS
@@ -66,6 +68,5 @@ find_package_handle_standard_args(FastRTPS
     FastCDR_LIBRARY_DEBUG
     FastRTPS_LIBRARY_RELEASE
     FastRTPS_LIBRARY_DEBUG
-    FastRTPS_LIBRARIES_RELEASE
-    FastRTPS_LIBRARIES_DEBUG
+    FastRTPS_LIBRARIES
 )
--- a/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
+++ b/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
@@ -48,23 +48,21 @@ find_library(FastCDR_LIBRARY_RELEASE
 find_library(FastCDR_LIBRARY_DEBUG
     NAMES fastcdr${CMAKE_DEBUG_POSTFIX}-${fastcdr_MAJOR_MINOR_VERSION})
 
-if(FastCDR_LIBRARY_RELEASE)
-    if(FastCDR_LIBRARY_DEBUG)
-        set(FastCDR_LIBRARIES
-            optimized ${FastCDR_LIBRARY_RELEASE}
-            debug ${FastCDR_LIBRARY_DEBUG}
-        )
-    else()
-        set(FastCDR_LIBRARIES
-            ${FastCDR_LIBRARY_RELEASE}
-        )
-    endif()
+if(FastCDR_LIBRARY_RELEASE AND FastCDR_LIBRARY_DEBUG)
+    set(FastCDR_LIBRARIES
+        optimized ${FastCDR_LIBRARY_RELEASE}
+        debug ${FastCDR_LIBRARY_DEBUG}
+    )
+elseif(FastCDR_LIBRARY_RELEASE)
+    set(FastCDR_LIBRARIES
+        ${FastCDR_LIBRARY_RELEASE}
+    )
+elseif(FastCDR_LIBRARY_DEBUG)
+    set(FastCDR_LIBRARIES
+        ${FastCDR_LIBRARY_DEBUG}
+    )
 else()
-    if(FastCDR_LIBRARY_DEBUG)
-        set(FastCDR_LIBRARIES
-            ${FastCDR_LIBRARY_DEBUG}
-        )
-    endif()
+    set(FastCDR_LIBRARIES "")
 endif()
 
 find_library(FastRTPS_LIBRARY_RELEASE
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Doesn't this always find the debug version of the library if it is available? That would be inappropriate imo.

Also I think there is a CMake variable for the suffix (`CMAKE_DEBUG_POSTFIX`?).

esteve:
@dirk-thomas should I change this so that it picks the release or the debug DLL depending on the build configuration instead? However, I think that Fast-RTPS would only build one or another, but not both debug and release.

dirk-thomas:
Maybe the module should return all available libraries using the keywords to identify `debug` and `optimized` libraries (https://cmake.org/cmake/help/v3.5/command/target_link_libraries.html)?

esteve:
@dirk-thomas I've split the variables into two sets (`_RELEASE` and `_DEBUG`) in https://github.com/ros2/rmw_fastrtps/pull/116/commits/0ddbe4722277f715ae087ec5c531a702b0de0ce6, however I don't know how that should be managed in `ament_export_libraries`. Is the following ok? https://github.com/esteve/rmw_fastrtps/blob/0ddbe4722277f715ae087ec5c531a702b0de0ce6/rmw_fastrtps_cpp/CMakeLists.txt#L63

dirk-thomas:
The module should continue to export just `FastRTPS_LIBRARIES` and the value should contain both when available: `debug somedbg.so optimized someopt.so`.

esteve:
@dirk-thomas I've put back both variables into `FastRTPS_LIBRARIES` and added the `optimized` and `debug` keywords in https://github.com/ros2/rmw_fastrtps/pull/116/commits/7bbb69100ca2be642de4c7221b069e484c8c2160

fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
```diff
@@ -42,20 +42,31 @@ find_package(fastrtps REQUIRED CONFIG)
 string(REGEX MATCH "^[0-9]+\\.[0-9]+" fastcdr_MAJOR_MINOR_VERSION "${fastcdr_VERSION}")
 string(REGEX MATCH "^[0-9]+\\.[0-9]+" fastrtps_MAJOR_MINOR_VERSION "${fastrtps_VERSION}")
 
-find_library(FastRTPS_LIBRARY
+find_library(FastRTPS_LIBRARY_RELEASE
     NAMES fastrtps-${fastrtps_MAJOR_MINOR_VERSION} fastrtps)
 
-find_library(FastCDR_LIBRARY
+find_library(FastRTPS_LIBRARY_DEBUG
+    NAMES fastrtps${CMAKE_DEBUG_POSTFIX}-${fastrtps_MAJOR_MINOR_VERSION} fastrtps)
```
dirk-thomas: Does this line find `fastrtps` (without a debug suffix) and exports it with the `debug` keyword? That seems wrong to me.

esteve: What should be the correct way of exporting the libraries? On Linux and OSX, FastRTPS does not add any suffix if built with debug symbols.

Also, is the following correct? https://github.com/ros2/rmw_fastrtps/blob/7bbb69100ca2be642de4c7221b069e484c8c2160/fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake#L67-L70 I added those variables to `find_package_handle_standard_args`, but I wonder what would happen if `rmw_fastrtps_cpp` is not built for debug.

In any case, given that the current nightly build is broken, if there's any way to solve this, feel free to make changes to this branch, the timezone difference just makes communication rather difficult :-/

dirk-thomas: If both libraries are available they can be exported with the `debug` and `optimized` keywords. If only one is available it can be exported without any keyword.

dirk-thomas: If we can't resolve the problem with the broken nightly we might consider to revert the original change since the beta 2 freeze is approaching and I would rather have all jobs passing for that.

esteve:
@dirk-thomas I think I managed to split the libraries into debug/optimized properly. Could you please make any changes you deem necessary? Otherwise, yeah, I agree #75 should be reverted, though it'd be great to get this merged before beta 2. Thanks!

fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
```diff
@@ -42,31 +42,65 @@ find_package(fastrtps REQUIRED CONFIG)
...
+            ${FastCDR_LIBRARY_RELEASE}
+        )
+    endif()
+else()
+    if(FastCDR_LIBRARY_DEBUG)
+        set(FastCDR_LIBRARIES
+            ${FastCDR_LIBRARY_DEBUG}
+        )
+    endif()
+endif()
```
dirk-thomas: Can you please reduce the nesting by one level and also initialize the variable when neither is found, like:

```
if(RELEASE AND DEBUG)
elseif(RELEASE)
elseif(DEBUG)
else()
endif()
```

dirk-thomas: > Could you please make any changes you deem necessary?

I can make these changes myself as soon as I have time.

esteve: I hope this is what you meant:

https://github.com/ros2/rmw_fastrtps/pull/116/commits/3c34a21082ab8ff9e3c0a3b6d5e3d2a608eab68e

dirk-thomas: :+1:

dirk-thomas:
Latest round of CI builds:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2699/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/240/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2156/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2821/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/2822/)

esteve:
@dirk-thomas thanks for the patience! Here's hoping the builds are fine :crossed_fingers:

esteve:
Unfortunately the debug build failed for Windows, I've pushed https://github.com/ros2/rmw_fastrtps/pull/116/commits/89f0688ac810953315d613e99e0012a53df83c83 in case `CMAKE_DEBUG_POSTFIX` is not set to "d" on Windows, though honestly I don't know for certain if that would fix it. Could any of you paste a screenshot of what the workspace looks like on Jenkins?

dirk-thomas:
`ws/install/fastrtps/lib` contains:

* fastrtps/cmake/
* fastrtpsd-1.4.lib
* fastrtpsd-1.4.pdb
* libfastrtpsd-1.4.lib

dirk-thomas:
@esteve #75 has been reverted in #115 due to the failing builds on one platform. Therefore this needs to be rebased. Let us know when we can trigger some new CI builds for this.

esteve:
I've just rebased this on top of master. This includes https://github.com/ros2/rmw_fastrtps/pull/116/commits/89f0688ac810953315d613e99e0012a53df83c83 which was pushed before the revert and should fix the problem with Windows. Could you run another CI job again? Thanks!

fastrtps_cmake_module/cmake/Modules/FindFastRTPS.cmake
```diff
@@ -36,9 +36,65 @@ set(FastRTPS_FOUND FALSE)
...
   NAMES fastrtps/)
 
+find_package(fastcdr REQUIRED CONFIG)
+find_package(fastrtps REQUIRED CONFIG)
+
+string(REGEX MATCH "^[0-9]+\\.[0-9]+" fastcdr_MAJOR_MINOR_VERSION "${fastcdr_VERSION}")
+string(REGEX MATCH "^[0-9]+\\.[0-9]+" fastrtps_MAJOR_MINOR_VERSION "${fastrtps_VERSION}")
+
+find_library(FastCDR_LIBRARY_RELEASE
+    NAMES fastcdr-${fastcdr_MAJOR_MINOR_VERSION} fastcdr)
```
dirk-thomas: Nitpick: the rest of the file uses two spaces for indentation. It would be good if the added code does too.

esteve: Fixed in https://github.com/ros2/rmw_fastrtps/pull/116/commits/fd352fed1e5bf47e7cd4ddd2f2f4f6cb3e73477e

esteve:
http://ci.ros2.org/job/ci_windows/2847/consoleFull got past `rmw_fastrtps_cpp` woohoo!

dirk-thomas:
Thanks.

PR #117, opened 2017-06-07
add allocator to split function

Karsten1987:
connects to ros2/rcutils#6


dirk-thomas:
The diff seems to contain unrelated changes which are already on master. Please rebase so that the PR only shows the actual proposed changes.

Karsten1987:
I just rebased.

PR #119, opened 2017-06-15
Refactor get topic names and types

wjwwood:
Connects to ros2/ros2#361

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -3133,6 +3133,12 @@ rmw_get_node_names(
     node_names->data[i] = rcutils_strdup(participant_names[i].c_str(), allocator);
     if (!node_names->data[i]) {
       RMW_SET_ERROR_MSG("failed to allocate memory for node name")
+      rcutils_ret = rcutils_string_array_fini(node_names);
+      if (rcutils_ret != RCUTILS_RET_OK) {
+        RCUTILS_LOG_ERROR_NAMED(
+          "rmw_fastrtps_cpp",
+          "failed to cleanup during error handling: %s", rcutils_get_error_string_safe())
+      }
       return RMW_RET_BAD_ALLOC;
     }
   }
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -2882,12 +3122,19 @@ rmw_get_node_names(
...
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
   for (size_t i = 0; i < participant_names.size(); ++i) {
-    size_t name_length = participant_names[i].size() + 1;
-    node_names->data[i] = static_cast<char *>(rmw_allocate(name_length * sizeof(char)));
-    snprintf(node_names->data[i], name_length, "%s", participant_names[i].c_str());
+    node_names->data[i] = rcutils_strdup(participant_names[i].c_str(), allocator);
+    if (!node_names->data[i]) {
+      RMW_SET_ERROR_MSG("failed to allocate memory for node name")
+      return RMW_RET_BAD_ALLOC;
```
dirk-thomas: How are previously allocated `data[i]` freed in this case? Does it need something like `fail_cleanup` above?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -461,13 +466,8 @@ class ReaderInfo : public ReaderListener
...
     if (change->kind == ALIVE) {
-      if (
-        it == topicNtypes.end() ||
-        it->second.find(proxyData.typeName()) == it->second.end())
-      {
-        topicNtypes[fqdn].insert(proxyData.typeName());
-        trigger = true;
-      }
+      topicNtypes[fqdn].insert(proxyData.typeName());
+      trigger = true;
```
dirk-thomas: With `it` being removed from the `if` block the declaration of `auto it = ...` can be moved into the `else` block.

Same below.

wjwwood: Sure: ba254ee

rmw_fastrtps_cpp/src/functions.cpp
wjwwood: See: bfa8aab I didn't use a label because this was at the end of the function anyways.

PR #120, opened 2017-06-19
Fix Compiler Warning

allenh1:
In newer versions of GCC, a compiler warning appears when you catch a polymorphic type by value.

```
Scanning dependencies of target rmw_fastrtps_cpp
[ 50%] Building CXX object CMakeFiles/rmw_fastrtps_cpp.dir/src/functions.cpp.o
/home/allenh1/ros2_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/functions.cpp: In function ‘rmw_node_t* create_node(const char*, const char*, eprosima::fastrtps::ParticipantAttributes)’:
/home/allenh1/ros2_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/functions.cpp:841:17: warning: catching polymorphic type ‘class std::bad_alloc’ by value [-Wcatch-value=]
   } catch (std::bad_alloc) {
                 ^~~~~~~~~
```

The fix is to catch a reference (or a pointer). I personally don't understand why it is _bad_ to catch a polymorphic type by value (I guess it could be bad if your constructor is explicit?), though, if anyone knows why this is a _bad thing_, I would appreciate some insight.

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -838,7 +838,7 @@ rmw_node_t * create_node(
 
   try {
     node_impl = new CustomParticipantInfo();
-  } catch (std::bad_alloc) {
+  } catch (std::bad_alloc &) {
     RMW_SET_ERROR_MSG("failed to allocate node impl struct");
     goto fail;
   }


allenh1:
Ok, found the [reason.](https://www.securecoding.cert.org/confluence/display/cplusplus/ERR61-CPP.+Catch+exceptions+by+lvalue+reference)

> Because the variable declared by the exception-declaration is copy-initialized, it is possible to slice the exception object as part of the copy operation, losing valuable exception information 

> Further, if the copy constructor of the exception object throws an exception, the copy initialization of the exception-declaration object results in undefined behavior. (See ERR60-CPP. Exception objects must be nothrow copy constructible for more information.)

nuclearsandwich:
Are these changes necessary/suitable for the feature freeze window?

allenh1:
@nuclearsandwich I was wondering that myself.

I would argue that this is a suitable change for a code freeze window, but I could still use some clarification.

mikaelarguedas:
This is not a bugfix and doesn't target our supported platforms so I'd say it is not necessary / suitable.

allenh1:
@mikaelarguedas that's certainly reasonable. As a general rule, anything tagged as bug should be something to work on during a feature freeze, right?

dirk-thomas:
We are only in feature freeze - not code freeze. And this is clearly fixing "bad" code so this should be merged now. No need to let it sit since there is no regression potential.

nuclearsandwich:
> We are only in feature freeze - not code freeze.

The purpose of the feature freeze is to slow changes to source repositories long enough for testing and verification of both source builds and binary packages. If we continue to merge code just because it doesn't count as a "feature" we'll be pushing back the time when we can verify a complete set of packages.

> This is not a bugfix targeting our supported platforms so I'd say it is not necessary / suitable

Given this statement, I would like to wait on merging this for now so I can keep a high signal to noise ratio when checking for out-of-date packages. If bugfixes that do target beta2 need to be merged into this repository then I we can let this in on those coattails.

dirk-thomas:
> The purpose of the feature freeze is to slow changes to source repositories long enough for testing and verification of both source builds and binary packages. If we continue to merge code just because it doesn't count as a "feature" we'll be pushing back the time when we can verify a complete set of packages.

I don't see how this change would affect testing and stabilizing of the code.

I don't want to spend more time on this. If there are additional +1's to merge this now (since it fixes a valid warning as well as has no regression potential) I will move forward and merge it. If not it will just sit here.

nuclearsandwich:
> I don't see how this change would affect testing and stabilizing of the code.

It means that either I cut another release of rmw_fastrtps and trigger rebuilds of everything downtsream of it on the ROS 2 farm or ignore the stale release warnings I set up so I can keep packages fresh. One false positive isn't a big deal but if we continue making changes that don't affect the release it's going to increase the likelihood that an actually stale package gets lost in the noise.

mikaelarguedas:
Agreed it's a net improvement and low risk, but I think @nuclearsandwich should be the one to make the call because he is the one blooming all these ros2 packages. I'm ok with merging now or merging later

tfoote:
We don't have a unit test failing right now, and our default linters and compilers don't catch it. But  the newer compiler has more protection. At some point in the future this will be a warning when we start testing with the newer compiler. And our policy is to have no warnings. If it was a warning in our CI it would not be a discussion. This is just code cleanup.

If we're holding things right now for a currently ongoing rebuild is fine. However I think it clearly fits inside our feature freeze madate where we're supposed to be doing testing and debugging to verify and improve the release experience. We shouldn't hold these sorts of bug fixes for the whole duration of our testing cycle before beta2.

nuclearsandwich:
> If we're holding things right now for a currently ongoing rebuild is fine. However I think it clearly fits inside our feature freeze mandate where we're supposed to be doing testing and debugging to verify and improve the release experience. We shouldn't hold these sorts of bug fixes for the whole duration of our testing cycle before beta2.

This works for me. This morning I was working quickly trying to make sure this didn't set back being able to deliver an initial crop of debs for testing.  We still have to refresh ament_lint after resolving python issues there so there will definitely be another "ground up" rebuild then. If @allenh1 and the reviewers are amenable, I'll merge this the next time rmw_fastrtps would get rebuilt by upstream anyway.

nuclearsandwich:
@allenh1 since other changes have been merged here you can go ahead and merge this at your leisure.

PR #123, opened 2017-06-22
use hpp for c++ code

wjwwood:
For consistency and to not confuse `cppcheck`.

Connects to ros2/ros2#368

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RMW_FASTRTPS_CPP__TYPESUPPORT_IMPL_H_
-#define RMW_FASTRTPS_CPP__TYPESUPPORT_IMPL_H_
+#ifndef RMW_FASTRTPS_CPP__TYPESUPPORT_IMPL_HPP_
+#define RMW_FASTRTPS_CPP__TYPESUPPORT_IMPL_HPP_
 
 #include <fastcdr/FastBuffer.h>
 #include <fastcdr/Cdr.h>
@@ -70,7 +70,7 @@ rosidl_generator_c__void__Array__init(
   }
   void * data = NULL;
   if (size) {
-    data = (void *)calloc(size, member_size);
+    data = static_cast<void *>(calloc(size, member_size));
     if (!data) {
       return false;
     }
@@ -319,7 +319,7 @@ void serialize_array<std::string>(
     // tmpstring is defined here and not below to avoid
     // memory allocation in every iteration of the for loop
     std::string tmpstring;
-    auto string_field = (rosidl_generator_c__String *) field;
+    auto string_field = static_cast<rosidl_generator_c__String *>(field);
     for (size_t i = 0; i < member->array_size_; ++i) {
       tmpstring = string_field[i].data;
       ser.serialize(tmpstring);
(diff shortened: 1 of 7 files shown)


wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2788/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/314/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2260/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2926/)

The other changes (to force cppcheck to interpret `.h` as C++) will need to be done separately once the newly updated Windows machines are turned on (it won't fail until that happens so it will be hard to test). Anyways these changes can be done separately.

mikaelarguedas:
Do you think it's worth it to split the renaming of the files and the change of the include rules into different commits so that the affected files preserve their history? (I know that means that the state of the code in between the 2 commits won't compile but I think it's worth it)

allenh1:
@mikaelguerdas the diff makes it look like it picked up on the renaming. I think it's safe.

mikaelarguedas:
Id did, unfortunately the history is still this single commit : https://github.com/ros2/rmw_fastrtps/commits/2403a757da9e55e8cc31979909cbb6abba4452a0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/MessageTypeSupport.hpp So my concern remains

allenh1:
Ah, I see. Ya, that would be rather unfortunate.

wjwwood:
@mikaelarguedas are you sure that having a separate commit would change anything? I'm pretty sure that GitHub would have to implement the `--follow` option of `git log`:

https://stackoverflow.com/questions/5646174/how-to-make-github-follow-directory-history-after-renames

I think you can follow the file rename through the git blame interface on GitHub:

https://github.com/ros2/rmw_fastrtps/blame/2403a757da9e55e8cc31979909cbb6abba4452a0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/MessageTypeSupport.hpp

mikaelarguedas:
Oh I thought id did but I stand corrected by Github UI o_O (https://github.com/ros2/joystick_drivers/commits/ros2/joy/src/joy_node_linux.cpp). Totally fine as is then, sorry for the noise.

wjwwood:
Forgot to run `ament_cpplint`, doh. Fixing up now, I'll have to run CI again.

wjwwood:
New CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2789/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/315/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2261/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2927/)

@mikaelarguedas (or anyone else) can you re-review this now?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -344,8 +344,8 @@ size_t get_array_size_and_assign_field(
   size_t space)
 {
   std::vector<unsigned char> * vector = reinterpret_cast<std::vector<unsigned char> *>(field);
-  void * ptr = (void *)sub_members_size;
-  size_t vsize = vector->size() / (size_t)align_(max_align, 0, ptr, space);
+  void * ptr = reinterpret_cast<void *>(sub_members_size);
```
wjwwood: btw, we're casting a `size_t` to a `void *` here :nauseated_face:.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -376,7 +377,7 @@ bool TypeSupport<MembersType>::serializeROSmessage(
 
   for (uint32_t i = 0; i < members->member_count_; ++i) {
     const auto member = members->members_ + i;
-    void * field = (char *)ros_message + member->offset_;
+    void * field = const_cast<char *>(static_cast<const char *>(ros_message)) + member->offset_;
```
wjwwood: Also, I could not convert the `const void *` directly to `char *` so I did it in two steps...

wjwwood:
Can someone please re-review this one?

wjwwood:
I'm gonna merge this one, since it is green without the related cppcheck changes in the other pr's. Note that linux-aarch is unstable because I had to restart it but only after I pushed some, but not all, of the other branches. At any rate I'm sure this pr is good to go.

PR #124, opened 2017-06-23
fix memory leak

dirk-thomas:
When calling the constructor without any arguments the pointer of the allocated buffer is later overwritten and therefore leaked: https://github.com/eProsima/Fast-RTPS/blob/f16b8f365c44aa76066b35004e84e9a352949d8c/include/fastrtps/rtps/common/CDRMessage_t.h#L53

Calling the constructor with the size argument `0` ensures that no memory is being allocated within the constructor.

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -437,7 +437,7 @@ public:
   {
     ReaderProxyData proxyData;
     if (change->kind == ALIVE) {
-      CDRMessage_t tempMsg;
+      CDRMessage_t tempMsg(0);
       tempMsg.wraps = true;
       tempMsg.msg_endian = change->serializedPayload.encapsulation ==
         PL_CDR_BE ? BIGEND : LITTLEEND;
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -507,7 +507,7 @@ class WriterInfo : public ReaderListener
   {
     WriterProxyData proxyData;
     if (change->kind == ALIVE) {
-      CDRMessage_t tempMsg;
+      CDRMessage_t tempMsg(0);
```
clalancette: I'm OK with this fix for now (since CDRMessage_t is not in our code), but it feels fragile to me.  For instance, it looks like the WriterInfo class has the same leak.  It seems to me that we should probably add (or get eProsima to add) another constructor to CDRMessage_t so that we can specify all of the pieces we want it to be without accessing the underlying fields directly.  Otherwise, I can see this occurring again.

dirk-thomas: If your are ok with the fix as is can you approve it?

dirk-thomas: The patch has been updated to also cover the writer part.

clalancette: I'll approve it, yeah.  Can you please open an issue against Fast-RTPS so we can address this better in the future?  Thanks.

dirk-thomas: I don't think the API in FastRTPS needs any changes. More documentation would certainly help. If you think a ticket or patch for FastRTPS makes sense please go ahead and create it.

clalancette: I ended up opening https://github.com/eProsima/Fast-RTPS/issues/115, since it seems this could be made a lot nicer to use.

Issue #127, opened 2017-06-27
Bad rmw_count_subscribers value


fmrico:
I am using the master branch with Fast-RTPS.

I've been testing the ros1_bridge with one ros1 talker and two ros2 listener. 

![pre_run](https://user-images.githubusercontent.com/3810011/27606297-053f466a-5b80-11e7-9da6-eeac8ad45ac4.png)

When running, I checked on introspection that it only showed that there was 1 subscriber in /chatter, instead of the two that were active. 

![run](https://user-images.githubusercontent.com/3810011/27606332-1fd00c76-5b80-11e7-90ac-fece28732c00.png)

When I interrupted one of the ros2 listener, the other stopped receiving, and the instrocpection shows that there is no subscribers for that topic.

![final](https://user-images.githubusercontent.com/3810011/27606363-3bdfcb0e-5b80-11e7-805c-b7d8a5332f5a.png)

I have debugged a bit and I have reached rmw_fastrtps/rmw_functions.cpp, where it seems that the rmw_count_subscribers function does not return the correct value.

allenh1:
Thanks for the report! After playing around with this on my machine, I was not able to reproduce this exact behavior, but I did find something a little off:

```
allenh1@machina:~/ros2_ws$ ros2 run ros1_bridge dynamic_bridge
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
  Passing message from ROS 1 to ROS 2
[...]

PR #129, opened 2017-06-28
Fix publisher/subscriber Count

allenh1:
Fix the counting for the subscribers and publishers to a topic.

connects to #127

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -475,6 +475,11 @@ public:
         if (loc != std::end(it->second)) {
           topicNtypes[fqdn].erase(loc, loc + 1);
           trigger = true;
+        } else {
+          RCUTILS_LOG_DEBUG_NAMED(
+            "rmw_fastrps_cpp",
+            "unexpected removal of subscription on topic '%s' with type '%s'",
+            fqdn.c_str(), proxyData.typeName().c_str());
         }
       }
     }
@@ -546,7 +551,11 @@ public:
         if (loc != std::end(it->second)) {
           topicNtypes[fqdn].erase(loc, loc + 1);
           trigger = true;
-        }
+        } else {
+          RCUTILS_LOG_DEBUG_NAMED(
+            "rmw_fastrps_cpp",
+            "unexpected removal of subscription on topic '%s' with type '%s'",
+            fqdn.c_str(), proxyData.typeName().c_str());
       }
     }
     mapmutex.unlock();
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -465,16 +465,20 @@ public:
     bool trigger = false;
     mapmutex.lock();
     if (change->kind == ALIVE) {
-      topicNtypes[fqdn].insert(proxyData.typeName());
+      topicNtypes[fqdn].push_back(proxyData.typeName());
       trigger = true;
     } else {
       auto it = topicNtypes.find(fqdn);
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -465,16 +465,20 @@ class ReaderInfo : public ReaderListener
...
-      if (
-        it != topicNtypes.end() &&
-        it->second.find(proxyData.typeName()) != it->second.end())
-      {
-        topicNtypes[fqdn].erase(proxyData.typeName());
-        trigger = true;
+      auto vector_find = [] (const auto & v, const auto & data) {
+          auto ret = std::find(std::begin(v), std::end(v), data);
+          return ret;
+        };
```
dirk-thomas: Since you only use the lambda once I don't think that the overhead in complexity and reduced readability is justified. Please keep it as a simple if condition.

Same below.

allenh1:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2836/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/340/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2292/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2972/)

dirk-thomas:
* Linux packaging [Build Status](http://ci.ros2.org/job/ci_packaging_linux/32/)

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -535,16 +536,17 @@ class WriterInfo : public ReaderListener
...
-      {
-        topicNtypes[fqdn].erase(proxyData.typeName());
-        trigger = true;
+      if (it != topicNtypes.end()) {
+        const auto & loc =
+          std::find(std::begin(it->second), std::end(it->second), proxyData.typeName());
+        if (loc != std::end(it->second)) {
+          topicNtypes[fqdn].erase(loc, loc + 1);
+          trigger = true;
+        }
```
wjwwood: Is the else case here an error? Seems to me that if we get an event for deregistering a type that we're not currently tracking it might be a bug?

The one case I could imagine that would be a false positive is that a subscriber comes up and goes down, but this program only receives the latter notification for some reason. Maybe it deserves a warning? I'm not sure. It seems like a place where a real bug could go unnoticed because we're ignoring this atypical case.

wjwwood:
I'm rebuilding with this change to see if `ros2 topic list -t` shows duplicates.

wjwwood:
If so I'll try to make a test in `rcl` for this.

allenh1:
* Linux packaging [Build Status](http://ci.ros2.org/job/ci_packaging_linux/33/)

wjwwood:
Never mind, the `rmw_get_*_names_and_types()` functions already push the types into a `std::set` in a separate step related to demangling. So this should be good. The cli output looks good too.

allenh1:
Alright, cool. @dirk-thomas would you mind giving this a review for me?

dirk-thomas:
I triggered a new set of CI builds since the previous ones didn't cover the last two commits:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2838/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/342/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2294/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2974/)

dirk-thomas:
Merging with fate in aarch64 :wink:

Issue #131, opened 2017-07-03
refactor functions.cpp


Karsten1987:
This is a top level issue for the refactoring task. The idea is to refactor the `_functions.cpp` into multiple files, making the package easier to maintain.

dirk-thomas:
Please hold off to start any work on this until we can sync and ensure that only one person touches the repo during that refactoring.

Karsten1987:
yes agreed. I'll plan on getting rmw_fastrtps_cpp as well as rmw_connext_cpp refactored within one week. What is the best way to sync on this one?

dirk-thomas:
We talked about this briefly in yesterdays meeting. Nobody else will touch the two repos `rmw_connext` and `rmw_fastrtps` for the time being. So you have exclusive access and can refactor both in a similar way as `rmw_opensplice_cpp`. Please make sure to structure the patch like this:

* First commit copying the `functions.cpp` to different file names (`.cpp` as well as `.h` / `.hpp`) without any changes. This will make sure history is browsable.
* Second commit removing blocks of code from all new files not needed in them (as well as the obsolete `functions.cpp`). This will make sure that its easy to review.
* Third commit update CMake, includes etc. to make the code compile again.
* Any additional changes you would like to apply. Keeping those separate from the "code shuffling" is important to make the review easier.

dirk-thomas:
Addressed by #135.

PR #132, opened 2017-07-07
add missing mutex lock

dirk-thomas:
Fixes ros2/ros2#376.

I know this might conflict with the refactoring @Karsten1987 is about to do (#131). If you haven't started the refactoring yet this could maybe be merged before. Otherwise since this patch is minimal I can easily rebase it afterwards.

### DIFF
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -648,6 +648,7 @@ public:
 
   bool hasData()
   {
+    std::lock_guard<std::mutex> lock(internalMutex_);
     return !list.empty();
   }
 
@@ -1879,6 +1880,7 @@ public:
 
   bool hasData()
   {
+    std::lock_guard<std::mutex> lock(internalMutex_);
     return !list.empty();
   }
 
--- a/rmw_fastrtps_cpp/src/functions.cpp
+++ b/rmw_fastrtps_cpp/src/functions.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <array>
+#include <atomic>
 #include <cassert>
 #include <condition_variable>
 #include <limits>
@@ -579,7 +580,7 @@ class ClientListener : public SubscriberListener
 {
 public:
   explicit ClientListener(CustomClientInfo * info)
-  : info_(info),
+  : info_(info), list_has_data_(false),
     conditionMutex_(NULL), conditionVariable_(NULL) {}
 
 
@@ -606,6 +607,7 @@ public:
           } else {
             list.push_back(response);
           }
+          list_has_data_.store(true);
         }
       }
     }
@@ -621,11 +623,13 @@ public:
       if (!list.empty()) {
         response = list.front();
         list.pop_front();
+        list_has_data_.store(!list.empty());
       }
     } else {
       if (!list.empty()) {
         response = list.front();
         list.pop_front();
+        list_has_data_.store(!list.empty());
       }
     }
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2855/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/359/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2311/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2994/)

Karsten1987:
I believe that `has_data` is called periodically. With this lock guard, the call would be blocking and potentially breaks any realtime safe execution. Can we replace this with `std::try_lock`?

http://en.cppreference.com/w/cpp/thread/try_lock

dirk-thomas:
I am not sure I understand: what should the code do if the `try_lock` fails?

wjwwood:
> I am not sure I understand: what should the code do if the try_lock fails?

It could busy wait (either by looping there, or triggering the condition variable that the `rmw_wait` function is using so the predicate gets evaluated again), but that doesn't sound very nice.

Karsten1987:
I may not have the full picture here, but my thinking was that if this mutex can't be acquired because the data has yet not been completely read from the DDS, then I would say `has_data()` should return false, indicating that there is no data to be taken.

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -648,6 +648,7 @@ class ClientListener : public SubscriberListener
 
   bool hasData()
   {
+    std::lock_guard<std::mutex> lock(internalMutex_);
     return !list.empty();
```
wjwwood: You could store this state in an atomic boolean variable just before any calls to notify `conditionVariable_` after `list` is modified. That way you would not need a lock here.

Karsten1987: +1 for the atomic variable

dirk-thomas:
> I may not have the full picture here, but my thinking was that if this mutex can't be acquired because the data has yet not been completely read from the DDS, then I would say has_data() should return false, indicating that there is no data to be taken.

The mutex being currently locked does by no means imply that no data is available. It could be that a different part of the code is holding a lock while data is available.

rmw_fastrtps_cpp/src/functions.cpp
dirk-thomas: Updated in 4be8c428f8f61a1f58335ac7c1d0a26e9008ffc6.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2859/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/363/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2314/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2999/)

dirk-thomas:
Ready for re-review.

dirk-thomas:
@Karsten1987 I just noticed that we wanted to wait in case this conflicts with your refactoring. I am sorry I missed that. If it generates more trouble for you please feel free to revert this and I will redo the patch after your PR.

Karsten1987:
my idea was to finish my PR and maybe cherry-pick this one commit. Would
that work or is that no good practice?

rmw_fastrtps_cpp/src/functions.cpp
```diff
@@ -606,6 +607,7 @@ class ClientListener : public SubscriberListener
           } else {
             list.push_back(response);
           }
+          list_has_data_.store(true);
```
dirk-thomas: With the `list_has_data_` flag now being changed outside of the `conditionMutex_` lock this introduces a race condition with the `rmw_wait()` function. After the function locked the mutex it checks `hasData()` of the listeners and then makes a decision if it has to call `wait`. With this change the boolean flag might change at any time which make `rmw_wait()` call `wait` whe it shouldn't.

See ros2/rmw_fastrtps#147 for the proposed fix.

Before this patch the flag was represented by the state of the container and therefore only changed within the scope of the locked condition.

PR #135, opened 2017-07-11
Refactor functions cpp

Karsten1987:
connects to ros2/rmw_fastrtps#131

summary of commits:

1.) split `functions.cpp` into multiple files (git blame on each file still shows correct history)
2.) cleanup these files with only their respective necessary parts
3.) change `CMakeLists.txt` to compile all files
4.) change `TypeSupport_impl.hpp` to correctly compile template specifications as inline
5.) a little bit of style for functions

I'll leave it as it is for now, hoping on getting this PR merged quickly. Further improvements can thus be done in separate PRs.


rmw_fastrtps_cpp/src/namespace_prefix.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef NAMESPACE_PREFIX_HPP_
+#define NAMESPACE_PREFIX_HPP_
+
+#include <vector>
+#include <string>
+
+extern "C"
+{
+extern const char * const ros_topic_prefix;
```
allenh1: If you're in an `extern` group, do you need to `extern` the internal definitions? Not sure if this was there previously, but I think the internal `extern`'s should be removed.

Karsten1987:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2868/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2321/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/3006/)

rmw_fastrtps_cpp/src/namespace_prefix.hpp
Karsten1987: aren't these two different things? The `extern "C"` is responsible for symbol mangling and the `extern const..` is for telling that the definition of this variable is happening in another file?

dirk-thomas:
The patch looks good to me. Since the PR should be merged without squashing at the end can you please squash some of the later commits wherever it makes sense (the first two / three commits must stay separated though).

dirk-thomas:
Currently the first commit adds a `.bck` file and the last commit removes it again. I will rewrite the history on this branch to get rid of this temporary file.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2892/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/379/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2331/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3016/)

Issue #122, opened 2017-06-20
audit rmw_fastrtps_cpp


mikaelarguedas:
meta-ticket, follow-up of #107 : 
includes
- [ ] audit calls to malloc and new and replace them if appropriate
- [ ] audit any function call that can throw and either catch them or replace them
- [ ] remove magic numbers: partially captured in  #72 
- [ ] remove irrelevant todos
- [ ] fix cleanup on all failure scenarios: catprued in  #33 
- [x] split functions.cpp in digestable files: done in #135
- [ ] replace assert with return with proper value and error messages
- [ ] audit sanity checks to make sure they make sense (e.g. https://github.com/ros2/rmw_fastrtps/pull/138#discussion_r132242929)

(feel free to complete this list I likely missed a few important items)

mikaelarguedas:
refactor performed, audit not done yet

clalancette:
Given the sheer age of this, I'm going to close it out.  All of the code can probably use an audit, but `rmw_fastrtps` is so different from when this is opened that I don't think the same things apply anymore.  If you disagree, feel free to reopen.

PR #138, opened 2017-08-04
use rcutils logging

mikaelarguedas:


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.hpp
```diff
@@ -56,11 +58,11 @@ struct StringHelper<rosidl_typesupport_introspection_c__MessageMembers>
   {
     auto c_string = static_cast<rosidl_generator_c__String *>(data);
     if (!c_string) {
-      fprintf(stderr, "Failed to cast data as rosidl_generator_c__String\n");
+      RCUTILS_LOG_ERROR("Failed to cast data as rosidl_generator_c__String")
       return "";
     }
     if (!c_string->data) {
-      fprintf(stderr, "rosidl_generator_c_String had invalid data\n");
+      RCUTILS_LOG_ERROR("rosidl_generator_c_String had invalid data")
```
wjwwood: Why no named logger here?

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -143,12 +145,12 @@ rmw_create_client(const rmw_node_t * node,
...
-  fprintf(stderr, "Sub Partition %s\n", subscriberParam.qos.m_partition.getNames()[0].c_str());
-  fprintf(stderr, "Pub Topic %s\n", publisherParam.topic.topicName.c_str());
-  fprintf(stderr, "Pub Partition %s\n", publisherParam.qos.m_partition.getNames()[0].c_str());
-  fprintf(stderr, "***********\n");
+  RCUTILS_LOG_DEBUG("************ Client Details *********")
+  RCUTILS_LOG_DEBUG("Sub Topic %s", subscriberParam.topic.topicName.c_str())
+  RCUTILS_LOG_DEBUG("Sub Partition %s", subscriberParam.qos.m_partition.getNames()[0].c_str())
+  RCUTILS_LOG_DEBUG("Pub Topic %s", publisherParam.topic.topicName.c_str())
+  RCUTILS_LOG_DEBUG("Pub Partition %s", publisherParam.qos.m_partition.getNames()[0].c_str())
+  RCUTILS_LOG_DEBUG("***********")
```
wjwwood: Named? Also should the `#ifdef ...` be removed now that it can be disable/enabled with the logger level?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.hpp
mikaelarguedas: oh actually I put named only in places where it was already emulating a named logger. But yeah we can use named everywhere :+1:

rmw_fastrtps_cpp/src/rmw_client.cpp
mikaelarguedas: oh I thought that the severity levels were not observed. Thanks I'll remove them then!

mikaelarguedas:
All comments have been addressed.

I still have one concern about the debug prints, won't that make all tests relying on console output matching fail if built in debug mode ?

dhood:
I think for testing we'd set the severity to whatever's appropriate via https://github.com/ros2/rcutils/blob/1e4fd214ce20572d225e1ac2a7457dabdba26c5b/src/logging.c#L53 (from my understanding it's not tied to building in debug or not)

but, the default level looks to be [set to debug](https://github.com/ros2/rcutils/blob/1e4fd214ce20572d225e1ac2a7457dabdba26c5b/src/logging.c#L33) currently: @dirk-thomas is it appropriate to change that to info? that would mean this PR doesn't require any changes to current testing based on console output matching

dirk-thomas:
> is it appropriate to change that to info? that would mean this PR doesn't require any changes to current testing based on console output matching

While that would make the tests pass it would also prevent any debug output. Maybe the tests should set the verbosity level explicitly if they are that "picky" about the output?

dhood:
I thought that a default verbosity level of 'info' would be more appropriate in general, regardless of the tests. Isn't the standard behaviour to have debug prints be output only when they are explicitly enabled for users that are in the process of debugging something? (perhaps we intend to specify the "ROS default" somewhere higher in the stack one day so it doesn't matter what default value is coded in rcutils?)

mikaelarguedas:
launching CI now that https://github.com/ros2/rcutils/pull/46 has been merged:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2971/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/411/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3081/)

mikaelarguedas:
ci came back happy, placing this in review and launching one debug job just to confirm we wont have any bad surprise
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2972/)

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -215,8 +225,9 @@ rmw_create_client(const rmw_node_t * node,
         _unregister_type(participant, info->response_type_support_, info->typesupport_identifier_);
       }
     } else {
-      fprintf(stderr,
-        "[rmw_fastrtps] leaking type support objects because node impl is null\n");
+      RCUTILS_LOG_ERROR_NAMED(
+        "rmw_fastrps_cpp",
```
dirk-thomas: Spelling.

Same multiple times below.

mikaelarguedas: arg I had to chose an already typoed line for my copy-n-paste :confounded: thanks for catching it

mikaelarguedas: fixed in https://github.com/ros2/rmw_fastrtps/pull/138/commits/cda8cb8d82e9b02173cfc7e2fa80f71ce20a18a0

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.hpp
```diff
@@ -56,11 +58,15 @@ struct StringHelper<rosidl_typesupport_introspection_c__MessageMembers>
   {
     auto c_string = static_cast<rosidl_generator_c__String *>(data);
     if (!c_string) {
-      fprintf(stderr, "Failed to cast data as rosidl_generator_c__String\n");
+      RCUTILS_LOG_ERROR_NAMED(
+        "rmw_fastrtps_cpp",
+        "Failed to cast data as rosidl_generator_c__String")
```
dirk-thomas: A question unrelated to this PR: The `static_cast` will always succeed. So the error message and the test are a bit bogus. Should it be replaced with a check for `!data` and a corresponding error message instead?

mikaelarguedas: that's a good point. Can you create a follow-up PR for this given that this is out of the scope of this one ?

mikaelarguedas: referenced this in the [audit rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/issues/122) issue

mikaelarguedas:
@ros2/team ready for review

Issue #72, opened 2016-12-07
Magic_numbers in calculateMaxSerializedSize


mikaelarguedas:
This is a follow-up of https://github.com/ros2/rmw_fastrtps/issues/36
There are still magic numbers ([101](https://github.com/ros2/rmw_fastrtps/blob/8623ccde0fbdeab8142ed4062978c977446327f0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L772), [257](https://github.com/ros2/rmw_fastrtps/blob/8623ccde0fbdeab8142ed4062978c977446327f0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L805) and [100](https://github.com/ros2/rmw_fastrtps/blob/8623ccde0fbdeab8142ed4062978c977446327f0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L426)) use to determine the maximum size of a sample.
It would be good to investigate if we can get rid of them and improve the logic of the [MaxSize calculation](https://github.com/ros2/rmw_fastrtps/blob/8623ccde0fbdeab8142ed4062978c977446327f0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L135)

allenh1:
@mikaelarguedas This might also be something I could work on

mikaelarguedas:
@allenh1 Is this one still on your radar ?

allenh1:
Yep -- I was waiting on the refactor to finish, but looks like I can go after it now.

mikaelarguedas:
There is now only [one magic number left](https://github.com/ros2/rmw_fastrtps/blob/79d8dc4ed25306ce44d97fede4d8fa0f474f7a34/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L427).

@richiprosima I see there is a [todo in your name](https://github.com/ros2/rmw_fastrtps/blob/79d8dc4ed25306ce44d97fede4d8fa0f474f7a34/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L443) to modify it, do you have any insight as of if this variable should be kept ?

richiprosima:
[...]

PR #140, opened 2017-08-10
Clean up Magic Numbers

allenh1:
This is still in progress, but opening this for visibility.

connects to #72

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -755,7 +755,7 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
           current_alignment += encapsulation +
             eprosima::fastcdr::Cdr::alignment(current_alignment, encapsulation) +
-            (member->string_upper_bound_ ? member->string_upper_bound_ + 1 : 257);
+            member->string_upper_bound_ + 1;
           break;
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
           {
@@ -769,12 +769,10 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
       }
     } else {
       size_t array_size = member->array_size_;
-
       // Whether it is a sequence.
       if (array_size == 0 || member->is_upper_bound_) {
-        current_alignment += 4 + eprosima::fastcdr::Cdr::alignment(current_alignment, 4);
-
-        if (array_size == 0) {array_size = 101;}
+        current_alignment += encapsulation +
+          eprosima::fastcdr::Cdr::alignment(current_alignment, encapsulation);
       }
 
       switch (member->type_id_) {
@@ -807,7 +805,7 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
             for (size_t index = 0; index < array_size; ++index) {
               current_alignment += encapsulation +
                 eprosima::fastcdr::Cdr::alignment(current_alignment, encapsulation) +
-                (member->string_upper_bound_ ? member->string_upper_bound_ + 1 : 257);
+                member->string_upper_bound_ + 1;
             }
           }
           break;
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -720,7 +720,8 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
   size_t initial_alignment = current_alignment;
 
   // Encapsulation
-  current_alignment += 4;
+  const size_t encapsulation = 4;
+  current_alignment += encapsulation;
 
   for (uint32_t i = 0; i < members->member_count_; ++i) {
     const auto * member = members->members_ + i;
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -732,30 +733,34 @@ size_t TypeSupport<MembersType>::calculateMaxSerializedSize(
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT8:
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_CHAR:
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT8:
-          current_alignment += 1;
+          current_alignment += sizeof(int8_t);
```
dirk-thomas: Are those actually dependent on the system or is the alignment coming from the DDS spec? In the latter case the literal number makes more sense imo.

Same below.

allenh1: That's a good question, and one I do not know the answer to. Perhaps it would be ok to say this?

```c++
current_alignment += 1;  /* sizeof(int8_t) */
```

allenh1:
@mikaelarguedas I just tested this branch (locally) with your new tests, and all seems to be ok.

> Are those actually dependent on the system or is the alignment coming from the DDS spec? 

Do you know the answer to this question? I don't know enough about DDS to make a claim there.

mikaelarguedas:
> mikaelarguedas I just tested this branch (locally) with your new tests, and all seems to be ok.

cool!

> Do you know the answer to this question? I don't know enough about DDS to make a claim there.

No I don't, if it's in the spec I would expect it to be in the [RTPS spec](http://www.omg.org/spec/DDSI-RTPS/2.2/PDF/).
If we end up being able to use if it's not in the spec, would it make more sense to use `alignof` rather than `sizeof` here ?


I guess the other unknown is what we can do about the `space` variable that is set to 100

allenh1:
> If we end up being able to use if it's not in the spec, would it make more sense to use alignof rather than sizeof here ?

Maybe. Is that preferred?

> I guess the other unknown is what we can do about the space variable that is set to 100

I totally _spaced_ on that one. I'll fix that.

mikaelarguedas:
> > If we end up being able to use if it's not in the spec, would it make more sense to use alignof rather than sizeof here ?

> Maybe. Is that preferred?

TL;DR I think that if the `current_alignment` variable represents a size we should use `sizeof`, if it represent an alignment we should use `alignof`. In this specific case we use it on types where the 2 functions will return the same value so it doesn't matter that much. (rambling below)


If the alignment if not specified in the spec, I think we should use `alignof` to find the alignment. Though for primitive types like the ones used here it should give the same result as `sizeof`. But it would be more semantically correct and people wouldnt assume they can call `sizeof` on more complicated structures (where the result will differ quite a lot from `alignof`).

Though I didn't look at how this `current_alignment` is used, it would be valuable to know if it is supposed to represent an actual alignment or a size (because `calculateMaxSerializedSize` [returns](https://github.com/ros2/rmw_fastrtps/pull/140/files#diff-e31b7631040b83b65bc41128a06dcc02R817) arguably a size that is added to the `current_alignment`). According to what it represents we can chose the appropriate function to use.

allenh1:
> If the alignment if not specified in the spec, I think we should use alignof to find the alignment. Though for primitive types like the ones used here it should give the same result as sizeof. But it would be more semantically correct and people wouldnt assume they can call sizeof on more complicated structures (where the result will differ quite a lot from alignof).

I think you're right. Unfortunately, we just don't know what it is supposed to be. So, I think the best idea is to change it to `alignof` (since it makes the most sense to me), then spin up a CI job. If all is green on arm64, most of my concerns will disappear.

mikaelarguedas:
I would check the RTPS and CDR specs first though as suggested in https://github.com/ros2/rmw_fastrtps/pull/140#discussion_r132535322

allenh1:
> I would check the RTPS and CDR specs first though as suggested in #140 (comment)

Ok, sounds like a plan.

allenh1:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/432/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2429/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3103/)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -418,12 +418,13 @@ bool TypeSupport<MembersType>::serializeROSmessage(
...
           {
-            const MembersType * sub_members = (const MembersType *)member->members_->data;
+            const MembersType * sub_members =
+              static_cast<const MembersType *>(member->members_->data);
             void * subros_message = nullptr;
             size_t array_size = 0;
             size_t sub_members_size = sub_members->size_of_;
             size_t max_align = calculateMaxAlign(sub_members);
-            size_t space = 100;
+            size_t space = 1;  /* start with 1 byte */
```
dirk-thomas: Can you please clarify why `1` is being used here? The comment doesn't provide much information.

The comment should also be `//` instead of `/* ... */`.

allenh1: I picked `1` since everything in the system is based on octets, so that's the smallest unit of space one can have. That's the most sensible value I could come up with -- does this makes sense? I'm not very confident with it at `1`, but I think it's better than having it set to `100`.

dirk-thomas: As far as I can see the variable is being used as an "upper limit" on how much alignment can accumulate. I don't see why `1` would make sense for this?

dirk-thomas: Btw. there is a second location where that constant is being used: https://github.com/ros2/rmw_fastrtps/blob/af079efdc2636c79ca6f83de37a508b4b0cbdbe9/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L684

allenh1: > As far as I can see the variable is being used as an "upper limit" on how much alignment can accumulate. I don't see why 1 would make sense for this?

In that case, the value should be `4`. I'll update it as such.

> Btw. there is a second location where that constant is being used.

I'll update that as well. Thanks.

dirk-thomas:
> In that case, the value should be `4`. I'll update it as such.

Why?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
dirk-thomas: > In that case, the value should be `4`. I'll update it as such.

Why?

allenh1: If I'm understanding how DDS serialization works (which is a big if), that's how something of size 0 would be aligned.

dirk-thomas: > If I'm understanding how DDS serialization works (which is a big if), that's how something of size 0 would be aligned.

As far as I read the code this is not how this variable is being used.

mikaelarguedas:
@allenh1  any update on this ?

mikaelarguedas:
@allenh1 If there is no obvious path forward here (figuring out if we can get rid of space completely or a default that makes more sense than the current one), please change (or split) this PR so that the removal of the other magic numbers can be merged while we decide what to do with the space variable

allenh1:
@mikaelarguedas sounds good to me

allenh1:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/485/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2494/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3189/)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
mikaelarguedas: @allenh1 Did we come to a final answer on this ? should this use sizeof or should this be kept a hardcoded value based on the specs ?

mikaelarguedas:
> It appears windows failed with the connext middleware, so I don't think this failure is related to the PR.

Yes the failure is unrelated, this is a test known to be flaky on windows.

---

Can you please address / answer the comments ?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
allenh1: I think we agreed upon the sizeof, but I'll defer the call, as I am no expert.

mikaelarguedas: I didnt look at the spec. If you checked both specifications (DDS and RTPS) and that nothing specifies the padding I'm fine with merging this as is

allenh1: In DDS, the only statement about padding was dealing with the required 4 bytes. I'm not sure about RTPS, I'll have to get back to you.

allenh1:
@mikaelarguedas Ok, after looking at the RTPS spec, things are still fine (the sizes are recorded within the header).

Good to merge?

mikaelarguedas:
:+1: please squash and merge ;)

PR #141, opened 2017-08-10
Make sure to free memory associated with the service_name.

clalancette:

connects to ros2/ros2#387

CI jobs are in the above issue

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -233,6 +233,9 @@ fail:
     delete info;
   }
 
+  if (rmw_client->service_name != nullptr) {
+    rmw_free(const_cast<char *>(rmw_client->service_name));
+  }
   rmw_free(rmw_client);
 
   return NULL;
@@ -272,7 +275,8 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
     }
     delete info;
   }
-  delete (client);
+  rmw_free(const_cast<char *>(client->service_name));
+  rmw_free(client);
 
   return RMW_RET_OK;
 }
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -235,8 +235,9 @@ fail:
 
   if (rmw_client->service_name != nullptr) {
     rmw_free(const_cast<char *>(rmw_client->service_name));
+    rmw_client->service_name = nullptr;
   }
-  rmw_free(rmw_client);
+  rmw_client_free(rmw_client);
 
   return NULL;
 }
@@ -276,7 +277,8 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
     delete info;
   }
   rmw_free(const_cast<char *>(client->service_name));
-  rmw_free(client);
+  client->service_name = nullptr;
+  rmw_client_free(client);
 
   return RMW_RET_OK;
 }
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -276,8 +276,10 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
     }
     delete info;
   }
-  rmw_free(const_cast<char *>(client->service_name));
-  client->service_name = nullptr;
+  if (client->service_name != nullptr) {
+    rmw_free(const_cast<char *>(client->service_name));
+    client->service_name = nullptr;
+  }
   rmw_client_free(client);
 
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -222,6 +222,9 @@ rmw_create_client(const rmw_node_t * node,
     delete info;
   }
 
+  if (rmw_client->service_name != nullptr) {
+    rmw_free(const_cast<char *>(rmw_client->service_name));
```
dirk-thomas: Please set `rmw_client->service_name` to `nullptr` afterwards. While the variable isn't being used anymore it makes debugging much easier when the pointer isn't left with the same value but pointing to deallocated memory.

Same below.

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -261,7 +264,8 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
     }
     delete info;
   }
-  delete (client);
+  rmw_free(const_cast<char *>(client->service_name));
```
mikaelarguedas: should we check for if service_name is null before calling rmw_free on it ? (like its done above) Same below

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -261,7 +264,8 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
...
+  rmw_free(client);
```
dirk-thomas: This should use the symmetric function to `rmw_client_allocate` which is `rmw_client_free`.

Similar for the service below.

rmw_fastrtps_cpp/src/rmw_client.cpp
allenh1: https://github.com/ros2/rmw/blob/master/rmw/src/allocators.c#L37

Since the `rmw_free` function just calls free, I think it's ok to remove the null check around the one above.

clalancette: If you go through rmw_create_client() successfully, `service_name` is guaranteed to be non-NULL.  So the only way that service_name could be NULL in destroy is if someone hand-allocated a structure, set `service_name` to NULL, and then passed it to destroy (unlikely, but not impossible).  I'm not sure what our policy is in that situation, because quite a few other things can go wrong in that situation.

rmw_fastrtps_cpp/src/rmw_client.cpp
clalancette: Yeah, good point, will fix.

rmw_fastrtps_cpp/src/rmw_client.cpp
clalancette: Will fix this as well.

rmw_fastrtps_cpp/src/rmw_client.cpp
clalancette: > Since the rmw_free function just calls free, I think it's ok to remove the null check around the one above.

Yeah, in this instance, I agree with you, since free is guaranteed to do the correct thing on a NULL pointer.  However, if you think of rmw_free() as a "black box" function that you don't know anything about, then it is helpful to be conservative and not pass NULL into there.

mikaelarguedas: > if you think of rmw_free() as a "black box" function that you don't know anything about, then it is helpful to be conservative and not pass NULL into there.

+1 these functions are expected to be overriden by users wanting different memory strategies using alternative allocators. So I think it makes sense to sanity check it before

clalancette: I guess using that logic, I should check for NULL in the destroy() as well.  I'll do that.

rmw_fastrtps_cpp/src/rmw_service.cpp
```diff
@@ -235,6 +235,10 @@ rmw_create_service(const rmw_node_t * node,
     delete info;
   }
 
+  if (rmw_service->service_name != nullptr) {
+    rmw_free(const_cast<char *>(rmw_service->service_name));
+    rmw_service->service_name = nullptr;
+  }
   rmw_free(rmw_service);
```
dhood: `rmw_service_free` here too for symmetry, right?

clalancette: Arg, yes, missed that one.  Will fix, thanks.

clalancette:
Since rmw_service_free() is a thin wrapper around rmw_free(), I just did a local build to verify.  Things seem to be happy, so I'm going to assume that this won't change anything in CI.  I'm going to use the previous ACK from @dirk-thomas and merge this.  Thanks everyone!

PR #142, opened 2017-08-18
Release the condition mutex while detaching condition variable

dhood:
Otherwise if something triggers while we're detaching, it will deadlock waiting for the condition mutex

In particular, for the local_parameter tests in test_rclcpp, triggering of the graph guard condition can happen while detaching conditions.

I haven't seen an explanation for why the mutex was being unlocked while calling `detachCondition`. it was added in https://github.com/ros2/rmw_fastrtps/commit/4da558f6fcec1cc63fbd6b4b3dc4fee83b879913#diff-8225b098f1ec4debcdcee03d65ae95d5R1579 but from what I can see the condition mutex is not used by any `detachCondition`. Perhaps that was just a way to prevent it from happening _most_ of the time.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3037/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/440/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2442/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3120/)

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_wait.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_wait.cpp
@@ -182,15 +182,19 @@ rmw_wait(
     }
   }
 
+  // We're done with the condition variable now, but there's still a chance that something it was
+  // attached to will trigger and deadlock trying to lock the condition mutex.
+  // To avoid that we release it explicitly ASAP.
+  conditionVariable = NULL;
+  lock.unlock();
+
   for (size_t i = 0; i < subscriptions->subscriber_count; ++i) {
     void * data = subscriptions->subscribers[i];
     CustomSubscriberInfo * custom_subscriber_info = static_cast<CustomSubscriberInfo *>(data);
     if (!custom_subscriber_info->listener_->hasData()) {
       subscriptions->subscribers[i] = 0;
     }
-    lock.unlock();
     custom_subscriber_info->listener_->detachCondition();
-    lock.lock();
   }
 
   for (size_t i = 0; i < clients->client_count; ++i) {
@@ -199,9 +203,7 @@ rmw_wait(
     if (!custom_client_info->listener_->hasData()) {
       clients->clients[i] = 0;
     }
-    lock.unlock();
     custom_client_info->listener_->detachCondition();
-    lock.lock();
   }
 
   for (size_t i = 0; i < services->service_count; ++i) {
@@ -210,9 +212,7 @@ rmw_wait(
     if (!custom_service_info->listener_->hasData()) {
       services->services[i] = 0;
     }
-    lock.unlock();
     custom_service_info->listener_->detachCondition();
-    lock.lock();
   }
 
   if (guard_conditions) {
@@ -222,9 +222,7 @@ rmw_wait(
       if (!guard_condition->getHasTriggered()) {
         guard_conditions->guard_conditions[i] = 0;
       }
-      lock.unlock();
       guard_condition->detachCondition();
-      lock.lock();
     }
   }
   // Make timeout behavior consistent with rcl expectations for zero timeout value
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/rmw_wait.cpp
```diff
@@ -182,15 +182,19 @@ rmw_wait(
     }
   }
 
+  // We're done with the condition variable now, but there's still a chance that something it was
+  // attached to will trigger and deadlock trying to lock the condition mutex.
+  // To avoid that we release it explicitly ASAP.
+  conditionVariable = NULL;
```
dirk-thomas: Changing the local variable should have any effect. The condition variable is still attached to all the listeners. And locally it is not being used after this line.

dhood:
yep, that matches the situation as I observed/understand. I suppose my point was actually "I haven't seen an explanation for why the mutex was _only_ being unlocked while calling `detachCondition`, as opposed to just unlocking it once before detaching all conditions." As in, I don't see why it had to be locked the rest of the time. From what I can see it wasn't necessary, but I wanted to highlight my uncertainty for reviewers.

dirk-thomas:
The current code which unlock the condition before the call of `detachCondition` and then re-locks it again definitely looks fishy. I would assume that if the lock is necessary that approach opens the door for a race since the condition is not locked in between these lines.

The main question is when does `conditionMutex` need to be locked? Each listener uses the `conditionMutex_` to manage the access to the `data_` variable. It only needs to do that while `rmw_wait` is also trying to access the flag. When `rmw_wait` is not running the listener falls back to "unprotected" access of the `data_` member since it is the only entity using it. Therefore I think the `rmw_wait` functions needs to keep the lock while it is going through the loops which detach the conditions since each loop calls `hasData` on the listener. Based on that rational I don't think the `unlock` call can be move up as in the current patch since that would leave all `hasData` calls unprotected.

dhood:
> Based on that rational I don't think the `unlock` call can be move up as in the current patch since that would leave all `hasData` calls unprotected.

Re-stating to check my understanding: we don't want the listener's data to change between the time we finish waiting and the time we're checking it with `hasData`. To prevent that, we keep the `conditionMutex` locked so that listeners will not be able to update their data.

We must want to make sure that only the listener that notified the condition variable [gets set to 0](https://github.com/ros2/rmw_fastrtps/pull/142/files#diff-8044b1eed73db087f43f89fe2967b7ffR194), is that right? If so, that's the part I didn't realise. I thought that it was OK to not have the `conditionMutex` locked in `rmw_wait` when we're calling `hasData` on listeners, provided that they're locking their internal mutex / using atomics, if appropriate: worst case we end up with multiple listeners having `hasData` return true. But now it seems we _need_ only one to return true, yeah?

How about this: we keep the `conditionMutex` locked while we're doing all of the `hasData` checks. Then, we release the lock for good, and we detach all conditions. I think that this should satisfy (1) having the `conditionMutex` locked until we're done checking `hasData` for all listeners, and (2) having the `conditionMutex` unlocked while calling `detachCondition` to prevent the deadlock

dirk-thomas:
That sounds good to me and it doesn't leave any short term gaps where the mutex is unlocked and re-locked. :+1:

dirk-thomas:
Likely superseded by #147.

dhood:
update for posterity:

> We must want to make sure that only the listener that notified the condition variable [gets set to 0](https://github.com/ros2/rmw_fastrtps/pull/142/files#diff-8044b1eed73db087f43f89fe2967b7ffR194), is that right? If so, that's the part I didn't realise. I thought that it was OK to not have the `conditionMutex` locked in `rmw_wait` when we're calling `hasData` on listeners, provided that they're locking their internal mutex / using atomics, if appropriate: worst case we end up with multiple listeners having `hasData` return true. But now it seems we _need_ only one to return true, yeah?

I was initially correct that there's no issue with listeners updating their data after the wait() call, and there's no need to protect those calls with the condition mutex.

https://github.com/ros2/rmw_fastrtps/pull/147 includes this change in addition to changes to listeners to make the hasData calls threadsafe

PR #145, opened 2017-08-18
expose native participant

dirk-thomas:
This is an example how to expose native handles from a rmw implementation. It only exposes the `eprosima::fastrtps::Participant` but other handles could be supported in a similar fashion. The patch demonstrate how this is possible and that it can be achieved by each rmw implementation without requiring any changes in the `rmw` interface.

When the function is being used with a different rmw implementation at runtime `get_participant` returns `NULL`.

In order to try this feature I modified the `talker` in the `demo_nodes_cpp` package. The changes are straight forward. Only getting from the C++ `Node` to the `rmw_node_t` is not "very pretty":

* `package.xml`

```
<depend>rmw_fastrtps_cpp</depend>
```
* `CMakeLists.txt`

```
find_package(rmw_fastrtps_cpp REQUIRED)
...
ament_target_dependencies(talker "rmw_fastrtps_cpp")
```

* `talker.cpp`

```
#include "rmw_fastrtps_cpp/internal/get_participant.hpp"
...
// auto node = rclcpp::node::Node::make_shared("talker");
...
rcl_node_t * rcl_node = node->get_node_base_interface()->get_rcl_node_handle();
rmw_node_t * rmw_node = rcl_node_get_rmw_handle(rcl_node);
eprosima::fastrtps::Participant * p = rmw_fastrtps_cpp::get_participant(rmw_node);
```

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -52,6 +52,7 @@ include_directories(include)
 add_library(rmw_fastrtps_cpp
   src/client_service_common.cpp
   src/demangle.cpp
+  src/get_participant.cpp
   src/identifier.cpp
   src/namespace_prefix.cpp
   src/qos.cpp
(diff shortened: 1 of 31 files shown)


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3055/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/450/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2456/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3128/)

PR #148, opened 2017-08-24
export the imported targets

dirk-thomas:
Without this patch the exported `rmw_fastrtps_cpp_LIBRARIES` can't be used on Windows to link another target against FastRTPS. `rmw_fastrtps_cpp` itself doesn't use `FastRTPS_LIBRARIES` but links against the imported targets instead. Since this works this patch changes the exported libraries to contain the same. This will allow downstream packages to use `rmw_fastrtps_cpp` as intended.

Connect to ros2/demos#164.

Ready for review.


dirk-thomas:
I created #149 to track that this problem should be resolved if we can figure out how...

PR #155, opened 2017-10-04
Drop duplicated rmw_init.cpp in rmw_fastrtps_cpp/CMakeLists.txt

jwang11:
src/rmw_init.cpp appear twice in add_library(rmw_fastrtps_cpp, ...). Drop the duplicated one.


wjwwood:
Thanks!

PR #156, opened 2017-10-13
Enable logging level manipulation from rmw_fastrtps

sriramster:
* To enable the same, enable setting log levels from rmw_fastrtps layers,
* Add generic logging level parameters to 'rmw.h' which would be used by other
  layers above like rcl/rclcpp.
* Hide DDS specific implementation inside rmw_fastrtps.

Remove log level setting from rmw_create_node, as the idea of the API is to purely create a node than setting up fastrtps. 

Signed-off-by: Sriram Raghunathan <sriram.max@gmail.com>

Connect to ros2/rmw#124


dirk-thomas:
Functions which are not defined in the rmw interface should not be prefixed with `rmw_`.

In the related PR ros2/rmw#124 it doesn't add any function or argument to the rmw interface to pass a log severity. How do you plan to pass the parameter?

sriramster:
Need some help with it, I've not experimented with the other DDS implementations. This patch set, was specific to fastrtps. And, needed a review comment to check if its ok to add an rmw_ interface to set log levels from upper layers, which for now is specific only to fastrtps. Let me know.

dirk-thomas:
First the rmw interface needs to be updated, either by:

* adding a parameter to the `rmw_init` function or
* by adding a new function

Once we agree on the interface change all rmw implementations need to be updated in order to match the updated interface. Currently these are `rmw_fastrtps`, `rmw_connext`, `rmw_opensplice` and the package `rmw_implemenation`.

sriramster:
One reason why I avoided modifying rmw_init was, I'd a theoretical guess here that rmw_init would do more than just configuring logging parameters. So i'd thought. 

* Create a different function for setting up logging, and call this API from inside rmw_init(); 

Let me know if this goes with the idea of ros's core principles.

dirk-thomas:
> One reason why I avoided modifying rmw_init was, I'd a theoretical guess here that rmw_init would do more than just configuring logging parameters. So i'd thought.

That is true. We could certainly add a separate function for it. But if `rmw_init` and `rmw_set_log_level` are then always being called together then a parameter would be fine too. @ros2/team What do others prefer?

dhood:
If `rmw_set_log_level` were a separate function it'd be easier to call it if someone wants to change the level at runtime (assuming the vendor's logging supports runtime changes)

sriramster:
Adding, to the comment from @dhood , an API also gives the userspace ros application developer to control it from an application code.

mikaelarguedas:
Agreed with @dhood and @jwang11. An API would also give more granularity, I think in most cases a user would want to see the INFO output of a node without having all the INFO output of the underlying rmw_implementation.

dirk-thomas:
Sounds like we have a consensus to add a new function to the rmw interface to set the log level :+1:

> I think in most cases a user would want to see the INFO output of a node without having all the INFO output of the underlying rmw_implementation.

This is unrelated to the question how the parameter is passed to rmw.

sriramster:
The parameter needs to be set by a native call to rmw_implementation. Since doing it from elsewhere would change code at multiple layers.

sriramster:
@dirk-thomas Have updated the fastrtps specific logging here. Please let me know if these changes are fine.

rmw_fastrtps_cpp/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ add_library(rmw_fastrtps_cpp
   src/namespace_prefix.cpp
   src/qos.cpp
   src/rmw_init.cpp
+  src/rmw_fastrtps_configure.cpp
```
dirk-thomas: This file doesn't exist. This should probably be `rmw_logging.cpp`.

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "rmw/rmw.h"
+
+#include "fastrtps/log/Log.h"
+
+extern "C"
+{
+using eprosima::fastrtps::Log;
+
+eprosima::fastrtps::Log::Kind rmw_convert_severity_type(rmw_log_severity_t severity)
```
dirk-thomas: As mentioned before: functions not defined in the rmw interface should not be prefixed with `rmw_`.

sriramster:
@dirk-thomas Sorry for the errors. I've fixed them. I was'nt aware of testing, so missed it. I've run the test build as well. Let me know if anything else to be done.

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,56 @@
...
+ */
+  case RMW_LOG_SEVERITY_ERROR:
+  case RMW_LOG_SEVERITY_FATAL:
+  case RMW_LOG_SEVERITY_DEBUG:
+       return Log::Kind::Error;
+  default:
+       //Fallback to Info if undefined types
+       return Log::Kind::Info;
+  }
+}
```
dirk-thomas: This code block currently doesn't pass the linter tests. Please run the linters (either with `ament test` or directly `ament_uncrustify` / `ament_cpplint`) and address the style warnings.

dirk-thomas:
> Let me know if anything else to be done.

Currently the branch has conflicts and can't be merged. Please rebase your branch to be mergable with master.

Also are you planning on providing similar PRs for [rmw_connext](https://github.com/ros2/rmw_connext) and [rmw_opensplice](https://github.com/ros2/rmw_opensplice)? The set of PRs can only be merged once all currently supported rmw impl. implement the new function (even if the new function doesn't do anything for some rmw impl.).

sriramster:
@dirk-thomas Have fixed the issues pending. Sorry for this delay, this was my first patch so had to learn a few ros specific disciplines. Let me know if you've any task that I could take up purely as a student and contribute back. Thanks for the patience.

dirk-thomas:
Please do not make unrelated changes within the same pull request. Keep the focus of this PR on function to set the logging severity. 

Unrelated changes like the use of `RMW_CHECK_TYPE_IDENTIFIERS_MATCH` should go into separate pull requests.

sriramster:
@dirk-thomas  done. That was supposed to be my next pull request. Mixed it up.

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * Error : Max priority
+ * Warning : Medium priority
+ * Info : Low priority
+ * From rmw logging severity there are FATAL & DEBUG types as well
+ * We map them to ERROR type of Fast-RTPS which has maximum priority
+ */
+  case RMW_LOG_SEVERITY_ERROR:
+  case RMW_LOG_SEVERITY_FATAL:
+  case RMW_LOG_SEVERITY_DEBUG:
+       return Log::Kind::Error;
```
dirk-thomas: I don't think that `Log::Kind::Error` is a reasonable mapping for `RMW_LOG_SEVERITY_DEBUG`.

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,56 @@
...
+       // Fallback to Info if undefined types
+       return Log::Kind::Info;
+  }
+}
+
+rmw_ret_t
+rmw_set_log_severity(rmw_log_severity_t severity)
+{
+  Log::Kind _severity = convert_rmw_severity_type(severity);
+  eprosima::fastrtps::Log::SetVerbosity(_severity);
```
dirk-thomas: If the severity can't be mapped (the `default` case in the other function) it might be better to print an error message rather than setting a fallback level without letting the user know.

sriramster: @dirk-thomas  ok. In that case I'll have to change the API to return a type rmw_ret_t. And, may I use RMW_SET_ERROR_MSG for error setting on unknown types? There has been instances where inside rmw RCUTILS macro for error logging has been used. Let me know on this.

sriramster: @dirk-thomas Could you let me know your thoughts on the previous comment?

dirk-thomas: Since this function is private, only being used once in the same .cpp file, and the calling function is only a couple of lines ong I would suggest to just remove the separate function and move the implementation into `rmw_set_log_severity`. That would simplify the code. `rmw_set_log_severity` can then indeed return an error code and set the error message accordingly.

sriramster: Ok. Thanks. I'll make the necessary changes and update the commit.

sriramster:
@dirk-thomas Sorry about the delay in addressing the review comments. I've pushed in the changes with latest review comments. Please have a look.

sriramster:
@dirk-thomas  Could you have a look at the updated pull request?

sriramster:
@dirk-thomas Hi Dirk, Could you have a look the above commit? And let me know if this is ok.

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,58 @@
...
+// to ERROR type of Fast-RTPS which has maximum priority
+    case RMW_LOG_SEVERITY_DEBUG:
+      _severity = Log::Kind::Warning;
+      break;
+    case RMW_LOG_SEVERITY_ERROR:
+    case RMW_LOG_SEVERITY_FATAL:
+      _severity = Log::Kind::Error;
+      break;
+    default:
+      RMW_SET_ERROR_MSG("node handle is null");
```
dirk-thomas: This error message seems to not match the case.

sriramster: @dirk-thomas Updated the commit with logging msg. Please have a look.

sriramster:
@dirk-thomas Could you review the latest change set?

rmw_fastrtps_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,58 @@
...
+      _severity = Log::Kind::Info;
+      break;
+// Fast-RTPS supports the following logging 'Kind's.
+// Error : Max priority
+// Warning : Medium priority
+// Info : Low priority
+// From rmw logging severity there is FATAL severity type we map it
+// to ERROR type of Fast-RTPS which has maximum priority
+    case RMW_LOG_SEVERITY_DEBUG:
+      _severity = Log::Kind::Warning;
```
dirk-thomas: This mapping still doesn't make any sense. The expectation is that `DEBUG` is more verbose than `INFO`. Therefore mapping it to `Warning` were `INFO` is mapped to `Info` seems wrong to me. I think this needs to be `Info` instead.

sriramster:
@dirk-thomas  Thanks. Had a theoretical reasoning mapping Debug to Warning type of fastrtps, like mentioned Warning was verbose than info. And felt debug messages needed to be more verbose. Anyways fixed it like you suggested. Please have a look.

dirk-thomas:
I refactored the patch slightly in a01006af174b956dff2bd6d49eb5a4466eab95d9. This looks good to me now.

I will wait with merging this for the similar PRs on the other rmw impl.

sriramster:
@dirk-thomas  Thank you.

PR #162, opened 2017-10-20
use auto deduction and nullptr to keep coding style consistent

jwang11:
This commit use auto to deduce complex TYPE and replace NULL with nullptr. It
helps to keep coding style consistent across the project

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -69,7 +69,7 @@ rosidl_generator_c__void__Array__init(
   if (!array) {
     return false;
   }
-  void * data = NULL;
+  void * data = nullptr;
   if (size) {
     data = static_cast<void *>(calloc(size, member_size));
     if (!data) {
@@ -94,7 +94,7 @@ rosidl_generator_c__void__Array__fini(rosidl_generator_c__void__Array * array)
     assert(array->capacity > 0);
     // finalize all array elements
     free(array->data);
-    array->data = NULL;
+    array->data = nullptr;
     array->size = 0;
     array->capacity = 0;
   } else {
@@ -186,7 +186,7 @@ static size_t calculateMaxAlign(const MembersType * members)
           break;
         case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
           {
-            const MembersType * sub_members = (const MembersType *)member.members_->data;
+            auto sub_members = (const MembersType *)member.members_->data;
             alignment = calculateMaxAlign(sub_members);
           }
           break;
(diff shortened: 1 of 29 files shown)


mikaelarguedas:
Thanks @jwang11, we never took time of homogenizing the style since we took ownership of this repo. Appreciated.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3372/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/623/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2705/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3425/)

dirk-thomas:
`auto` makes the code shorter in many cases without reducing the readability, e.g.:

```
auto impl = static_cast<CustomClientInfo *>(client->data);
```

But in some cases I think it would be beneficial to keep the specific type since `auto` "removes" some of the information, e.g.;

```
rcutils_ret_t ret = _assign_partitions_to_attributes(...);
```

mikaelarguedas:
> But in some cases I think it would be beneficial to keep the specific type since auto "removes" some of the information, e.g.;
>```
>rcutils_ret_t ret = _assign_partitions_to_attributes(...);
>```

That's a fair point. When reviewing I considered it ok if the information could be extracted from the next line easily. But I think you're right, people not familiar with the codebase should be able to have all the context by reading only the line in question without having to check what other functions return or knowing that an `RMW_RET_OK` is an enum of type `rmw_ret_t`. 
@jwang11 would you mind reducing the scope to use auto only for the cases where readibility is not impacted?

jwang11:
yes, I agree auto usage should not impact readability an actually I have kept some std::xxx type there. I'll double check the commit.

jwang11:
conflict is resolved, please continue review

mikaelarguedas:
Baased on previous comments, it would be good to reduce the scope of this and not use auto for return codes types and `const rosidl_message_type_support_t *` that are not trivial to deduce from the line itself

jwang11:
For "const rosidl_message_type_support_t *" , it really should be kept, but
Return code type can be easily reduced from next line, do you think it impact readability?
  auto ret = rmw_names_and_types_check_zero(topic_names_and_types);
  if (ret != RMW_RET_OK)

mikaelarguedas:
We don't really have a policy for this, I think the conclusion from the previous discussion is that the line itself should provide enough context to know what the returned type is. I would be fine having auto if the next line provides the context but also understand the reasoning that a single line should be enough to have all the information, maybe @ros2/team wants to pitch in on this

dirk-thomas:
From my point of view using `auto ret` is problematic because the following line `if (ret != RMW_RET_OK)` could either be correct (since `ret` is of type `rmw_ret_t`) or it could be wrong (if `ret` would be of type `rcl_ret_t`). And since that is not obvious anymore I would like to keep the strong explicit typing on that kind of variables.

clalancette:
I agree with @dirk-thomas on this; I'd prefer to be more explicit on the return variables.

jwang11:
So, sounds reasonable to keep explicit return variable. I am fine to update the commit.

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3395/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/640/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3442/)

PR #164, opened 2017-10-25
pointers are checked for null

dejanpan:
Closes https://github.com/ros2/rmw_connext/issues/260.

Bugfix, defensive programming.


dirk-thomas:
Thanks for the patch.

PR #165, opened 2017-10-27
Fix the issues to dereference to nullptr

gaoethan:
_Issue_: the `info` in `fail` is always non-nullptr
_issue_: no checking to memory allocation of `rmw_client`, which
       may leads to free `nullptr` in the following code:

`  rmw_client_free(rmw_client);`

_Issue_: uninteded memory free from `rmw_client->service_name`:

```
  if (rmw_client->service_name != nullptr) {
    rmw_free(const_cast<char *>(rmw_client->service_name));
    rmw_client->service_name = nullptr;
  }
```

when the code goes to `fail` before it's actually allocated here:

```
  rmw_client->service_name = reinterpret_cast<const char *>(
    rmw_allocate(strlen(service_name) + 1));
```

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -191,6 +191,11 @@ rmw_create_client(
   info->writer_guid_ = info->request_publisher_->getGuid();
 
   rmw_client = rmw_client_allocate();
+  if (!rmw_client) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for client");
+    goto fail;
+  }
+
   rmw_client->implementation_identifier = eprosima_fastrtps_identifier;
   rmw_client->data = info;
   rmw_client->service_name = reinterpret_cast<const char *>(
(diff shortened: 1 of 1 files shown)


gaoethan:
Based on this PR, I drop another potential change to give a "default constructor" to `struct CustomClientInfo` just like the one in [struct CustomClientResponse](https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp#L48-L49) which initialize its struct member pointer to `nullptr`. It is because the compiler defaults to initialize it when run the following code:

`info = new CustomClientInfo();`

but in case that the compiler doesn't do this, the access and memory free for the `info` member(`request_publiser_`, `response_subscriber_` and `listener_`) in `fail` is "Segmentation fault"  when the code goes to `fail` before they're actually allocated in their later code. any comment ?

dirk-thomas:
> _Issue_: the `info` in `fail` is always non-nullptr

I don't think this is an "issue". I would consider this defensive programming. If a `goto` is ever introduced a bit earlier in the function it is just safer to check `if (info) {` before trying to use it with `info->...`. Therefore I would propose to keep the check.

gaoethan:
@dirk-thomas it makes sense to avoid future errors and I've restored the checking for `info`. thanks.

gaoethan:
@clalancette thanks for your comment, I agree with you that a helper function to check and free the info can make it shared among the others, as per my observation, it also has this kind of optimization in the other components, so my thinking is whether it's possible to generalize a helper API for all of them to make this kind of behaviour unified, of course, this may involve more changes and it's better to address them in another PR if it works. :)

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3399/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/644/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2728/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3446/)

dirk-thomas:
Thank you for the improvement.

I will move ahead with the merge even though the MacOS build hasn't been triggered yet (due to problems with the connectivity of these machines).

Issue #166, opened 2017-10-27
rmw_count_publisher|subscriber - excessive string allocation


deng02:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - source
- Version or commit hash:
  - HEAD
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue
You can reproduce this in multiple ways but the easiest is to create a large number of ROS2 publishers and subscribers and run the dynamic bridge, which counts the ROS2 pub/subs every second. 

#### Behavior
CPU utilization is higher than expected.  Profiling shows a hotspot with operator new and cfree being called, which backtracks to the functions rmw_count_publishers and rmw_count_subscribers being called in the dnamic bridge. Specifically the pushback into the unfiltered_topics vector e.g. in rmw_count_subscribers:

```
std::map<std::string, std::vector<std::string>> unfiltered_topics;
  ReaderInfo * slave_target = impl->secondarySubListener;
  slave_target->mapmutex.lock();
  for (auto it : slave_target->topicNtypes) {
    for (auto & itt : it.second) {
      // truncate the ROS specific prefix
      auto topic_fqdn = _demangle_if_ros_topic(it.first);
      unfiltered_topics[topic_fqdn].push_back(itt);
    }
  }
  slave_target->mapmutex.unlock()
```

Similar code exists for the rmw_count_publishers function.

Code inspection shows that the vector values held by the unfiltered_topics map aren't used at all and that it's just the number we care about.  Further, all of the topic_fqdn keys aren't really necessary since we're only really interested in one.  The function only uses it for the purposes of a debug message.

#### Proposed fix

To reduce string allocation, drop the non-matching topic_fqdn keys and don't bother pushing the topic types onto the vector.  Just count the number of topic types for matching topics and return.

dirk-thomas:
Addressed by #167.

@deng02 If you use "Fixes #166" in the description of the pull request it will automatically close the referenced issue when the PR is being merged.

deng02:
@dirk-thomas I'll definitely do that next time!  Thanks for the review/merge.

PR #167, opened 2017-10-27
Remove string allocation in the count of subscribers and publishers

deng02:
Addresses https://github.com/ros2/rmw_fastrtps/issues/166

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_count.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_count.cpp
@@ -48,37 +48,21 @@ rmw_count_publishers(
 
   CustomParticipantInfo * impl = static_cast<CustomParticipantInfo *>(node->data);
 
-  std::map<std::string, std::vector<std::string>> unfiltered_topics;
   WriterInfo * slave_target = impl->secondaryPubListener;
   slave_target->mapmutex.lock();
+  *count = 0;
   for (auto it : slave_target->topicNtypes) {
-    for (auto & itt : it.second) {
-      // truncate the ROS specific prefix
-      auto topic_fqdn = _demangle_if_ros_topic(it.first);
-      unfiltered_topics[topic_fqdn].push_back(itt);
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/rmw_count.cpp
```diff
@@ -48,34 +48,19 @@ rmw_count_publishers(
 
   CustomParticipantInfo * impl = static_cast<CustomParticipantInfo *>(node->data);
 
-  std::map<std::string, std::vector<std::string>> unfiltered_topics;
```
clalancette: I think you need to set `*count` to 0 here (like you do in `rmw_count_subscribers`), no?

deng02: Yes, I'll fix this.

deng02: PR Updated

rmw_fastrtps_cpp/src/rmw_count.cpp
```diff
@@ -48,34 +48,20 @@ rmw_count_publishers(
...
     "rmw_fastrtps_cpp",
     "looking for subscriber topic: %s", topic_name)
-  for (auto it : unfiltered_topics) {
-    RCUTILS_LOG_DEBUG_NAMED(
-      "rmw_fastrtps_cpp",
-      "available topic: %s", it.first.c_str())
-  }
   RCUTILS_LOG_DEBUG_NAMED(
     "rmw_fastrtps_cpp",
     "number of matches: %zu", *count)
```
dirk-thomas: Should those two log messages be merged into one now?

Same below.

deng02: Agreed.  I'll collapse this.

deng02: PR updated

dirk-thomas:
Thank you for the patch and the quick updates.

PR #168, opened 2017-10-30
Avoid duplicated code in calculateMaxSerializedSize for array and nor…

jwang11:
Treat normal member as the case that array_size = 1, it unify the calculation code

PR #171, opened 2017-10-31
Unify and simplify de/serializeROSmessage processing

jwang11:
For most type_id, there is no difference for serialization which can be unified in a common template function. However, for bool and string, there is extra check or logic for normal but array does not. Although it is weird, this commit keep the logic intact.

Signed-off-by: jwang <jing.j.wang@intel.com>

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -203,12 +203,14 @@ static size_t calculateMaxAlign(const MembersType * members)
 
 // C++ specialization
 template<typename T>
-void serialize_array(
+void serialize_field(
   const rosidl_typesupport_introspection_cpp::MessageMember * member,
   void * field,
   eprosima::fastcdr::Cdr & ser)
 {
-  if (member->array_size_ && !member->is_upper_bound_) {
+  if (!member->is_array_) {
+    ser << *static_cast<T *>(field);
+  } else if (member->array_size_ && !member->is_upper_bound_) {
     ser.serializeArray(static_cast<T *>(field), member->array_size_);
   } else {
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
This is a follow up of https://github.com/ros2/rmw_fastrtps/pull/168 now that most of the logic for array has been simplified and magic numbers removed.

Thanks @jwang11 we'll try to get this reviewed soon

mikaelarguedas:
> Although it is weird, this commit keep the logic intact.

Yes this extra logic is necessary, thanks for keeping it as is.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/659/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2742/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3459/)

mikaelarguedas:
While it's counter intuitive to call (de)serializeArray for non array types I appreciate the code deduplication.

I'll leave a chance to @ros2/team to voice any concern before merging

dirk-thomas:
Before the code was calling the following for non-array fields:

```
ser << *static_cast<uint8_t *>(field);
```

With the patch it is calling this line instead:

```
ser.serializeArray(static_cast<uint8_t *>(field), 1);
```

I don't think we can rely on them being equivalent without making assumptions about how `serializeArray` is implemented. And that could change anytime. Therefore this change doesn't look like a good idea to me.

jwang11:
```
ser.serializeArray(static_cast<uint8_t *>(field), 1);
```
Ok, If it introduce the dependency on specific FastRTPS implementation, I can update the commit to handle non-array type separately, but still in same function.

jwang11:
Code updated. Separate the process for single field and array according to comments, consolidate them in central function which obviously help code deduplication.

dirk-thomas:
Thank you for improving the code.

PR #178, opened 2017-12-05
use newly exposed unread-count to avoid getting out-of-sync

dirk-thomas:
This fixes ros2/rclcpp#280 and requires the recent API extension from eProsima/Fast-RTPS#166.

The following set of builds only shows that it doesn't introduce a regression but doesn't prove that it actually addresses the problem:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3733/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/888/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3071/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3830/)

Connect to ros2/rclcpp#280.


dirk-thomas:
> Can we add a regression test for that so that we make sure all vendors behave the same when we exceed the queue size ?

I created a ticket for it since I don't think we have time today before the release: ros2/rclcpp#421

PR #180, opened 2017-12-11
Segmentation error to dereference nullptr

gaoethan:
`getEDPReaders` is possible to return `nullptr`

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_node.cpp
@@ -127,13 +127,18 @@ create_node(
   node_impl->secondaryPubListener = tnat_2;
 
   edp_readers = participant->getEDPReaders();
-  if (edp_readers.first && edp_readers.second) {
-    if (!(edp_readers.first->setListener(tnat_1) & edp_readers.second->setListener(tnat_2))) {
-      RMW_SET_ERROR_MSG("Failed to attach ROS related logic to the Participant");
-      goto fail;
-    }
-  } else {
-    RMW_SET_ERROR_MSG("Failed to get valid reader for node subscriber and publisher");
+  if (!edp_readers.first) {
+    RMW_SET_ERROR_MSG("edp_readers.first is null");
+    goto fail;
+  }
+
+  if (!edp_readers.second) {
+    RMW_SET_ERROR_MSG("edp_readers.second is null");
+    goto fail;
+  }
+
+  if (!(edp_readers.first->setListener(tnat_1) & edp_readers.second->setListener(tnat_2))) {
+    RMW_SET_ERROR_MSG("Failed to attach ROS related logic to the Participant");
     goto fail;
   }
 
--- a/rmw_fastrtps_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_node.cpp
@@ -127,8 +127,13 @@ create_node(
   node_impl->secondaryPubListener = tnat_2;
 
   edp_readers = participant->getEDPReaders();
-  if (!(edp_readers.first->setListener(tnat_1) & edp_readers.second->setListener(tnat_2))) {
-    RMW_SET_ERROR_MSG("Failed to attach ROS related logic to the Participant");
+  if (edp_readers.first && edp_readers.second) {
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -127,8 +127,13 @@ create_node(
   node_impl->secondaryPubListener = tnat_2;
 
   edp_readers = participant->getEDPReaders();
-  if (!(edp_readers.first->setListener(tnat_1) & edp_readers.second->setListener(tnat_2))) {
-    RMW_SET_ERROR_MSG("Failed to attach ROS related logic to the Participant");
+  if (edp_readers.first && edp_readers.second) {
```
Karsten1987: maybe just my personal taste, but I would proof every potentially failing test individually. This makes the error message less ambiguous whether condition 1 or condition 2 fails.  
Also, I like the style of testing for negative first and return/goto fail early. 
```
if (!edp_readers.first) {
  RMW_SET_ERROR_MSG("edp_readers.first is null");
  got fail;
}
...
```

But then again, only my 2 cents.

rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -272,12 +277,12 @@ rmw_destroy_node(rmw_node_t * node)
 
   // Begin deleting things in the same order they were created in rmw_create_node().
   std::pair<StatefulReader *, StatefulReader *> edp_readers = participant->getEDPReaders();
-  if (!edp_readers.first->setListener(nullptr)) {
+  if (!edp_readers.first || !edp_readers.first->setListener(nullptr)) {
```
Karsten1987: same comment as above of testing the conditions individually.

rmw_fastrtps_cpp/src/rmw_service.cpp
```diff
@@ -237,7 +241,7 @@ rmw_create_service(
     delete info;
   }
 
-  if (rmw_service->service_name != nullptr) {
+  if (rmw_service && rmw_service->service_name != nullptr) {
```
Karsten1987: it looks to me like a redundant test, because `rmw_service` was checked for null in line 202. Also, try to use the same notation of `(!rmw_service->service_name)` instead of `!= nullptr`

gaoethan: Actually, if it goes to `fail` before `rmw_service` is really initialized (a. k. a assigned to a valid address at line 201), even though the `rmw_service` is `nullptr`, but `rmw_service->service_name` is actually not **0x0** and it should be some value like **0x10**, then there will be "Segmentation fault (core dumped)" error

rmw_fastrtps_cpp/src/rmw_node.cpp
gaoethan: Thanks for the suggestion, why I put them together is that:
1. their definition is a std::pair<> a. k. a std::pair<StatefulReader *, StatefulReader *> edp_readers;
2. getEDPReaders() returns them both NULL at the same time and no case to return one of them NULL

rmw_fastrtps_cpp/src/rmw_service.cpp
Karsten1987: please apologize, I haven't noticed the modified condition was inside the `fail`. Your change is completely correct here.

rmw_fastrtps_cpp/src/rmw_node.cpp
Karsten1987: Looking at the current documentation of this function [here](https://github.com/eProsima/Fast-RTPS/blob/master/include/fastrtps/participant/Participant.h#L86-L94), one could argue there is no notion that they are always returned NULL - even though this might be the case in the current verison. However, this implementation may change over time, so that we can't certainly rely on this assumption.

Karsten1987:
Running CI on this:

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3793/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/3129/)
WIN: [Build Status](http://ci.ros2.org/job/ci_windows/3879/)

Karsten1987:
thanks for iterating on this. Merging!

PR #183, opened 2017-12-29
Small performance improvments

guillaumeautran:
Address small performance issues to ROS2 impacting the `rmw_count_publishers` and `rmw_count_subscribers` method when called with a significante number of topics to count.
The root of the performance issues come from the duplication of `std::string` type of objects. These duplication are honerous with strings and cause a significante amount of CPU cycle to be wasted duplicating things.

### DIFF
--- a/rmw_fastrtps_cpp/src/namespace_prefix.cpp
+++ b/rmw_fastrtps_cpp/src/namespace_prefix.cpp
@@ -32,10 +32,22 @@ const std::vector<std::string> _ros_prefixes =
 std::string
 _get_ros_prefix_if_exists(const std::string & topic_name)
 {
-  for (auto prefix : _ros_prefixes) {
-    if (topic_name.rfind(std::string(prefix) + "/", 0) == 0) {
+  for (const auto & prefix : _ros_prefixes) {
+    if (topic_name.rfind(prefix, 0) == 0 && topic_name.at(prefix.length()) == '/') {
       return prefix;
     }
   }
   return "";
+}
+
+/// Strip the ROS specific prefix if it exists from the topic name.
+std::string
+_strip_ros_prefix_if_exists(const std::string & topic_name)
+{
+  for (const auto & prefix : _ros_prefixes) {
+    if (topic_name.rfind(prefix, 0) == 0 && topic_name.at(prefix.length()) == '/') {
+      return topic_name.substr(prefix.length());
+    }
+  }
+  return topic_name;
 }
--- a/rmw_fastrtps_cpp/src/rmw_count.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_count.cpp
@@ -51,8 +51,8 @@ rmw_count_publishers(
   WriterInfo * slave_target = impl->secondaryPubListener;
   slave_target->mapmutex.lock();
   *count = 0;
-  for (auto it : slave_target->topicNtypes) {
-    auto topic_fqdn = _demangle_if_ros_topic(it.first);
+  for (const auto & it : slave_target->topicNtypes) {
+    const auto topic_fqdn = _demangle_if_ros_topic(it.first);
     if (topic_fqdn == topic_name) {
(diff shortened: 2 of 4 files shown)


guillaumeautran:
FYI: @dirk-thomas @deng02

clalancette:
The changes look reasonable to me.  I've kicked off CI here:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3818/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/957/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3158/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3905/)

guillaumeautran:
Fixed the codestyle issues. Please re-kick the CI build.

clalancette:
Here you go:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3820/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/959/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3160/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3907/)

rmw_fastrtps_cpp/src/namespace_prefix.cpp
```diff
@@ -29,13 +29,25 @@ const std::vector<std::string> _ros_prefixes =
 }  // extern "C"
 
 /// Return the ROS specific prefix if it exists, otherwise "".
-std::string
+const std::string
```
dirk-thomas: Since the return type is a value and not a pointer or reference adding `const` doesn't make any sense.

Same below.

guillaumeautran: That's correct (checked with the profiler). I've fixed that and uploaded the change.

rmw_fastrtps_cpp/src/namespace_prefix.cpp
```diff
@@ -32,10 +32,22 @@ const std::vector<std::string> _ros_prefixes =
 std::string
 _get_ros_prefix_if_exists(const std::string & topic_name)
 {
-  for (auto prefix : _ros_prefixes) {
-    if (topic_name.rfind(std::string(prefix) + "/", 0) == 0) {
+  for (const auto & prefix : _ros_prefixes) {
+    if ((topic_name.rfind(prefix, 0) == 0) && (topic_name.at(prefix.length()) == '/')) {
```
dirk-thomas: I would suggest to remove the extra parenthesis around each of the two boolean conditions.

Same below.

rmw_fastrtps_cpp/src/rmw_count.cpp
```diff
@@ -51,8 +51,8 @@ rmw_count_publishers(
   WriterInfo * slave_target = impl->secondaryPubListener;
   slave_target->mapmutex.lock();
   *count = 0;
-  for (auto it : slave_target->topicNtypes) {
-    auto topic_fqdn = _demangle_if_ros_topic(it.first);
+  for (const auto & it : slave_target->topicNtypes) {
+    auto const topic_fqdn = _demangle_if_ros_topic(it.first);
```
dirk-thomas: Please use `const auto` / `auto const` consistently throughout the code base.

rmw_fastrtps_cpp/src/namespace_prefix.cpp
guillaumeautran: Done.

rmw_fastrtps_cpp/src/rmw_count.cpp
guillaumeautran: I started the other way around and missed updating these two.
Fixed now

dirk-thomas:
This looks good to me. Thank you for the patch and for iterating on it.

PR #192, opened 2018-03-23
Remove topic partitions

rohitsalem:
connects to [ros2/ros2#476](https://github.com/ros2/ros2/issues/476)
To use the topic/service name directly without using the partitions.


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4140/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1218/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4245/)

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4141/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1219/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3423/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4246/)

rohitsalem:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/4152/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1233/) 
*  macOS [Build Status](https://ci.ros2.org/job/ci_osx/3434/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/4256/)

rohitsalem:
Connects to the issue [ros2/ros2#476](https://github.com/ros2/ros2/issues/476)

mikaelarguedas:
@rohitsalem if you want the issues/PRs to be connected on waffle, you need to put it the "connects to <link to issue>" sentence in the description of the issue (the first comment) as waffle doesn't parse all the comments. It also needs to be the github representation of the link and not the full link.
You can find more details on the [waffle section of the onboarding guide](https://github.com/ros2/ros2/wiki/ROS-2-On-boarding-Guide#waffleio-how-to)

rohitsalem:
@mikaelarguedas  thanks for letting me know! edited my first comment on this PR

mikaelarguedas:
You're welcome, note that you don't need to explicitly provide the full link, Github will recognize `ros2/ros2#476` and link to `https://github.com/ros2/ros2/issues/476` automatically

rohitsalem:
Yeah, got it. That way I thought it would be easier for someone to navigate to the issue from here.

rohitsalem:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/4175/)
* Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1247/)
* macOS: [Build Status](https://ci.ros2.org/job/ci_osx/3459/)
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/4272/)

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -124,25 +122,23 @@ rmw_create_client(
   subscriberParam.topic.topicKind = NO_KEY;
   subscriberParam.topic.topicDataType = response_type_name;
   subscriberParam.historyMemoryPolicy = PREALLOCATED_WITH_REALLOC_MEMORY_MODE;
-  rcutils_ret_t ret = _assign_partitions_to_attributes(
-    service_name, ros_service_response_prefix,
-    qos_policies->avoid_ros_namespace_conventions, &subscriberParam);
-  if (ret != RCUTILS_RET_OK) {
-    // error msg already set
-    goto fail;
+  if (!qos_policies->avoid_ros_namespace_conventions){
```
mikaelarguedas: Style: Indentation and spacing seems to not be matching the guidelines. You can find style violations by running the tests of the package locally

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -124,25 +122,23 @@ rmw_create_client(
...
+  subscriberParam.topic.topicName = std::string(ros_service_response_prefix) + service_name;
```
mikaelarguedas: I think there should be a "/" used to separate the ros prefix and the service/topic name (at least that's what the code in https://github.com/ros2/rmw_fastrtps/pull/192/files#diff-0ce60ac271e069c6686c50d38559e0a5L42 was doing for the string.
e.g. ROS topic `foo` => DDS topic `rt/foo`

rohitsalem: The code in https://github.com/ros2/rmw_fastrtps/pull/192/files#diff-0ce60ac271e069c6686c50d38559e0a5L42 is first extracting the base topic_name from the whole topic name (which includes the namespaces), by taking whatever is there after the last '/'. And then re-adding prefix to the base topic_name along with  '/'. So it will have ros prefix + '/' + base name. But I guess here the default(complete) topic name which we receive will have a "/", isn't it?  example: we will have `/chatter` by default rather than  `chatter`, otherwise I'll have add a check to check if there is a `/` by default or not and then add it to the topic name.

rmw_fastrtps_cpp/src/rmw_client.cpp
rohitsalem: Thanks for the catch. I've updated it in this commit https://github.com/ros2/rmw_fastrtps/pull/192/commits/0a516ea6f5528e7ce38a7ae40379cd778e89ff33

rmw_fastrtps_cpp/src/rmw_client.cpp
mikaelarguedas: > But I guess here the default(complete) topic name which we receive will have a "/", isn't it?

Apparently if called from rcl, it always has a leading forward slash as [rcl enforces it](https://github.com/ros2/rcl/blob/d41c923927850274f63aee1bde565d3a8266a7bd/rcl/src/rcl/publisher.c#L148) before passing the name to rmw.

Not sure if we can make that assumption for people calling rmw API directly though.
So my guess (maybe someone from @ros2/team more familiar with the matter can pitch in) is that, if `avoid_ros_namespace_conventions` is false, we should check for the presence of the slash and add it if needed.
If `avoid_ros_namespace_conventions` is true, we should not modify the name and let user creat topics with the nam he specified.

wjwwood: By the time it is passed into `rmw` all topics will be fully qualified, which includes a leading `/`.

mikaelarguedas: > By the time it is passed into rmw all topics will be fully qualified

I agree with that, I was just wondering if it's safe to make this assumption here for people calling this API directly (and not via rcl).
I guess the goal here is to avoid adding the namespace prefix but still assume ROS naming conventions are respected. Does that mean that there is no way to subscribe to a DDS topic named e.g. "image" using this API ? only "/image" ?

wjwwood: > I was just wondering if it's safe to make this assumption here for people calling this API directly (and not via rcl).

No one should do that 😄, but if I were to write the documentation for these rmw functions now I'd say it is expected that the topic name is good as-is and needs no modification. So basically, prefix with `rt__` or w/e if `avoid_ros_namespace_conventions` is true, otherwise do nothing to the topic name.

Separately the use case of subscribing to a DDS topic `foo` rather than `/foo`, should be handled. I hadn't thought about that, because in the previous case you simply used the base name you wanted and that was it. There were ideas in the design document about how to handle that, but nothing was implemented:

http://design.ros2.org/articles/topic_and_service_names.html#examples-and-ideas-for-communicating-with-non-ros-topics

Even if you were to strip leading `/` when `avoid_ros_namespace_conventions` is true, you still would not be able to specify a partition name.

mikaelarguedas: >  So basically, prefix with rt__ or w/e if avoid_ros_namespace_conventions is true, otherwise do nothing to the topic name.

Sounds good to me

> Even if you were to strip leading / when avoid_ros_namespace_conventions is true, you still would not be able to specify a partition name.

Fair enough.

Ideally people could use whatever is exposed in the API, in this case we don't expose partitions to users, so users would only be able to be matched with a topic that has matching QoS (so only if empty partitions on both sides).
We can definitely postpone the discussion on how to enable users to subscribe to pure DDS topics :+1:

rohitsalem: So are we deciding on leaving it as is? This will affect the topic names coming in the future which don't start with a "/". Or some remapping might be required to append a "/' for those topic or service names.

mikaelarguedas: As we don't have any currently supported use case where the topic will not enter this function without a leading slash we keep it as is for now :+1: . If the use case comes up we'll adapt the behavior.

wjwwood: I was thinking about this more, and I suppose we could choose to pass the unexpanded topic name directly to the rmw layer if the `avoid_ros_namespace_conventions` is set to true.

For example, we currently:

- pass topic name into `rcl_create_publisher()` (for example), e.g. `image`
- expand topic name with `rcl_expand_topic_name()`, e.g. `image` -> `/image`
- store expanded topic name in `publisher.topic_name`
- pass `publisher.topic_name` to `rmw_create_publisher()`

And instead, iff `avoid_ros_namespace_conventions` is true we could:

- pass topic name into `rcl_create_publisher()` (for example), e.g. `image`
- ~~expand topic name with `rcl_expand_topic_name()`, e.g. `image` -> `/image`~~
- store ~~expanded~~ topic name in `publisher.topic_name`
- pass `publisher.topic_name` to `rmw_create_publisher()`
  - `rmw_create_publisher()` would also not prefix with the ROS prefix

The only thing I'd be worried about is what does `rcl_publisher_get_topic_name()` return? The expanded topic name or the one used in `rmw_create_publisher()`?

----

I don't mean to delay this pull request, but I think we need to resolve this and update the design doc before committing to changes in the source.

mikaelarguedas: Agreed.
I suggest letting this set of PRs move forward and have first an update to the design doc to remove the partitions and describe the new behavior from from this set of PRs in the "ROS case" ([reported on the parent issue](https://github.com/ros2/ros2/issues/476#issuecomment-378408720).
We can have a separate discussion regarding what should happen in the "avoid_ros_namespace_conventions" case, update the design doc and propose code changes once design doc approved.

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -20,10 +20,8 @@
 #include "rmw/rmw.h"
 
 #include "rosidl_typesupport_introspection_cpp/identifier.hpp"
-
```
mikaelarguedas: Nitpick: Change unrelated to this PR. As we have an empty line between all include statements that come from different packages, I think this should stay.

rmw_fastrtps_cpp/src/rmw_service_server_is_available.cpp
```diff
@@ -64,32 +64,14 @@ rmw_service_server_is_available(
...
-  if (pub_partitions.size() != 1) {
-    RCUTILS_LOG_ERROR_NAMED(
-      "rmw_fastrtps_cpp",
-      "Topic %s is not a ros topic", pub_topic_name.c_str())
-    RMW_SET_ERROR_MSG((std::string(pub_topic_name) + " is a non-ros topic\n").c_str());
-    return RMW_RET_ERROR;
-  }
-  auto pub_fqdn = pub_partitions[0] + "/" + pub_topic_name;
+
+  auto pub_fqdn = pub_topic_name;
```
Karsten1987: this variable is not necessary anymore, is it?

rmw_fastrtps_cpp/src/rmw_service_server_is_available.cpp
```diff
@@ -64,32 +64,14 @@ rmw_service_server_is_available(
...
   pub_fqdn = _demangle_if_ros_topic(pub_fqdn);
```
Karsten1987: you can pass directly the outcome of getTopicName() in it.

rmw_fastrtps_cpp/src/rmw_service_server_is_available.cpp
```diff
@@ -64,32 +64,14 @@ rmw_service_server_is_available(
...
-  if (sub_partitions.size() != 1) {
-    RCUTILS_LOG_ERROR_NAMED(
-      "rmw_fastrtps_cpp",
-      "Topic %s is not a ros topic", pub_topic_name.c_str())
-    RMW_SET_ERROR_MSG((std::string(sub_topic_name) + " is a non-ros topic\n").c_str());
-    return RMW_RET_ERROR;
-  }
-  auto sub_fqdn = sub_partitions[0] + "/" + sub_topic_name;
+
+  auto sub_fqdn = sub_topic_name;
```
Karsten1987: same as above

Issue #149, opened 2017-08-24
exported libraries should not expose imported targets of FastRTPS / FastCDR


dirk-thomas:
See #148 for the current workaround to make this package usable by downstream packages.

sloretz:
If I understand correctly then the workaround in #148 seems reasonable to me. Is the issue that `rmw_fastrtps_cpp_LIBRARIES` contains an IMPORTED target from an upstream library? Would it be fine if it contained an IMPORTED target from the current library, or does it need to have full paths like [recommended for xxx_LIBRARIES in find-modules](https://cmake.org/cmake/help/v3.5/manual/cmake-developer.7.html#standard-variable-names)?

I think the workaround is a symptom of defining variables instead of following [cmake's recommendation that packages export their targets](https://cmake.org/cmake/help/v3.5/manual/cmake-packages.7.html#creating-relocatable-packages). Since the target `rmw_fastrtps_cpp` has transitive dependency info but `ament_cmake` is not exporting it, it seems appropriate to fix this in `ament_export_libraries()`.

Option 1 is closer to cmake's recommendation and preserves other transitive target information like compiler options and include directories. Option 2 seems more aligned with what I assume was an intentional design choice to make `ament_cmake` only provide variables. Thoughts or other strategies? I'll make a PR for Option 1 tomorrow if I hear nothing.

## Option 1
### export target `rmw_fastrtps_cpp`, install exported target file, and append the target name to `rmw_fastrtps_cpp_LIBRARIES`

The variable `rmw_fastrtps_cpp_LIBRARIES` would contain an imported target `rmw_fastrtps_cpp::rmw_fastrtps_cpp`.

The macro `ament_export_libraries()` would call `export()` and `install(EXPORT)` for each argument that is a non-IMPORTED target. The exported target name would be stored in `_AMENT_EXPORT_LIBRARIES` and the exported file name in a new variable `_AMENT_EXPORT_TARGET_FILES`. The template `ament_cmake_export_libraries.cmake.in` would need to `include()` these exported target files.


## Option 2
### Iterate through `INTERFACE_LINK_LIBRARIES` of targets starting at `rmw_fastrtps_cpp`, and add the library location to `rmw_fastrtps_cpp_LIBRARIES`

The variable `rmw_fastrtps_cpp_LIBRARIES` would contain paths to libraries.

[...]

PR #198, opened 2018-04-27
Export rmw_fastrtps_cpp target

sloretz:
This exports the target `rmw_fastrtps_cpp::rmw_fastrtps_cpp` and appends it to the CMake variable `${rmw_fastrtps_cpp_LIBRARIES}`. Exporting the target gives downstream users transitive dependencies automatically. Appending it to `..._LIBRARIES` causes the target to be linked against downstream targets who use `ament_target_dependencies(some_downstream_target "rmw_fastrtps_cpp")`.

This still requires `CONFIG_EXTRAS_POST`. I have not come up with a way to eliminate it. I looked at extending `ament_export_libraries(...)` to automatically append exported targets to `$..._LIBRARIES`, or get `ament_target_dependencies(...)` to automatically link against exported targets, but didn't come up with anything I feel is reasonable. I confirmed making `ament_export_libraries(...)` recursively export things in `INTERFACE_LINK_LIBRARIES` works for everything in `ros2.repos`, but I doubt my implementation would work for the general case because I explicitly ignored things beginning with a `-` (ex: `-lpthread` ) or a `$` (targets using generator expressions). 

connects to ros2/rmw_fastrtps#149

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -99,6 +99,7 @@ PRIVATE "RMW_FASTRTPS_CPP_BUILDING_LIBRARY")
 
 # specific order: dependents before dependencies
 ament_export_include_directories(include)
+ament_export_interfaces(export_rmw_fastrtps_cpp)
 ament_export_libraries(rmw_fastrtps_cpp)
 
 ament_export_dependencies(rosidl_typesupport_introspection_cpp)
@@ -106,6 +107,10 @@ ament_export_dependencies(rosidl_typesupport_introspection_c)
 ament_export_dependencies(rosidl_generator_c)
 ament_export_dependencies(rcutils)
 ament_export_dependencies(rmw)
+ament_export_dependencies(fastrtps_cmake_module)
+ament_export_dependencies(fastcdr)
+ament_export_dependencies(fastrtps)
+ament_export_dependencies(FastRTPS)
 
 register_rmw_implementation(
   "c:rosidl_typesupport_c:rosidl_typesupport_introspection_c"
@@ -127,6 +132,7 @@ install(
 
 install(
   TARGETS rmw_fastrtps_cpp
+  EXPORT export_rmw_fastrtps_cpp
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
--- a/rmw_fastrtps_cpp/rmw_fastrtps_cpp-extras.cmake
+++ b/rmw_fastrtps_cpp/rmw_fastrtps_cpp-extras.cmake
@@ -14,11 +14,4 @@
 
 # copied from rmw_fastrtps_cpp/rmw_fastrtps_cpp-extras.cmake
 
-find_package(fastrtps_cmake_module REQUIRED)
-find_package(fastcdr REQUIRED CONFIG)
-find_package(fastrtps REQUIRED CONFIG)
-find_package(FastRTPS REQUIRED MODULE)
-
-list(APPEND rmw_fastrtps_cpp_INCLUDE_DIRS ${FastRTPS_INCLUDE_DIR})
-# specific order: dependents before dependencies
-list(APPEND rmw_fastrtps_cpp_LIBRARIES fastrtps fastcdr)
+list(APPEND rmw_fastrtps_cpp_LIBRARIES rmw_fastrtps_cpp::rmw_fastrtps_cpp)


sloretz:
In progress while CI runs
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4311/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1307/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3542/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4386/)

rmw_fastrtps_cpp/CMakeLists.txt
```diff
@@ -99,13 +99,18 @@ PRIVATE "RMW_FASTRTPS_CPP_BUILDING_LIBRARY")
...
 
 ament_export_dependencies(rosidl_typesupport_introspection_cpp)
 ament_export_dependencies(rosidl_typesupport_introspection_c)
 ament_export_dependencies(rosidl_generator_c)
 ament_export_dependencies(rcutils)
 ament_export_dependencies(rmw)
+ament_export_dependencies(fastrtps_cmake_module)
+ament_export_dependencies(fastcdr)
+ament_export_dependencies(fastrtps)
+ament_export_dependencies(FastRTPS)
```
mjcarroll: Are both lines necessary here? Is there some sort of case-sensitivity?

sloretz:
This made the installed package non-relocatable because the exported targets file contains absolute paths. I don't see an easy fix. I'm just documenting what I found in case I'm curious about this again in the future.

Here's part of the installed exported target with this PR.

```cmake
set_target_properties(rmw_fastrtps_cpp::rmw_fastrtps_cpp PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/workspace/ros2_ws/install/rcutils/include;/workspace/ros2_ws/install/rosidl_typesupport_interface/include;/workspace/ros2_ws/install/rosidl_generator_c/include;/workspace/ros2_ws/install/rosidl_typesupport_introspection_c/include;/workspace/ros2_ws/install/rosidl_generator_cpp/include;/workspace/ros2_ws/install/rosidl_typesupport_introspection_cpp/include;/workspace/ros2_ws/install/rmw/include"
  INTERFACE_LINK_LIBRARIES "fastcdr;fastrtps;/workspace/ros2_ws/install/rosidl_typesupport_introspection_c/lib/librosidl_typesupport_introspection_c.so;/workspace/ros2_ws/install/rosidl_typesupport_introspection_cpp/lib/librosidl_typesupport_introspection_cpp.so;/workspace/ros2_ws/install/rcutils/lib/librcutils.so;/workspace/ros2_ws/install/rmw/lib/librmw.so;/workspace/ros2_ws/install/rosidl_generator_c/lib/librosidl_generator_c.so"
)
```

Exported targets [can be relocatable](https://cmake.org/cmake/help/v3.3/manual/cmake-packages.7.html#creating-relocatable-packages) if all paths given to `target_xxx()` functions are relative paths.

However, the absolute paths in this case are files in dependencies. Relative paths from dependencies hard-coded in the exported targets file might work for the ros2 packaging job because all the packages are shipped together, but it doesn't work if the user has any dependency installed with a different install prefix.

The right solution is to have any paths to the dependency's files come from `<dependency>Config.cmake`. The modern cmake way to do this is to have all dependencies export targets, and then `target_link_libraries()` to them. If `ament_cmake` used exported targets then the above snippet would look like this.

```cmake
set_target_properties(rmw_fastrtps_cpp::rmw_fastrtps_cpp PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES ""
  INTERFACE_LINK_LIBRARIES "fastcdr;fastrtps;rosidl_typesupport_introspection_c::rosidl_typesupport_introspection_c;rosidl_typesupport_introspection_cpp::rosidl_typesupport_introspection_cpp;rcutils::rcutils;rmw::rmw;rosidl_generator_c::rosidl_generator_c"
)
```

However `ament_cmake` makes relocatable packages using cmake variables that contain absolute paths. If a target uses variables from a dependency using `ament_cmake` then the target cannot be relocatable because any `INTERFACE_xxx` properties of it would contain those absolute paths. In this PR the absolute paths come from this call to `ament_target_dependencies()`

https://github.com/ros2/rmw_fastrtps/blob/a9c33de826339905b737b8ab01c4719b8a240aed/rmw_fastrtps_cpp/CMakeLists.txt#L85-L91

This calls `target_include_directories()` and `target_link_libraries()` using the content of `${<dependency>_INCLUDE_DIRS}` and `${<dependency>_LIBRARIES}` for each dependency. Since these dependencies are all using `ament_cmake` the include directories and libraries are populated by `ament_export_include_directories()` and `ament_export_libraries()`.

When `ament_export_include_directories()` is called [it stores a path as a string](https://github.com/ament/ament_cmake/blob/f74ccfa7399156d20c8d0eff8100445b7eefebd9/ament_cmake_export_include_directories/cmake/ament_export_include_directories.cmake#L40) prefixed by `${<dependency>_DIR}`. That variable contains the dependency's `CMAKE_INSTALL_PREFIX`. When the dependency is `find_package()`'d that string [is evaluated as an absolute path](https://github.com/ament/ament_cmake/blob/f74ccfa7399156d20c8d0eff8100445b7eefebd9/ament_cmake_export_include_directories/cmake/ament_cmake_export_include_directories-extras.cmake.in#L13) and stored in `${<dependency>_INCLUDE_DIR}`.

`ament_export_libraries()` is different but the result is the same. In this PR's case all the dependencies were target names. When the dependency is `find_packaged()` the library names are [resolved to absolute paths using `find_library()`](https://github.com/ament/ament_cmake/blob/f74ccfa7399156d20c8d0eff8100445b7eefebd9/ament_cmake_export_libraries/cmake/ament_cmake_export_libraries-extras.cmake.in#L37-L44).

Since `ament_cmake` does export a variable `${<dependency_DIR}` it might be possible to allow downstream packages to create relocatable exported targets if the exported target could be made to look like the following. I have no idea how to achieve this though.

```cmake
set_target_properties(rmw_fastrtps_cpp::rmw_fastrtps_cpp PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${rcutils_DIR}/../../../include${rosidl_typesupport_interface_DIR}/../../../include;${rosidl_generator_c_DIR}/../../../include;${rosidl_typesupport_introspection_c_DIR}/../../../include;${rosidl_generator_cpp_DIR}/../../../include;${rosidl_typesupport_introspection_cpp_DIR}/../../../include;${rmw_DIR}/../../../include"
  INTERFACE_LINK_LIBRARIES "fastcdr;fastrtps;${rosidl_typesupport_introspection_c_DIR}/../../../lib/librosidl_typesupport_introspection_c.so;${rosidl_typesupport_introspection_cpp_DIR}/../../../lib/librosidl_typesupport_introspection_cpp.so;${rcutils_DIR}/../../../lib/librcutils.so;${rmw_DIR}/../../../lib/librmw.so;${rosidl_generator_c_DIR}/../../../lib/librosidl_generator_c.so"
)
```

PR #201, opened 2018-05-15
Fix leak if client reponse is never taken

sloretz:
This fixes a leak exposed by the `rclcpp` test `test_time_source`. A buffer is created on the heap, but it is only freed if someone takes the response. This PR uses an `std::unique_ptr` instead of a raw pointer to let RAII free the buffer if an error occurs or the listener is destroyed without taking a response.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4413/)
    * False positive cmake warning ros2/build_cop#79
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1361/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3606/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4469/)

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
@@ -63,7 +63,7 @@ public:
 
     CustomClientResponse response;
     // Todo(sloretz) eliminate heap allocation pending eprosima/Fast-CDR#19
-    response.buffer_.reset(new eprosima::fastcdr::FastBuffer);
+    response.buffer_.reset(new eprosima::fastcdr::FastBuffer());
     eprosima::fastrtps::SampleInfo_t sinfo;
 
     if (sub->takeNextData(response.buffer_.get(), &sinfo)) {
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
@@ -46,9 +46,6 @@ typedef struct CustomClientResponse
 {
   eprosima::fastrtps::rtps::SampleIdentity sample_identity_;
   std::unique_ptr<eprosima::fastcdr::FastBuffer> buffer_;
-
-  CustomClientResponse()
-  : buffer_(nullptr) {}
 } CustomClientResponse;
 
 class ClientListener : public eprosima::fastrtps::SubscriberListener
@@ -65,7 +62,8 @@ public:
     assert(sub);
 
     CustomClientResponse response;
-    response.buffer_.reset(new eprosima::fastcdr::FastBuffer());
+    // Todo(sloretz) eliminate heap allocation pending eprosima/Fast-CDR#19
+    response.buffer_.reset(new eprosima::fastcdr::FastBuffer);
     eprosima::fastrtps::SampleInfo_t sinfo;
 
     if (sub->takeNextData(response.buffer_.get(), &sinfo)) {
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
```diff
@@ -43,7 +45,7 @@ typedef struct CustomClientInfo
 typedef struct CustomClientResponse
 {
   eprosima::fastrtps::rtps::SampleIdentity sample_identity_;
-  eprosima::fastcdr::FastBuffer * buffer_;
+  std::unique_ptr<eprosima::fastcdr::FastBuffer> buffer_;
```
Karsten1987: why does this have to be a heap allocation in the first place?

sloretz:
Thanks for the review @wjwwood 

The PR is complete and passes tests on my machine. I keep the in-progress label until I see green CI.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
```diff
@@ -100,19 +102,19 @@ class ClientListener : public eprosima::fastrtps::SubscriberListener
     if (conditionMutex_ != nullptr) {
       std::unique_lock<std::mutex> clock(*conditionMutex_);
       if (!list.empty()) {
-        response = list.front();
+        response = std::move(list.front());
```
Karsten1987: this move is necessary because of the unique_ptr inside the response, right? This is potentially dangerous because the list is afterwards invalid.
If you wouldn't call `pop_front()` immediately afterwards (which you do!), and try to access the first element of the list again, this will segfault.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
sloretz: eProsima/Fast-CDR#19

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
```diff
@@ -65,7 +62,8 @@ class ClientListener : public eprosima::fastrtps::SubscriberListener
     assert(sub);
 
     CustomClientResponse response;
-    response.buffer_.reset(new eprosima::fastcdr::FastBuffer());
+    // Todo(sloretz) eliminate heap allocation pending eprosima/Fast-CDR#19
+    response.buffer_.reset(new eprosima::fastcdr::FastBuffer);
```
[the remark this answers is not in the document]
sloretz: They're not required. It's the same as

```c++
std::string string1;
std::string string2();
```

I'll add them for consistency.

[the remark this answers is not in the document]
sloretz: 57192bd

sloretz:
Checking OSX CI first to make sure warning is fixed

[Build Status](https://ci.ros2.org/job/ci_osx/3607/)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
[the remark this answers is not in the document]
Karsten1987: wow. didn't know that :)
Sorry for the noise then!

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
sloretz: 210346b

I don't see an API in the list class that combines getting the first element and removing it. 
210346b  uses a lambda so at least all the risky stuff is in one spot. It should be fine as long as the move constructor doesn't throw an exception.

Karsten1987:
maybe I miss something here, but what are the changes compared to last time?

sloretz:
I mistakenly dismissed @Karsten1987's review. CI looks ok, merging.

PR #203, opened 2018-05-30
Static typesupport

MiguelCompany:
## Summary

At eProsima we are working on static typesupport for rmw_fastrtps. We now have the basic functionality ready for C++, but we still have the following todos:

- [X] create package rosidl_typesupport_fastrtps_c with support for C (i.e. python)
- [X] refactor to avoid #ifdef RMW_FASTRTPS_CPP_STATIC_TYPESUPPORT
  - [X] create package rmw_fastrtps_shared_cpp with common code
  - [X] create package rmw_fastrtps_dynamic_cpp with dynamic specific (introspection) code
  - [X] leave static specific code only on package rmw_fastrtps_cpp
- [X] pass linters

## Performance test

We have run some performance tests (which code is also in the repo). The test scenario involves two applications:

- Listener application with subscriber for "test" and publisher for "test_echo":
  - It will publish on "test_echo" anything received on "test"

- Talker application with publisher for "test" and subscriber for "test_echo":
  - It will publish on "test" anything received on "test_echo"
  - It will increase a counter when receiving on "test_echo"
  - It will show the counter value every 10 seconds
  - It will publish a single message on "test" upon startup

For each data type we run the test as follows:
- Start the listener
- Start the talker and leave it running for 2 minutes
- Take note of the number of times the message has gone from the talker to the listener and back

### Performance test results

Depending on the data type, we see an improvement from 0.15% to 84%. The improvement increases with the size of the message as well as with its complexity (number of fields / nested submessages). 

| Data type                        | master  | static  | improvement |
|----------------------------------|---------|---------|-------------|
| [String](https://github.com/ros2/common_interfaces/blob/master/std_msgs/msg/String.msg) <br/>("Hello World")           | 1145937 | 1148323 | 0,21%       |
| [Array16k](https://github.com/eProsima/rmw_fastrtps/blob/static_typesupport_old_/performance_test_msgs/msg/Array16k.msg) <br/>(16KB static array)                       | 891891  | 913705  | 2,45%       |
| [Array4m](https://github.com/eProsima/rmw_fastrtps/blob/static_typesupport_old_/performance_test_msgs/msg/Array4m.msg) <br/>(4MB static array)                        | 1198    | 1976    | 64,94%      |
[...]

PR #205, opened 2018-06-08
Nullify pointer to ClientListener in CustomClientInfo in destructor

sloretz:
This mitigates (does not fix) a race condition when one thread calls `rmw_wait()` and another calls `rmw_destroy_client()`. When `rmw_wait()` wakes up it calls `CustomClientInfo->listener_`, but both the instance of `CustomClientInfo` and the object pointed to by `listener_` have been destroyed. This nulls the field `listener_` and has `rmw_wait()` check this field before using it to reduce the likelyhood of a crash.

I observed this race condition while debugging a CI test failure in ros2/rclcpp#488. The symptom seen is a SIGABRT on OSX when trying to lock `ClientListener.internalMutex_`.

I'm not sure what a real fix would look like. `rmw_wait()` needs a way to tell the client it is holding onto has been destroyed. The other entities appear to be using the same pattern, so I suspect they suffer from similar race conditions too. (ticket to follow)

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/custom_client_info.hpp
@@ -55,6 +55,12 @@ public:
   : info_(info), list_has_data_(false),
     conditionMutex_(nullptr), conditionVariable_(nullptr) {}
 
+  ~ClientListener()
+  {
+    // TODO(sloretz) must acquire lock before editing info_
+    // Remove reference to self to avoid crash in rmw_wait()
+    info_->listener_ = nullptr;
+  }
 
   void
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I would suggest to wait with merging this until #203 has been merged or it has been decided to delay it for now.

mikaelarguedas:
My understanding is that #203 will not be integrated before feature freeze so we can move these race condition mitigation forward for the release.

@sloretz is this PR ready for review ?

dirk-thomas:
If #203 doesn't necessarily change the default behavior and we don't have concerns about regressions it is imo still on the list to possibly be merged.

sloretz:
Closing this since there have been enough changes to cause conflicts and it only mitigated an issue.

PR #208, opened 2018-06-16
update usage of rcutils_join_path()

wjwwood:
Connects to ros2/rcutils#102


wjwwood:
Ok I think 2680a9d should address the bug.

PR #207, opened 2018-06-14
Temporary buffer remove

MiguelCompany:
This PR includes the necessary changes to remove the temporary buffer used on (de)serialization of messages, but keeping only the introspection typesupport.

I am still performing tests locally, but I have started the PR so others can perform their own tests, and check how it compares to #203

PR #211, opened 2018-06-20
Avoid allocations

MiguelCompany:
This is an improvement over #207 that removes allocations on more cases, including rmw_serialize.

I am rebasing #203 on top of this one, so if this one gets merged the other will be easier to review and merge.

It is passing the tests locally, so I hope it can make it into bouncy

### DIFF
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -420,8 +420,11 @@ size_t next_field_align(
     std::vector<T> & data = *reinterpret_cast<std::vector<T> *>(field);
     current_alignment += eprosima::fastcdr::Cdr::alignment(current_alignment, padding);
     current_alignment += padding;
-    current_alignment += eprosima::fastcdr::Cdr::alignment(current_alignment, item_size);
-    current_alignment += item_size * data.size();
+    auto num_elems = data.size();
+    if (num_elems > 0) {
+      current_alignment += eprosima::fastcdr::Cdr::alignment(current_alignment, item_size);
+      current_alignment += item_size * data.size();
+    }
   }
   return current_alignment;
 }
@@ -437,7 +440,7 @@ size_t next_field_align<std::string>(
   if (!member->is_array_) {
     current_alignment += eprosima::fastcdr::Cdr::alignment(current_alignment, padding);
     current_alignment += padding;
-    auto str = *static_cast<std::string *>(field);
+    auto & str = *static_cast<std::string *>(field);
     current_alignment += str.size() + 1;
   } else if (member->array_size_ && !member->is_upper_bound_) {
     auto str_arr = static_cast<std::string *>(field);
@@ -935,7 +938,7 @@ size_t TypeSupport<MembersType>::getEstimatedSerializedSize(
   size_t ret_val = 4;
 
   if (members_->member_count_ != 0) {
-    ret_val += TypeSupport::getEstimatedSerializedSize(members_, ros_message, ret_val);
+    ret_val += TypeSupport::getEstimatedSerializedSize(members_, ros_message, 0);
   } else {
     ret_val += 1;
   }
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
@@ -1032,12 +1032,24 @@ bool TypeSupport<MembersType>::deserialize(
   assert(data);
   assert(payload);
 
-  auto buffer = static_cast<eprosima::fastcdr::FastBuffer *>(data);
-  if (!buffer->reserve(payload->length)) {
-    return false;
+  auto ser_data = static_cast<SerializedData *>(data);
+  if (ser_data->is_cdr_buffer) {
+    auto buffer = static_cast<eprosima::fastcdr::FastBuffer *>(ser_data->data);
+    if (!buffer->reserve(payload->length)) {
+      return false;
+    }
+    memcpy(buffer->getBuffer(), payload->data, payload->length);
+    return true;
   }
-  memcpy(buffer->getBuffer(), payload->data, payload->length);
-  return true;
+
+  eprosima::fastcdr::FastBuffer fastbuffer(
+    reinterpret_cast<char *>(payload->data),
+    payload->length);
+  eprosima::fastcdr::Cdr deser(
+    fastbuffer,
+    eprosima::fastcdr::Cdr::DEFAULT_ENDIAN,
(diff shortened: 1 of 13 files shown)


rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -994,13 +997,28 @@ bool TypeSupport<MembersType>::serialize(
...
   assert(payload);
 
-  auto ser = static_cast<eprosima::fastcdr::Cdr *>(data);
-  if (payload->max_size >= ser->getSerializedDataLength()) {
-    payload->length = static_cast<uint32_t>(ser->getSerializedDataLength());
-    payload->encapsulation = ser->endianness() ==
-      eprosima::fastcdr::Cdr::BIG_ENDIANNESS ? CDR_BE : CDR_LE;
-    memcpy(payload->data, ser->getBufferPointer(), ser->getSerializedDataLength());
-    return true;
+  auto ser_data = static_cast<SerializedData *>(data);
```
Karsten1987: The function could be strongly typed then, right? So this cast could go.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -994,13 +997,28 @@ bool TypeSupport<MembersType>::serialize(
...
+  if (ser_data->is_cdr_buffer) {
```
Karsten1987: I am a little bit confused. Why would you call `serialize` on already binary data?

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
```diff
@@ -1014,21 +1032,41 @@ bool TypeSupport<MembersType>::deserialize(
   assert(data);
   assert(payload);
 
-  auto buffer = static_cast<eprosima::fastcdr::FastBuffer *>(data);
-  if (!buffer->reserve(payload->length)) {
-    return false;
+  auto ser_data = static_cast<SerializedData *>(data);
```
Karsten1987: i think this function also could be strongly typed.

rmw_fastrtps_cpp/src/rmw_publish.cpp
```diff
@@ -45,17 +46,10 @@ rmw_publish(const rmw_publisher_t * publisher, const void * ros_message)
...
-    info->typesupport_identifier_))
-  {
-    RMW_SET_ERROR_MSG("cannot serialize data");
-    return RMW_RET_ERROR;
-  }
-  if (!info->publisher_->write(&ser)) {
+  rmw_fastrtps_cpp::SerializedData data;
+  data.is_cdr_buffer = false;
+  data.data = const_cast<void *>(ros_message);
+  if (!info->publisher_->write(&data)) {
```
Karsten1987: shouldn't this be `data.data` ?

dirk-thomas:
Looking at the first commit this doesn't seem to only be an improvement but it seems to also fix something. Can you please clarify what the problem without that fix is with the current code.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4771/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1636/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3957/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4828/)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
MiguelCompany: Not possible, as it is the implementation of Fast-RTPS TopicDataType interface, declared in [this line](https://github.com/eProsima/Fast-RTPS/blob/fb91928914f7f8e47d0ef10bae26b207524b75ad/include/fastrtps/TopicDataType.h#L52)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
MiguelCompany: Not possible, as it is the implementation of Fast-RTPS TopicDataType interface, declared in [this line](https://github.com/eProsima/Fast-RTPS/blob/fb91928914f7f8e47d0ef10bae26b207524b75ad/include/fastrtps/TopicDataType.h#L60)

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
MiguelCompany: This is the case when the call comes from rmw_publish_serialized_message.

Karsten1987: so right now, there is no call to serialize. The raw data will simply be wrapped in `Cdr` and directly send.
I feel like, this decision of whether it's actually necessary to serialize or not should be taken outside of the serialize function.

rmw_fastrtps_cpp/src/rmw_publish.cpp
MiguelCompany: As stated in [this comment](https://github.com/eProsima/rmw_fastrtps/blob/d4f1dd8453154f02f4fa5ddca798dc9b8bd8bf4a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.hpp#L45), it is not. The write method is provided with a pointer to a SerializedData struct, that will in turn be passed on to the serialize method on TypeSupport_impl.

dirk-thomas:
@MiguelCompany since I noticed you commenting on other questions in this ticket can you please respond to the question in [this comment](https://github.com/ros2/rmw_fastrtps/pull/211#issuecomment-398810406) asap?

MiguelCompany:
> Can you please clarify what the problem without that fix is with the current code.

There are some edge cases where the estimated size will be short by up to 4 bytes. That would provoke a reallocation of the internal buffer when serializing the last field on some messages.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp
MiguelCompany: When you say 'right now', do you mean on the code currently on master, or on this PR.

Whenever data is sent, the serialize function of TypeSupport class is called. In the code currently on master it just copies an externally serialized buffer into the SerializedPayload (that points) to the internal history record buffer.

The code on this PR does that on some cases (namely rmw_publish_serialized) and performs serialization directly on others (rmw_publish). On the latter case we save the allocation of a temporary buffer.

PR #215, opened 2018-07-16
add rmw_get_encoding_identifier

Karsten1987:
connects to ros2/rmw#143


rmw_fastrtps_cpp/src/rmw_get_encoding_identifier.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright 2016 Proyectos y Sistemas de Mantenimiento SL (eProsima).
```
wjwwood: Is this intentional?

Issue #216, opened 2018-07-19
rmw_count_subscribers / rmw_count_publishers performance improvments


guillaumeautran:
The current implementation of  `rmw_count_subscribers` and `rmw_count_publishers` functions as a linear runtime complexity , O(N), w.r.t. the number of ROS2 topics published. This is not a big problem when the calling code has to count the number of ROS2 subscribers/publishers for a single topic, but it becomes a lot more of a performance problem when the calling code needs to count the number of ROS2 subscribers/publishers for each topic seen on ROS2. In that case, the runtime complexity becomes O(N^2) very easily.

It should be possible to re-write these methods to bring done the runtime complexity to either a constant or at least O(logN).

Here is a screenshot of the runtime analysis for `rmw_count_publishers` (counting the subscribers will have the same issues).
![image](https://user-images.githubusercontent.com/13353403/42960130-f180d81e-8b58-11e8-9255-270cb5eea269.png)

dirk-thomas:
Addressed by #217.

PR #217, opened 2018-07-19
Improve runtime performance of the `rmw_count_XXX` functions.

guillaumeautran:
Change the runtime performance of `rmw_count_subscribers` and `rmw_count_publishers` to be O(logN) instead of linear.

connects to ros2/rmw_fastrtps#216

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_count.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_count.cpp
@@ -23,6 +23,7 @@
 #include "rmw/types.h"
 
 #include "demangle.hpp"
+#include "namespace_prefix.hpp"
 #include "rmw_fastrtps_shared_cpp/custom_participant_info.hpp"
 #include "rmw_fastrtps_shared_cpp/rmw_common.hpp"
 #include "reader_info.hpp"
@@ -49,15 +50,28 @@ __rmw_count_publishers(
     return RMW_RET_ERROR;
   }
 
-  auto impl = static_cast<CustomParticipantInfo *>(node->data);
-
-  WriterInfo * slave_target = impl->secondaryPubListener;
-  slave_target->mapmutex.lock();
   *count = 0;
-  for (const auto & it : slave_target->topicNtypes) {
-    const auto topic_fqdn = _demangle_if_ros_topic(it.first);
-    if (topic_fqdn == topic_name) {
-      *count += it.second.size();
+  auto ros_prefixes = _get_all_ros_prefixes();
+
+  // Build the list of all possible topic FQDN
+  std::vector<std::string> topic_fqdns;
+  topic_fqdns.push_back(topic_name);
+  if (topic_name[0] == '/') {
+    std::for_each(ros_prefixes.begin(), ros_prefixes.end(),
+      [&topic_fqdns, &topic_name](const std::string & prefix) {
+        topic_fqdns.push_back(prefix + topic_name);
+      });
+  }
+
+  auto impl = static_cast<CustomParticipantInfo *>(node->data);
+  WriterInfo * slave_target = impl->secondaryPubListener;
+
+  slave_target->mapmutex.lock();
+  // Search and sum up the publisher counts
+  for (const auto & topic_fqdn : topic_fqdns) {
+    const auto & it = slave_target->topicNtypes.find(topic_fqdn);
+    if (it != slave_target->topicNtypes.end()) {
+      *count += it->second.size();
     }
   }
   slave_target->mapmutex.unlock();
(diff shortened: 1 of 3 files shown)


guillaumeautran:
With this fix in place, the performance of the method is much better. Goes from 77% down to 19% for the same usecase (755 ROS2 topics)
image

guillaumeautran:
@dirk-thomas Just rebased the changes. Please let me know if anything needs to be changed.

rmw_fastrtps_shared_cpp/src/rmw_count.cpp
```diff
@@ -54,12 +55,24 @@ __rmw_count_publishers(
...
-  for (const auto & it : slave_target->topicNtypes) {
-    const auto topic_fqdn = _demangle_if_ros_topic(it.first);
-    if (topic_fqdn == topic_name) {
-      *count += it.second.size();
+
+  // Build the list of all possible topic FQDN
+  std::vector<std::string> topic_fqdns;
+  topic_fqdns.push_back(topic_name);
+  if (topic_name[0] == '/') {
+    std::for_each(_ros_prefixes.begin(), _ros_prefixes.end(),
```
dirk-thomas: Accessing `_ros_prefixes` while it starts with an underscore seems to be not a good idea since it is considered private. Maybe either change the declaration or expose it through a getter.

rmw_fastrtps_shared_cpp/src/rmw_count.cpp
```diff
@@ -54,12 +55,24 @@ __rmw_count_publishers(
...
+      [&topic_fqdns, &topic_name](const std::string & prefix) {
+        topic_fqdns.push_back(prefix + topic_name);
+      });
+  }
```
dirk-thomas: This block can be moved above line 53 and therefore outside of the mutex to minimize the time spent while locked,

Same below.

rmw_fastrtps_shared_cpp/src/rmw_count.cpp
```diff
@@ -54,12 +55,24 @@ __rmw_count_publishers(
   WriterInfo * slave_target = impl->secondaryPubListener;
   slave_target->mapmutex.lock();
   *count = 0;
-  for (const auto & it : slave_target->topicNtypes) {
-    const auto topic_fqdn = _demangle_if_ros_topic(it.first);
-    if (topic_fqdn == topic_name) {
-      *count += it.second.size();
+
+  // Build the list of all possible topic FQDN
+  std::vector<std::string> topic_fqdns;
```
dirk-thomas: Maybe a `set` would be the better data structure to later check if a specific item is in it. I would assume it performs better.

guillaumeautran: A `std::set` and `std::map` have the same runtime complexity `O(logN)` (for finding items). http://www.cplusplus.com/reference/set/set/find/ and http://www.cplusplus.com/reference/map/map/find/

There is no real differences but if you would rather have me use a `std::set`, I can change that.

rmw_fastrtps_shared_cpp/src/rmw_count.cpp
guillaumeautran: Sure, I can make a getter. That being said, the original code read: `_demangle_if_ros_topic(it.first)`. Notice the leading `_` character?

rmw_fastrtps_shared_cpp/src/rmw_count.cpp
dirk-thomas: The current patch uses `std::vector` which afaik has `O(N)`. So `O(logN)` for a `std::set` would be an improvement.

guillaumeautran: Sorry, you are correct. 
Will fix this as well.

guillaumeautran: Ok, I'm getting myself confused. 
For the `topic_fqdns` variable, using a vector is more efficient as we push_back (`O(1)`) all possible topic names then iterate over that and find through the list of existing ROS2 topics (`O(logN)`). So, changing to using a set for this datastructure simply does not make sense.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4977/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1792/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4990/)

dirk-thomas:
Thank you for the improvement.

PR #224, opened 2018-08-28
Include node namespaces in get_node_names

mjcarroll:
Connects to ros2/rmw#142

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_node.cpp
@@ -257,7 +257,7 @@ __rmw_create_node(
   participantAttrs.rtps.userData.resize(length);
   int written = snprintf(reinterpret_cast<char *>(participantAttrs.rtps.userData.data()),
       length, "name=%s;namespace=%s;", name, namespace_);
-  if (written < 0 || written > length - 1) {
+  if (written < 0 || written > static_cast<int>(length) - 1) {
     RMW_SET_ERROR_MSG("failed to populate user_data buffer");
     return nullptr;
   }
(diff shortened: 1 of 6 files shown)


rmw_fastrtps_shared_cpp/src/rmw_node.cpp
```diff
@@ -251,15 +251,12 @@ __rmw_create_node(
...
-  memcpy(participantAttrs.rtps.userData.data(), prefix, sizeof(prefix) - 1);
-  for (size_t i = 0; i < name_length; ++i) {
-    participantAttrs.rtps.userData[sizeof(prefix) - 1 + i] = name[i];
-  }
-  participantAttrs.rtps.userData[sizeof(prefix) - 1 + name_length] = ';';
+
+  size_t length = strlen(name) + strlen("name=;") +
+    strlen(namespace_) + strlen("namespace=;") + 1;
+  participantAttrs.rtps.userData.resize(length);
+  snprintf(reinterpret_cast<char *>(participantAttrs.rtps.userData.data()),
```
wjwwood: Return code should be checked here too I think.

rmw_fastrtps_shared_cpp/src/rmw_node_names.cpp
```diff
@@ -64,15 +69,26 @@ __rmw_get_node_names(
     RMW_SET_ERROR_MSG(rcutils_get_error_string_safe())
     return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
   }
+
+  rcutils_ret =
+    rcutils_string_array_init(node_namespaces, participant_names.size() + 1, &allocator);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string_safe())
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
```
wjwwood: Need to cleanup `node_names` (at least) before returning here.

PR #234, opened 2018-11-12
Methods to retrieve matched counts on pub/sub

mjcarroll:
Connects to ros2/rmw#155

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_publisher.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_publisher.cpp
@@ -183,11 +183,11 @@ fail:
 }
 
 rmw_ret_t
-rmw_count_matched_subscriptions(
+rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count)
 {
-  return rmw_fastrtps_shared_cpp::__rmw_count_matched_subscriptions(
+  return rmw_fastrtps_shared_cpp::__rmw_publisher_count_matched_subscriptions(
     publisher, subscription_count);
 }
 
--- a/rmw_fastrtps_cpp/src/rmw_publisher.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_publisher.cpp
@@ -128,7 +128,8 @@ rmw_create_publisher(
     goto fail;
   }
 
-  info->publisher_ = Domain::createPublisher(participant, publisherParam, nullptr);
+  info->listener_ = new PubListener(info);
+  info->publisher_ = Domain::createPublisher(participant, publisherParam, info->listener_);
 
   if (!info->publisher_) {
     RMW_SET_ERROR_MSG("create_publisher() could not create publisher");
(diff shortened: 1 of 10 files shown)


rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -128,8 +128,13 @@ rmw_create_publisher(
     goto fail;
   }
 
-  info->publisher_ = Domain::createPublisher(participant, publisherParam, nullptr);
+  info->listener_ = new PubListener(info);
+  if (!info->listener_) {
```
wjwwood: Unless you use the `nothrow` version of `new`, `new` will throw an exception (`std::bad_alloc`), so this should never be reached when it's `nullptr`.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -128,7 +128,8 @@ rmw_create_publisher(
     goto fail;
   }
 
-  info->publisher_ = Domain::createPublisher(participant, publisherParam, nullptr);
+  info->listener_ = new PubListener(info);
```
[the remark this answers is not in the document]
wjwwood: The use of `new` here is actually dangerous, because it will throw if malloc fails, so it will try to throw an exception across the rmw C API, which is undefined behavior I believe.

rmw_fastrtps_dynamic_cpp/src/rmw_publisher.cpp
```diff
@@ -128,8 +128,13 @@ rmw_create_publisher(
     goto fail;
   }
 
-  info->publisher_ = Domain::createPublisher(participant, publisherParam, nullptr);
+  info->listener_ = new PubListener(info);
```
wjwwood: Same issue as the rmw static case.

rmw_fastrtps_dynamic_cpp/src/rmw_serialize.cpp
```diff
@@ -48,7 +48,8 @@ rmw_serialize(
     }
   }
 
-  eprosima::fastcdr::FastBuffer buffer(reinterpret_cast<char *>(serialized_message->buffer), data_length);
+  eprosima::fastcdr::FastBuffer buffer(reinterpret_cast<char *>(serialized_message->buffer),
+    data_length);
```
wjwwood: Was this linter error preexisting?

nitpick: we usually wrap all or none (no leaving the first parameter on the same line as the call)

mjcarroll: It was existing when I got to it, will rewrap accordingly.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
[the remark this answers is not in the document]
wjwwood: See this for how new and nothrow work together: https://en.cppreference.com/w/cpp/memory/new/nothrow

PR #236, opened 2018-11-16
Add topic cache object for managing topic relations

ross-desmond:
* Add topic cache object to maintain participant relationship
* Separate responsibility of maintaining cached data from listener
* Use lock_guard over manual mutex lock
* Use c++14 functions to reduce code
* Add debug statements
* Add lookup of topics for input node
* Remove unnecessary logging
* Add namespace to graph functions

### DIFF
--- a/rmw_fastrtps_shared_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_shared_cpp/CMakeLists.txt
@@ -52,6 +52,7 @@ add_library(rmw_fastrtps_shared_cpp
   src/rmw_guard_condition.cpp
   src/rmw_logging.cpp
   src/rmw_node.cpp
+  src/rmw_node_info_and_types.cpp
   src/rmw_node_names.cpp
   src/rmw_publish.cpp
   src/rmw_publisher.cpp
@@ -68,8 +69,10 @@ add_library(rmw_fastrtps_shared_cpp
   src/rmw_wait_set.cpp
   src/TypeSupport_impl.cpp
 )
+
 target_link_libraries(rmw_fastrtps_shared_cpp
-  fastcdr fastrtps)
+  fastcdr fastrtps
+)
 
 # specific order: dependents before dependencies
 ament_target_dependencies(rmw_fastrtps_shared_cpp
--- a/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
@@ -1,0 +1,487 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <functional>
+#include <map>
+#include <set>
+#include <string>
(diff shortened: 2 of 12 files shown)


ross-desmond:
Connected to [rcl](https://github.com/ros2/rcl/pull/333)

thomas-moulard:
@ross-desmond PTAL - fixed copyright and cpplint issues
https://github.com/thomas-moulard/rmw_fastrtps/tree/node_graph_impl

Verified that all tests are succeeding locally.

rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
```diff
@@ -0,0 +1,490 @@
...
+          RCUTILS_LOG_ERROR_NAMED(
+            kLoggerTag,
+            "error during report of error: %s", rmw_get_error_string());
+        }
+      };
+    // Setup demangling functions based on no_demangle option
+    auto demangle_topic = _demangle_if_ros_topic;
+    auto demangle_type = _demangle_if_ros_type;
+    if (no_demangle) {
+      auto noop = [](const std::__cxx11::string & in) {
```
clalancette: It looks like macOS is cranky about this use of `std::__cxx11`: https://ci.ros2.org/job/ci_osx/4759/consoleFull, search for "__cxx11".  Can we just use `std::string` here?

clalancette: I went ahead and did it in 9ea473a .  If this isn't appropriate for some reason, we'll have to figure out another cross-platform way.

ross-desmond: Awesome, I went to change this and found it already there :)

rmw_fastrtps_shared_cpp/src/types/topic_cache.hpp
```diff
@@ -0,0 +1,236 @@
...
+   * Map of participant GUIDS to a set of topic-type.
+   */
+  ParticipantTopicMap participant_to_topics_;
+
+  /**
+   * Helper function to initialize a topic vector.
+   *
+   * @param topic_name
+   */
+  void initializeTopic(const std::string & topic_name, TopicToTypes & topic_to_types)
```
clalancette: This is definitely a subjective style thing, but I find it weird to have a private method where the second argument is always one of the member variables.  It would make more sense to me to have the arguments to this be `topic_name` and `guid`, and have it initialize both (by accessing the member variables), but there may be some reason that this doesn't work.

rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
```diff
@@ -0,0 +1,490 @@
...
+          std::string service_type = _demangle_service_type_only(itt);
+          if (service_type.length()) {
+            services[service_name].insert(service_type);
+          }
+        }
+      }
+    }
+  }
+
+  if (services.size()) {
```
clalancette: This is a subjective thing, but we could do an early return and lose one level of indentation.  Something like:

```
if (services.empty()) {
  return RMW_RET_OK;
}
```

rmw_fastrtps_shared_cpp/CMakeLists.txt
```diff
@@ -62,8 +63,8 @@ add_library(rmw_fastrtps_shared_cpp
   src/rmw_trigger_guard_condition.cpp
   src/rmw_wait.cpp
   src/rmw_wait_set.cpp
-  src/TypeSupport_impl.cpp
-)
+  src/TypeSupport_impl.cpp)
```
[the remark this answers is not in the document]
ross-desmond: Will remove.

rmw_fastrtps_shared_cpp/src/reader_info.hpp
```diff
@@ -17,9 +17,7 @@
 
 #include <cassert>
 #include <map>
-#include <mutex>
```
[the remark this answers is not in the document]
ross-desmond: Adding back.

PR #237, opened 2018-11-19
refactor to support init options and context

wjwwood:
Connects to ros2/rmw#154

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -92,7 +92,8 @@ rmw_shutdown(rmw_context_t * context)
     context->implementation_identifier,
     eprosima_fastrtps_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
-  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
+  // context impl is explicitly supposed to be nullptr for now, see rmw_init's code
+  // RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
   *context = rmw_get_zero_initialized_context();
   return RMW_RET_OK;
 }
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -82,4 +82,18 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
   context->impl = nullptr;
   return RMW_RET_OK;
 }
+
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
+{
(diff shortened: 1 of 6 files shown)


rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -12,13 +12,88 @@
...
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    eprosima_fastrtps_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
```
sloretz: Same comment about `context->impl` should be `null`

rmw_fastrtps_dynamic_cpp/src/rmw_init.cpp
```diff
@@ -12,13 +12,88 @@
...
+rmw_init_options_init(rmw_init_options_t * init_options, rcutils_allocator_t allocator)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(init_options, RMW_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RMW_RET_INVALID_ARGUMENT);
+  if (NULL != init_options->implementation_identifier) {
+    RMW_SET_ERROR_MSG("expected zero-initialized init_options");
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+  init_options->instance_id = 0;
+  init_options->implementation_identifier = eprosima_fastrtps_identifier;
```
sloretz: Should there be an `eprosima_fastrtps_dynamic_identifier` like connext has?

rmw_fastrtps_dynamic_cpp/src/rmw_init.cpp
```diff
@@ -12,13 +12,88 @@
...
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    eprosima_fastrtps_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
```
sloretz: same comment about `context->impl` being set to `null` in `rmw_init()`

rmw_fastrtps_dynamic_cpp/src/rmw_init.cpp
wjwwood: Maybe, but I think that would be follow up to the fastrtps dynamic pr's and not in scope for this pr.

PR #238, opened 2018-11-20
Improve service_is_available logic to protect that client is waiting …

DongheeYe:
…forever

When checking that service is ready, rmw_fastrtps uses CustomClientInfo's WriterInfo, ReaderInfo topic count.
It is increased when received message about Reader/Writer information from the service.
callstack is as below.
  eprosima::fastrtps::rtps::StatefulReader::processDataMsg
    eprosima::fastrtps::rtps::StatefulReader::change_received
      eprosima::fastrtps::rtps::EDPSimplePUBListener::onNewCacheChangeAdded
         WriterInfo::onNewCacheChangeAdded

But, although receiving the message from service, rtps is not ready to send a message yet.
If client tries to send request, there is no matched remote reader yet, the request is not properly sent to service. In that case, if client tries to wait until receiving the response, client will be blocked because any service does not send response.
After matching local writer proxy and remote reader, rtps is completely ready to send.

Especially, if SECURITY is ON, there is bigger time gap.
There is more possibility to fail to send request to service in SECURITY ON.

If you repeatedly test examples_rclcpp_minimal_client with examples_rclcpp_minimal_service,
you can reproduce the issue.

So, I add Listener to detect matching. callstack will be as below.
It will check that the service is ready using matched reader/writer count.
  eprosima::fastrtps::rtps::security::SecurityManager::discovered_reader
    eprosima::fastrtps::rtps::EDP::pairing_remote_reader_with_local_writer_after_security
     eprosima::fastrtps::PublisherImpl::PublisherWriterListener::onWriterMatched
       ClientPubListener::onPublicationMatched

Connects to ros2/rmw_fastrtps#239

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -92,6 +92,8 @@ rmw_create_client(
   info = new CustomClientInfo();
   info->participant_ = participant;
   info->typesupport_identifier_ = type_support->typesupport_identifier;
+  info->request_publisher_matched_count_ = 0;
+  info->response_subscriber_matched_count_ = 0;
 
   const service_type_support_callbacks_t * service_members;
   const message_type_support_callbacks_t * request_members;
@@ -178,8 +180,9 @@ rmw_create_client(
     RMW_SET_ERROR_MSG("failed to get datawriter qos");
     goto fail;
   }
+  info->pub_listener_ = new ClientPubListener(info);
   info->request_publisher_ =
-    Domain::createPublisher(participant, publisherParam, nullptr);
+    Domain::createPublisher(participant, publisherParam, info->pub_listener_);
   if (!info->request_publisher_) {
     RMW_SET_ERROR_MSG("create_publisher() could not create publisher");
     goto fail;
@@ -213,6 +216,10 @@ fail:
 
     if (info->response_subscriber_ != nullptr) {
       Domain::removeSubscriber(info->response_subscriber_);
+    }
+
+    if (info->pub_listener_ != nullptr) {
+      delete info->pub_listener_;
     }
 
     if (info->listener_ != nullptr) {
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
@@ -27,10 +27,12 @@
 #include "fastrtps/subscriber/SubscriberListener.h"
 #include "fastrtps/participant/Participant.h"
 #include "fastrtps/publisher/Publisher.h"
+#include "fastrtps/publisher/PublisherListener.h"
 
 #include "rmw_fastrtps_shared_cpp/TypeSupport.hpp"
 
 class ClientListener;
+class ClientPubListener;
 
 typedef struct CustomClientInfo
 {
@@ -42,6 +44,9 @@ typedef struct CustomClientInfo
   eprosima::fastrtps::rtps::GUID_t writer_guid_;
   eprosima::fastrtps::Participant * participant_;
   const char * typesupport_identifier_;
+  ClientPubListener * pub_listener_;
+  uint32_t response_subscriber_matched_count_;
+  uint32_t request_publisher_matched_count_;
 } CustomClientInfo;
 
 typedef struct CustomClientResponse
@@ -141,6 +146,18 @@ public:
     return list_has_data_.load();
   }
 
+  void onSubscriptionMatched(eprosima::fastrtps::Subscriber * sub,
+      eprosima::fastrtps::rtps::MatchingInfo & matchingInfo)
+  {
+    if (info_ == nullptr || sub == nullptr)
+      return;
+
+    if (matchingInfo.status == eprosima::fastrtps::rtps::MATCHED_MATCHING)
+      info_->response_subscriber_matched_count_++;
+    else
+      info_->response_subscriber_matched_count_--;
+  }
+
 private:
   CustomClientInfo * info_;
   std::mutex internalMutex_;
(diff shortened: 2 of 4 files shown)


mjcarroll:
Do you think that you can add a test in `test_rclcpp` that triggers this behavior? https://github.com/ros2/system_tests/tree/master/test_rclcpp

mjcarroll:
CI: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5663/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2272/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4681/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5552/)

mjcarroll:
Can you rebase to include the changes from: https://github.com/ros2/rmw_fastrtps/pull/240 (now in master)

DongheeYe:
I'd rebased the patch based on latest in master.

DongheeYe:
@mjcarroll 
I think it needs a stress test because it is a timing issue. It might not be occurred by single test. For example.. try to run a client repeatedly more than hundreds times after running a service.
```
 $ ./install/examples_rclcpp_minimal_service/lib/examples_rclcpp_minimal_service/service_main &
 $ for i in $(seq 1 500) ; do ./install/examples_rclcpp_minimal_client/lib/examples_rclcpp_minimal_client/client_main ; done
```

jacobperron:
With the above snippet, I am able to consistently reproduce the issue with the security flag on and by making my machine busy (building a workspace with colcon). If my CPU is not busy, then I do not see the issue.

With the proposed fix in this PR, the original issue appears to go away, but I occasionally experience the client stuck waiting for the service. This looks different than the original issue since it is repeatedly printing the following message:

```
[INFO] [minimal_client]: waiting for service to appear...
```

If I restart the server, it succeeds along with all consecutive calls to the client.
But if I restart the client instead, it almost certainly gets stuck waiting again with the same message after a few more iterations.

Example output when running loop from the provided snippet:

```
...
[INFO] [minimal_client]: result of 41 + 1 = 42
[INFO] [minimal_client]: result of 41 + 1 = 42
[INFO] [minimal_client]: waiting for service to appear...
[INFO] [minimal_client]: waiting for service to appear...
[INFO] [minimal_client]: waiting for service to appear...
[INFO] [minimal_client]: waiting for service to appear...
[INFO] [minimal_client]: waiting for service to appear...
^C[INFO] [rclcpp]: signal_handler(signal_value=2)
[ERROR] [minimal_client]: client interrupted while waiting for service to appear.
[INFO] [minimal_client]: result of 41 + 1 = 42
[INFO] [minimal_client]: result of 41 + 1 = 42
[INFO] [minimal_client]: result of 41 + 1 = 42
... # x9
[INFO] [minimal_client]: waiting for service to appear...
[INFO] [minimal_client]: waiting for service to appear...
...
```

It seems there may still be an issue with the implementation.

DongheeYe:
> With the above snippet, I am able to consistently reproduce the issue with the security flag on and by making my machine busy (building a workspace with colcon). If my CPU is not busy, then I do not see the issue.
> 
> With the proposed fix in this PR, the original issue appears to go away, but I occasionally experience the client stuck waiting for the service. This looks different than the original issue since it is repeatedly printing the following message:
> 
> ```
> [INFO] [minimal_client]: waiting for service to appear...
> ```
> If I restart the server, it succeeds along with all consecutive calls to the client.
> But if I restart the client instead, it almost certainly gets stuck waiting again with the same message after a few more iterations.
> 
> Example output when running loop from the provided snippet:
> 
> ```
> ...
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> ^C[INFO] [rclcpp]: signal_handler(signal_value=2)
> [ERROR] [minimal_client]: client interrupted while waiting for service to appear.
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> ... # x9
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> ...
> ```
> It seems there may still be an issue with the implementation.



> With the above snippet, I am able to consistently reproduce the issue with the security flag on and by making my machine busy (building a workspace with colcon). If my CPU is not busy, then I do not see the issue.
> 
> With the proposed fix in this PR, the original issue appears to go away, but I occasionally experience the client stuck waiting for the service. This looks different than the original issue since it is repeatedly printing the following message:
> 
> ```
> [INFO] [minimal_client]: waiting for service to appear...
> ```
> If I restart the server, it succeeds along with all consecutive calls to the client.
> But if I restart the client instead, it almost certainly gets stuck waiting again with the same message after a few more iterations.
> 
> Example output when running loop from the provided snippet:
> 
> ```
> ...
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> ^C[INFO] [rclcpp]: signal_handler(signal_value=2)
> [ERROR] [minimal_client]: client interrupted while waiting for service to appear.
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> [INFO] [minimal_client]: result of 41 + 1 = 42
> ... # x9
> [INFO] [minimal_client]: waiting for service to appear...
> [INFO] [minimal_client]: waiting for service to appear...
> ...
> ```
> It seems there may still be an issue with the implementation.

DDS RTPS does distributed discovery a participant.
And then ROS2 makes for client/service
1. Client request publisher and Service request subscriber
2. Client response subscriber and Service response publisher

rmw_service_server_is_available is a API that the client checks to be able to send request to service. 
In point of DDS layer, client can send a request properly after matching between RTPS remote/local reader and RTPS remote/local writer.
In current code, it returns true not before matching between reader&writer. So the ROS2 client tried to send a request to service (at this time, client does not print waiting message anymore) but the client cannot properly send the request to service and client waits forever.
For resolving this issue (https://github.com/ros2/rmw_fastrtps/issues/239), I added logic to check that the reader-writer are matched.

It can take some times to find and set properly between the client/service because of distributed system. I think that it is the reason that rmw_service_server_is_available API is needed. Client prints below message until rmw_service_server_is_available is true. I think the message is not issue.
```
[INFO] [minimal_client]: waiting for service to appear...
```

jacobperron:
>  Client prints below message until rmw_service_server_is_available is true. I think the message is not issue.

What I experience is that the client never sees the server as available (ie. the example prints the waiting message indefinitely), even though the server apparently running fine. I think the expected behavior is for `rmw_service_server_is_available` to ultimately return true in this scenario, although it may take some time.

PR #243, opened 2018-11-30
RMW_FastRTPS configuration from XML only

JuanCarlos-eProsima:
We have implemented a new feature to apply the RMW configuration only from XML files, avoiding the current configuration set by code. To do that, rmw_fastrtps checks an environment variable called `RMW_FASTRTPS_LEAVE_MIDDLEWARE_DEFAULT_QOS`

Connects to ros2/ros2#589

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
@@ -37,6 +37,7 @@ typedef struct CustomParticipantInfo
   ReaderInfo * secondarySubListener;
   WriterInfo * secondaryPubListener;
   rmw_guard_condition_t * graph_guard_condition;
+  bool leave_middleware_default_qos;
 } CustomParticipantInfo;
 
 class ParticipantListener : public eprosima::fastrtps::ParticipantListener
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
@@ -37,6 +37,12 @@ typedef struct CustomParticipantInfo
   ReaderInfo * secondarySubListener;
   WriterInfo * secondaryPubListener;
   rmw_guard_condition_t * graph_guard_condition;
+
+  // Flag to establish if the QoS of the participant,
+  // its publishers and its subscribers are going
+  // to be configured only from an XML file or if 
+  // their settings are going to be overwritten by code 
+  // with the default configuration.
   bool leave_middleware_default_qos;
 } CustomParticipantInfo;
 
--- a/rmw_fastrtps_shared_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_node.cpp
@@ -106,9 +106,9 @@ create_node(
     node_impl = new CustomParticipantInfo();
 
     node_impl->leave_middleware_default_qos = false;
-    const char * env_var = "RMW_FASTRTPS_LEAVE_MIDDLEWARE_DEFAULT_QOS";
+    const char * env_var = "RMW_FASTRTPS_USE_QOS_FROM_XML";
      // Check if the configuration from XML has been enabled from 
-    // the RMW_FASTRTPS_LEAVE_MIDDLEWARE_DEFAULT_QOS env variable.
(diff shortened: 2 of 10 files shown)


JuanCarlos-eProsima:
Hi @dirk-thomas, 
We have changed the environment variable to use your first suggestion `RMW_FASTRTPS_USE_QOS_FROM_XML`. It explains better this new feature.

mjcarroll:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5755/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2341/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4756/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5656/)

mjcarroll:
@JuanCarlos-eProsima It seems there are several linter errors, can you please get those sorted out?

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5774/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2353/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4769/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5670/)

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -37,6 +37,7 @@ typedef struct CustomParticipantInfo
   ReaderInfo * secondarySubListener;
   WriterInfo * secondaryPubListener;
   rmw_guard_condition_t * graph_guard_condition;
+  bool leave_middleware_default_qos;
```
mjcarroll: I think it would be helpful here to have a bit of documentation describing the behavior of each state.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
```diff
@@ -104,6 +104,25 @@ create_node(
 
   try {
     node_impl = new CustomParticipantInfo();
+
+    node_impl->leave_middleware_default_qos = false;
+    const char * env_var = "RMW_FASTRTPS_USE_QOS_FROM_XML";
+    // Check if the configuration from XML has been enabled from
+    // the RMW_FASTRTPS_USE_QOS_FROM_XML env variable.
+    char * config_env_val = nullptr;
+#ifndef _WIN32
```
mjcarroll: Since we are already depending on `rcutils`, https://github.com/ros2/rcutils/blob/master/include/rcutils/get_env.h#L62 would be a reasonable choice here.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
```diff
@@ -246,11 +265,32 @@ __rmw_create_node(
   // since the participant name is not part of the DDS spec
   participantAttrs.rtps.setName(name);
 
+  bool leave_middleware_default_qos = false;
+  const char * env_var = "RMW_FASTRTPS_USE_QOS_FROM_XML";
+  // Check if the configuration from XML has been enabled from
+  // the RMW_FASTRTPS_USE_QOS_FROM_XML env variable.
+  char * config_env_val = nullptr;
+#ifndef _WIN32
```
mjcarroll: Same as above, this functionality is in `rcutils.`

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
JuanCarlos-eProsima: I agree. I have added an explanation for this attribute because only with the name it could generate misunderstandings

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
JuanCarlos-eProsima: I checked this function before implementing the feature and I chose to implement the getenv again in this package because the function in rcutils is not safe. It stores the result of the "getenv" in a static variable that could be overwritten by another call before taking the result for this feature.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
JuanCarlos-eProsima: Please, check the comment in the previous conversation.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
mjcarroll: Sounds good. Thanks.

PR #218, opened 2018-07-25
split rmw_fastrtps_cpp into static and dynamic rmw impl

dirk-thomas:
This replaces #203. It reverts the latest commit on master before applying all the changes from the PR. I rewrote the commits slightly to avoid back-and-forth changes as well as separate fix-commits:
* I merged the commit 9b3be5dee8f0bd78b87e3a3260de4f2a6eda98db into 02783b91500b37eadd95e5c862295a601545f447.
* I merged the commit 845634cc2ff1a46dfd6d95e810b87207f21127cd into eb7d8e376bdc9c4513c8677f066a4779ae2a1f71.
* I split the commit 1429803fb7d393b4d46e19159e26035c970abf9e and merged it into 61b42337ddfd7aee402e898af84c11b93bcdb554, e374da92bbe6a1a1c415d37a9a602fe6de9e23b1 and eb7d8e376bdc9c4513c8677f066a4779ae2a1f71.
* I fixed the commit message of 61b42337ddfd7aee402e898af84c11b93bcdb554.

This also allows us to run CI against the current state again:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4949/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1769/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4108/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4965/)

PR #250, opened 2018-12-28
remove duplicate namespace_prefix symbols

dirk-thomas:
Related to ros2/ros2#635.

When trying to compile static libraries the symbols being present in all three packages collide with each other. Therefore this patch exposes the symbols in the shared packages and updates the other two packages to use the header from the shared package instead of defining their own.

This should be merged without squashing since one of the files is moved in the first commit.


dirk-thomas:
Since compiling statically was possible some time ago this patches fixes a regression introduced in #218.

PR #252, opened 2019-01-10
pass context to wait set and fini context

wjwwood:
Connects to ros2/rmw#163


rmw_fastrtps_shared_cpp/src/rmw_wait_set.cpp
```diff
@@ -24,8 +24,16 @@
...
-__rmw_create_wait_set(const char * identifier, size_t max_conditions)
+__rmw_create_wait_set(const char * identifier, rmw_context_t * context, size_t max_conditions)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    init context,
+    context->implementation_identifier,
+    identifier,
+    // TODO(wjwwood): replace this with RMW_RET_INCORRECT_RMW_IMPLEMENTATION when refactored
+    return NULL);
```
clalancette: `return nullptr` instead?

PR #262, opened 2019-03-14
update listener logic for accurate counting

dirk-thomas:
Hopefully resolved the permanent failure from ros2/build_cop#166. The test is still flaky though and won't address ros2/build_cop#155.


dirk-thomas:
> IIUC it looks like this is guarding against a mismatch in the number of eprosima::fastrtps::rtps::MATCHED_MATCHING and eprosima::fastrtps::rtps::REMOVED_MATCHING events.

Correct.

> how can that mismatch be reproduced?

Unfortunately I don't have a reproducible example for that.

dirk-thomas:
Worst case this is only a cleanup change aligning the logic to be the same as in the other two listeners :wink:

dirk-thomas:
Follow up commit 3ab987a43c225bad81947fcf406408d678edb72a to avoid compiler warning on Windows.

PR #264, opened 2019-03-20
Fixed race condition between taking sample and updating counter.

### LINKED ISSUE #258: FastRTPS drops messages under stress
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - source 
- Version or commit hash:
  - https://github.com/ros2/ros2/blob/release-crystal-20190117/ros2.repos
- DDS implementation:
  - Fast-RTPS
  - rclcpp

#### Steps to reproduce issue
1. Compile ros2 from https://github.com/ros2/ros2/blob/release-crystal-20190117/ros2.repos
2. Compile this package https://github.com/ApexAI/experiments
3. Run the test in https://github.com/ApexAI/experiments
4. stress the system with `stress-ng --matrix 0`

#### Expected behavior
Messages in this line should not get lost: https://github.com/ApexAI/experiments/blob/master/stress_test/src/stress_test.cpp#L77

#### Actual behavior
Messages in this line get lost: https://github.com/ApexAI/experiments/blob/master/stress_test/src/stress_test.cpp#L77

#### Additional information
* At least FastRTPS 1.7.1 and 1.7.0 lose sample, especially when the system is on high load.
* The problem does not occur with RTI Connext Pro.

The issue might be connected to #157.

@deeplearningrobotics and @pbaughman found the issue and created a reproducer.

richiware:
This PR fixes #258.

Internal Fast-RTPS counter and rmw_fastrtps `data_` counter can get out of sync if Fast-RTPS receives a new sample after rmw_fastrtps takes a sample and before it decreases `data_` counter.

The solution I provide is having the real count in the internal Fast-RTPS counter and `data_` always will get this information.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
@@ -102,15 +102,15 @@ public:
   }
 
   void
-  data_taken()
+  data_taken(eprosima::fastrtps::Subscriber * sub)
   {
     std::lock_guard<std::mutex> lock(internalMutex_);
 
     if (conditionMutex_ != nullptr) {
       std::unique_lock<std::mutex> clock(*conditionMutex_);
-      --data_;
+      data_ = sub->getUnreadCount();
     } else {
-      --data_;
+      data_ = sub->getUnreadCount();
     }
   }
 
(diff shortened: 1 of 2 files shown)


clalancette:
This looks good to me and is one we'll want to backport to Crystal as part of https://github.com/ros2/ros2/issues/669 .  Here's CI on the change:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6463/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2833/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5308/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6282/)

clalancette:
The test failures on aarch64 are known failures, so this looks good to me.

clalancette:
CI looked good, and the code looked good to me, so merging.  Thanks for the patch.

PR #267, opened 2019-03-25
Add function to get publisher actual qos settings

ivanpauno:
Implements api added in https://github.com/ros2/rmw/pull/169

Connects to ros2/rmw#169

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
@@ -101,4 +101,61 @@ __rmw_publisher_count_matched_subscriptions(
 
   return RMW_RET_OK;
 }
+
+rmw_ret_t
+__rmw_publisher_get_actual_qos(
+  const rmw_publisher_t * publisher,
+  rmw_qos_profile_t * qos)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(qos, RMW_RET_INVALID_ARGUMENT);
+
+  auto info = static_cast<CustomPublisherInfo *>(publisher->data);
+  if (info == nullptr) {
+    return RMW_RET_ERROR;
+  }
+  eprosima::fastrtps::Publisher * fastrtps_pub = info->publisher_;
+  if (fastrtps_pub == nullptr) {
+    return RMW_RET_ERROR;
+  }
+  const eprosima::fastrtps::PublisherAttributes attributes =
+    fastrtps_pub->getAttributes();
+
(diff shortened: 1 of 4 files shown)


wjwwood:
> The function doesn't get the actual value of avoid_ros_namespace_conventions field.

Can you just copy the value over from the stored input QoS? Maybe at the rcl level. It would be better to have it set correctly by the time the user gets it just to avoid unexpected behavior and more documentation to cover the special case.

> Also, in case that the underlying qos setting obtained has not equivalent in the ros rwm side, that value is set to RMW_*_SYSTEM_DEFAULT. I will also add that as a comment in the doc of the function.

I'd actually prefer new enum's like `RMW_*_UNKNOWN` so that it is clear what's happening, but I'm interested to see what others feel like. It seems like a shortcut that we might regret in the future.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
```diff
@@ -101,4 +101,61 @@ __rmw_publisher_count_matched_subscriptions(
...
+
+  auto info = static_cast<CustomPublisherInfo *>(publisher->data);
+  if (info == nullptr) {
+    return RMW_RET_ERROR;
+  }
+  eprosima::fastrtps::Publisher * fastrtps_pub = info->publisher_;
+  if (fastrtps_pub == nullptr) {
+    return RMW_RET_ERROR;
+  }
+  const eprosima::fastrtps::PublisherAttributes attributes =
```
hidmic: @ivanpauno just curious, can we get this one as a `const ref` instead?

ivanpauno: Yes, I'll modify it.

ivanpauno:
> Can you just copy the value over from the stored input QoS? Maybe at the rcl level. It would be better to have it set correctly by the time the user gets it just to avoid unexpected behavior and more documentation to cover the special case.

I prefer checking it at rcl level (where it's actually stored). At rmw level it can't been checked now. It could be checked if we store it somewhere.

> I'd actually prefer new enum's like RMW_*_UNKNOWN so that it is clear what's happening, but I'm interested to see what others feel like. It seems like a shortcut that we might regret in the future.

Sounds good to me.

ivanpauno:
@hidmic @wjwwood I've already targeted all the comments. avoid_ros_namespace_conventions field is filled with the correct value at rcl level, and not here.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6570/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2908/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5386/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6371/)

PR #270, opened 2019-03-28
Create API to expose Locator address of a participant from FastRTPS to ros2

sagniknitr:
One usecase maybe  for cli tool like rosnode machine info in ROS1

Signed-off-by: sagnikbasu95@gmail.com


sagniknitr:
will recreate after DCO changes

PR #271, opened 2019-04-03
Add qos interfaces with no-op

ross-desmond:
**Summary**
Add unsupported qos checks on initialize.

Signed-off-by: Ross Desmond <44277324+ross-desmond@users.noreply.github.com>

Connects to ros2/rmw#171


rmw_fastrtps_cpp/src/qos.hpp
```diff
@@ -39,7 +39,16 @@ bool
...
+
+RMW_LOCAL
+bool
+is_time_default(
+  const rmw_time_t & time);
+
+RMW_LOCAL
+bool
+is_valid_qos(
+  const rmw_qos_profile_t & qos_policies);
```
wjwwood: C doesn't have pass by reference, so you shouldn't use it in an extern C.

wjwwood: I see that existing functions do this, which is also wrong. It's basically a compiler feature, but we shouldn't rely on it unless we absolutely must:

https://stackoverflow.com/questions/1906000/c-by-reference-argument-and-c-linkage

wjwwood: Likely these simply do not need to be inside of an extern C as they will not be called from C (they're not part of the rmw API for example).

rmw_fastrtps_dynamic_cpp/src/qos.cpp
```diff
@@ -157,4 +157,32 @@ get_datawriter_qos(
...
+    RMW_SET_ERROR_MSG("Lifespan unsupported for fastrtps");
+    return false;
+  }
+  if (qos_policies.liveliness == RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE ||
+    qos_policies.liveliness == RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC)
+  {
+    RMW_SET_ERROR_MSG("Liveliness unsupported for fastrtps");
+    return false;
+  }
+  return true;
```
wjwwood: Is this duplicate code from above? Maybe it can be consolidated?

rmw_fastrtps_shared_cpp/src/qos.cpp
```diff
@@ -157,4 +157,32 @@ get_datawriter_qos(
...
+    RMW_SET_ERROR_MSG("Lifespan unsupported for fastrtps");
+    return false;
+  }
+  if (qos_policies.liveliness == RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE ||
+    qos_policies.liveliness == RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC)
+  {
+    RMW_SET_ERROR_MSG("Liveliness unsupported for fastrtps");
+    return false;
+  }
+  return true;
```
wjwwood: And again, looks duplicated.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
```diff
@@ -363,6 +364,33 @@ __rmw_destroy_node(
...
+  if (nullptr == node_info) {
+    RMW_SET_ERROR_MSG("node info handle is null");
+    return RMW_RET_ERROR;
+  }
+  if (nullptr == node_info->participant) {
+    RMW_SET_ERROR_MSG("node internal participant is invalid");
+    return RMW_RET_ERROR;
+  }
+
+  // node_info->participant->assert_liveliness();
```
wjwwood: I guess this is planed for the future? But should this function not fail in the meantime?

wjwwood: Sorry, I see that it does fail with `RMW_RET_UNSUPPORTED`, but maybe consider still setting the error msg in this case.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -203,6 +204,12 @@ __rmw_wait(
     }
   }
 
+  if (events) {
+    for (size_t i = 0; i < events->event_count; ++i) {
+      // no-op
+    }
+  }
```
wjwwood: Maybe just comment this or remove it and place a TODO?

rmw_fastrtps_cpp/src/qos.hpp
mm318: Agreed that the existing functions that do this are wrong. Also, if `extern "C"` is already used during the declaration of the function (such as in the header file), it is not needed again when defining the function (as in the .cpp file).

I have removed the use of `extern "C"` in this case.

rmw_fastrtps_dynamic_cpp/src/qos.cpp
mm318: It does looks like this is duplicate code. I have removed this file.

rmw_fastrtps_shared_cpp/src/qos.cpp
mm318: This copy of `qos.cpp` in the `rmw_fastrtps_shared_cpp` package will be kept.

rmw_fastrtps_shared_cpp/src/rmw_node.cpp
mm318: Added error message.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
mm318: Commented out and added a `TODO`.

PR #274, opened 2019-04-30
Updates for preallocation API.

mjcarroll:
Stubs out the expanded API, which isn't used in FastRTPS currently

Connects to ros2/rmw#159

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_publisher.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_publisher.cpp
@@ -36,23 +36,23 @@ extern "C"
 {
 rmw_ret_t
 rmw_init_publisher_allocation(
-  const rosidl_message_type_support_t * type_supports,
+  const rosidl_message_type_support_t * type_support,
   const rosidl_message_bounds_t * message_bounds,
   rmw_publisher_allocation_t * allocation)
 {
-  (void) type_supports;
+  // Unused in current implementation.
+  (void) type_support;
   (void) message_bounds;
   (void) allocation;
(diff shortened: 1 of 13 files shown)


rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -34,6 +34,27 @@ using TopicDataType = eprosima::fastrtps::TopicDataType;
...
+rmw_init_publisher_allocation(
+  const rosidl_message_type_support_t * type_support,
+  const rosidl_message_bounds_t * message_bounds,
+  rmw_publisher_allocation_t * allocation)
+{
+  // Unused in current implementation.
+  (void) type_support;
+  (void) message_bounds;
+  (void) allocation;
+  return RMW_RET_OK;
```
clalancette: I know that this is a stub that is meant to be filled out later.  But until later comes, I think this should probably return an error since we aren't actually doing what the user asked.  Same below.  Thoughts?

rmw_fastrtps_cpp/src/rmw_take.cpp
```diff
@@ -24,41 +24,51 @@
...
 {
 rmw_ret_t
-rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * taken)
+rmw_take(
+  const rmw_subscription_t * subscription,
+  void * ros_message,
+  bool * taken,
+  rmw_subscription_allocation_t * allocation)
 {
+  (void) allocation;
```
clalancette: This shouldn't be necessary since we are using it by passing it through.  Same below.

PR #279, opened 2019-05-20
fix log_debug typo in rmw_count

1r0b1n0:
Signed-off-by: Robin Vanhove <rov@robopec.com>


dirk-thomas:
Thank you for the patch.

PR #282, opened 2019-05-23
Disable all liveliness until it is actually supported

emersonknapp:
When we first implemented QoS event support for Fast-RTPS 1.8, we were under the impression that AUTOMATIC Liveliness callbacks would be supported in the release, but this turned out to not be the case.

Fixes https://github.com/ros2/demos/issues/351

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -161,9 +161,10 @@ bool
 is_valid_qos(const rmw_qos_profile_t & qos_policies)
 {
   if (qos_policies.liveliness == RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE ||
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@emersonknapp with this PR Fast-RTPS liveliness tests are now failing on all platforms. Would you suggest we introduce skips on those tests until the implementation is ready or make a different adjustment?

emersonknapp:
@nuclearsandwich I was thinking to introduce the fix as part of this PR https://github.com/ros2/rcl/pull/451, but we could do a separate one if you think that's better

nuclearsandwich:
> I was thinking to introduce the fix as part of this PR [ros2/rcl#451](https://github.com/ros2/rcl/pull/451), but we could do a separate one if you think that's better

Smaller, focused PRs if you would be so kind @emersonknapp.

emersonknapp:
Good point @nuclearsandwich - here's that https://github.com/ros2/rcl/pull/452

PR #284, opened 2019-06-03
get_actual_qos() is not setting all of rmw_qos_profile_t

mm318:
The current implementation of `get_actual_qos()` is not setting/output the values for lifespan, deadline, and liveliness QoS policies.

Depends on https://github.com/ros2/rmw/pull/175.


mm318:
The CI results can be found here: https://github.com/ros2/rmw/pull/175#issuecomment-498496634

PR #287, opened 2019-06-07
Implement get_actual_qos() for subscriptions

mm318:
Currently, publishers have the get_actual_qos() feature/function. It would make sense for subscriptions to have them, too.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
@@ -50,4 +50,16 @@ dds_qos_to_rmw_qos(
   const AttributeT & dds_qos,
   rmw_qos_profile_t * qos);
 
+extern template RMW_FASTRTPS_SHARED_CPP_PUBLIC
+void
+dds_qos_to_rmw_qos<eprosima::fastrtps::PublisherAttributes>(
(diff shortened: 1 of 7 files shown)


mm318:
Refactored to avoid duplicate code. Thanks!

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
```diff
@@ -28,6 +28,10 @@ class PublisherAttributes;
 }  // namespace fastrtps
 }  // namespace eprosima
 
+RMW_FASTRTPS_SHARED_CPP_PUBLIC
+bool
+is_valid_qos(const rmw_qos_profile_t & qos_policies);
```
ivanpauno: any reason for moving this declaration?

mm318: I felt it would be more consistent with the rest of the `rmw` implementations that the `dds_qos_to_rmw_qos()` follows `get_datareader_qos()`. `is_valid_qos()` is also the first function among the ones declared here to be called in a normal program flow.

ivanpauno: Sounds good

ivanpauno:
CI: https://github.com/ros2/rmw/pull/177#issuecomment-501261185

PR #294, opened 2019-06-13
Export typesupport_fastrtps package dependencies

kurcha01-arm:
Fixes rosidl_typesupport_fastrtps_c and rosidl_typesupport_fastrtps_cpp
not being listed as dependencies to this package.

Signed-off-by: Kurtis Charnock <kurtis.charnock@arm.com>


kurcha01-arm:
Hello,

[ro2/cross_compile](https://github.com/ros2/cross_compile) is failing because the linker can't find these packages. find_package() finds them when building natively but using ld's --sysroot option with a ros2 install location outside the sysroot prevents them from being found. This also stops any workarounds with options like rpath, since the sysroot path gets prefixed to any rpath specified.

By adding the packages as dependencies, find_package doesn't need to search the filesystem, fixing the problem.

See also [rosbag2/pull/132](https://github.com/ros2/rosbag2/pull/132)

rmw_fastrtps_dynamic_cpp/CMakeLists.txt
```diff
@@ -111,6 +111,8 @@ PRIVATE "RMW_FASTRTPS_DYNAMIC_CPP_BUILDING_LIBRARY")
 ament_export_include_directories(include)
 ament_export_libraries(rmw_fastrtps_dynamic_cpp)
 
+ament_export_dependencies(rosidl_typesupport_fastrtps_cpp)
+ament_export_dependencies(rosidl_typesupport_fastrtps_c)
```
dirk-thomas: Please swap the order of both line to be in alphabetical order.

kurcha01-arm: Done

PR #300, opened 2019-07-04
use target_include_directories

Karsten1987:
Signed-off-by: karsten knese <karsten.knese@us.bosch.com>

When overlaying a workspace with FastRTPS, this changed helped me to point to the local FastRTPS include directories rather than the include dirs from `/opt/ros/dashing`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7482/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3622/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6137/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7318/)

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -44,8 +44,6 @@ find_package(rosidl_generator_c REQUIRED)
 find_package(rosidl_typesupport_fastrtps_c REQUIRED)
 find_package(rosidl_typesupport_fastrtps_cpp REQUIRED)
 
-include_directories(include)
-
 add_library(rmw_fastrtps_cpp
   src/get_client.cpp
   src/get_participant.cpp
@@ -81,6 +79,11 @@ add_library(rmw_fastrtps_cpp
   src/rmw_wait_set.cpp
   src/serialization_format.cpp
   src/type_support_common.cpp
+)
+target_include_directories(rmw_fastrtps_cpp
+  INTERFACE ${fastcdr_INCLUDE_DIR}
+  INTERFACE ${fastrtps_INCLUDE_DIR}
+  PRIVATE include
 )
 target_link_libraries(rmw_fastrtps_cpp
   fastcdr fastrtps)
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
@Karsten1987 is this ready for review?

Karsten1987:
yes, it's ready for review!

Karsten1987:
One more round of CI given that this PR dates back quite a bit:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8188/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4165/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6668/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8064/)

dirk-thomas:
Depending on which of the two dependencies is in the underlay / overlay this won't cut it. The include directories need to be sorted explicitly - e.g. using `ament_target_dependencies`.

Karsten1987:
Can you elaborate how `ament_target_dependencies` should be used in the overlay then? Not sure I follow your comment.
Also, does that mean you object to this PR?

dirk-thomas:
> Can you elaborate how `ament_target_dependencies` should be used in the overlay then?

I don't understand what "Used in the overlay" means. That CMake function uses `ament_include_directories_order` internally to order the include directories of all dependencies according to the workspace order. You can use either of the two to order the include directories.

> does that mean you object to this PR?

Since the PR aims to fix the order I would expect it to not do that at the moment correctly. Based on that I think it needs to be updated.

Karsten1987:
@dirk-thomas the patch might not address the overlay situation, but I feel it's still an appropriate patch and makes sense to merge - nonetheless because of using a modern CMake approach. Let me know what you think.

dirk-thomas:
> the patch might not address the overlay situation, but I feel it's still an appropriate patch and makes sense to merge

The problem with the patch is that it even introduces wrongly ordered include directories in certain cases. E.g. if `fastcdr` is in an underlay while `fastrtps` is in an overlay. That is why this needs to use logic which performs the ordering based on the location of the packages in the workspaces. So still no, I don't think this should be merged as is.

PR #302, opened 2019-07-16
Support Fast-RTPS 1.9.0

wjwwood:
Needed to support Fast-RTPS 1.9.0.

Connects to https://github.com/ros2/ros2/pull/747


dirk-thomas:
What is the status of this ticket?

wjwwood:
We still don't have green CI for adding it. I haven't had time to finish ironing out the issues.

PR #308, opened 2019-07-26
Add mechanism to forward fastrtps logging into rcl logging system.

windelbouwman:
This change forwards the fastrtps logging into the rcl logging. Note that you require a debug build of fastrtps to receive logging of detailed levels.

This is more a proof-of-concept. I'm not sure what the appropriate way is.

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -12,10 +12,46 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <fastrtps/log/Log.h>
+#include <memory>
+
+#include "rcutils/logging_macros.h"
 #include "rmw/impl/cpp/macros.hpp"
 #include "rmw/rmw.h"
-
 #include "rmw_fastrtps_cpp/identifier.hpp"
+
+using eprosima::fastrtps::LogConsumer;
+using eprosima::fastrtps::Log;
+
+class RcutilsLogConsumer : public LogConsumer
+{
+public:
+  RcutilsLogConsumer()
+  {
+  }
+
+  virtual ~RcutilsLogConsumer()
+  {
+  }
+
+  virtual void Consume(const Log::Entry & entry)
+  {
+    const char * logger_name = "rmw_fastrtps";
+    switch (entry.kind) {
+      case Log::Kind::Error:
+        RCUTILS_LOG_ERROR_NAMED(logger_name, "%s", entry.message.c_str());
+        break;
+
+      case Log::Kind::Warning:
+        RCUTILS_LOG_WARN_NAMED(logger_name, "%s", entry.message.c_str());
+        break;
+
+      case Log::Kind::Info:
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -12,11 +12,47 @@
...
+  {
+  }
+
+  virtual ~RcutilsLogConsumer()
+  {
+  }
+
+  virtual void Consume(const Log::Entry & entry)
+  {
+    const char * logger_name = "rmw_fastrtps";
```
jubeira: This could be a static `constexpr`.

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -77,6 +113,14 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     options->implementation_identifier,
     eprosima_fastrtps_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+
+  // Forward fastrtps log messages into rcl logging system.
+  LogConsumer * my_consumer = new RcutilsLogConsumer();
```
jubeira: Create the consumer directly inside the `unique_ptr` if that's the only use it has.

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -77,6 +113,14 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
...
+  Log::Reset();
+  Log::ClearConsumers();
+  Log::RegisterConsumer(std::unique_ptr<LogConsumer>(my_consumer));
+  Log::SetVerbosity(Log::Info);
```
jubeira: This should be configurable at some point.

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -12,11 +12,47 @@
...
-
 #include "rmw_fastrtps_cpp/identifier.hpp"
 
+using eprosima::fastrtps::LogConsumer;
+using eprosima::fastrtps::Log;
+
+class RcutilsLogConsumer : public LogConsumer
+{
+public:
+  RcutilsLogConsumer()
```
jubeira: I'd change to `= default`.

jubeira:
Hi @windelbouwman, did you have the chance to check my previous comment?

> Do you have a use-case and a simple way to test this feature? I've compiled FastRTPS in Debug mode and enabled debug messages in its CMakeLists file, but I still couldn't see any messages coming out when creating nodes and publishing to topics. Am I missing something?

Thanks!

mabelzhang:
@windelbouwman, did you still want to continue on this PR?

windelbouwman:
@mabelzhang I'm unsure. I still do not know what is the best strategy to forward the fastrtps logging. I would like to include it into the ros logging, so all logging is bundled in one place. Hence this PR. Feel free to close the pull request. I wanted to ask the question whether this was done already or not.

gonzodepedro:
This is not done already. Feel free to reopen with addressed comments and a use-case.

PR #311, opened 2019-08-29
Return specific error code when node is not found

ivanpauno:
Addresses https://github.com/ros2/ros2cli/pull/322#issuecomment-525364023

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
@@ -81,7 +81,11 @@ rmw_ret_t __get_guid_by_name(
         });
 
     if (guid_node_pair == impl->listener->discovered_names.end()) {
-      RMW_SET_ERROR_MSG("Node name not found: ns='%s', name='%s", node_namespace, node_name);
+      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+        "Node name not found: ns='%s', name='%s",
+        node_namespace,
+        node_name
+      );
       return RMW_RET_NODE_NAME_NON_EXISTENT;
     }
     guid = guid_node_pair->first;
--- a/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
@@ -81,11 +81,8 @@ rmw_ret_t __get_guid_by_name(
         });
 
     if (guid_node_pair == impl->listener->discovered_names.end()) {
-      RCUTILS_LOG_ERROR_NAMED(
-        kLoggerTag,
-        "Unable to find GUID for node: %s", node_name);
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
```diff
@@ -85,7 +85,7 @@ rmw_ret_t __get_guid_by_name(
         kLoggerTag,
         "Unable to find GUID for node: %s", node_name);
       RMW_SET_ERROR_MSG("Unable to find GUID for node ");
-      return RMW_RET_ERROR;
+      return RMW_RET_NON_EXISTENT_NODE_NAME;
```
dirk-thomas: Is the caller still expected to clear the error message? Or with the explicit return code the error message could be removed?

ivanpauno: We usually set the error message when returning an explicit return code, so I will follow that.
Deleting the logging of the error, as is unnecessary.

rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
```diff
@@ -59,7 +59,7 @@ rmw_ret_t __get_guid_by_name(
   const char * node_namespace, GUID_t & guid)
 {
   auto impl = static_cast<CustomParticipantInfo *>(node->data);
-  if (strcmp(node->name, node_name) == 0) {
+  if (strcmp(node->name, node_name) == 0 && strcmp(node->namespace_, node_namespace) == 0) {
```
ivanpauno: Realized of this preexisting bug when uncommenting some tests in `rcl` `test_graph.cpp`.

dirk-thomas:
> You might consider curating the changes into two commits

The unrelated change should even go into a separate PR imo. The change to introduce a new return code is already huge (across many repos). It shouldn't be combined with other changes.

ivanpauno:
> > You might consider curating the changes into two commits
> 
> The unrelated change should even go into a separate PR imo. The change to introduce a new return code is already huge (across many repos). It shouldn't be combined with other changes.

I needed to add name and namespace validation in `rcl`, because if not some function calls in tests started returning `RCL_RET_NODE_NAME_NON_EXISTENT` when they shouldn't. After that, the PR in `rcl` went big.
Some parts of the tests where also commented, supposedly because we didn't have name/namespace validation, so I uncommented them. After that I realized that the error was this, and I corrected it.

I can comment again the problematic parts of the code in `rcl`, undo this extra change, and send them in a new PR. But I cannot delete the adding of node/namespace validation in `rcl` PR, because I wouldn't get green CI (and that's much bigger than this correction).

I think that splitting this PR in two commits (doing rebase and merge) is the easier way to go.

ivanpauno:
I've already sent and merged the unrelated fix in a separated PR https://github.com/ros2/rmw_fastrtps/pull/316

PR #312, opened 2019-09-03
Switch to one Participant per Context

ivanpauno:
Implementation of https://github.com/ros2/design/pull/250.


ivanpauno:
@wjwwood @hidmic It would be good to have some feedback on this.
As I commented in the description I combined sending a message with using group data qos, but it doesn't seem to be working well on `Fast-RTPS`. In case we think this approach is preferred, I will create an issue in `Fast-RTPS`, to see if it can be corrected.
If not, we can switch to the other approach, sending a big message with all the information.

ivanpauno:
Ok, I finally finished migrating from group data usage, to avoid using it.
The graph API is completely implemented now.

I have to do a rebase (which will be a headache), but this is ready for review.
I did some manual testing with publish/subscribe, client/services, and using the CLI and it's working fine.

@wjwwood @hidmic Please take a look if possible, together with https://github.com/ivanpauno/rmw_dds_common/pull/4.

P.S.: I will be OoO until 18th Nov.

ivanpauno:
Checking what is the result of CI ...

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8679/)

ivanpauno:
I've rebased with master.
I have done some manual testing locally, and it seems to be working ok.

There's still work to be done (propagating some API changes upstream, updating sros2).

Also, I'm concerned on how we will implement this in all the rmw implementations:
- If we do it one at a time, there are things that are going to work different in them and we will have to disable some cross-vendor tests. Also, I don't know how we will handle security (sros2), with two different implementations living together.
- If we want to change all of them at the same time, it will be a lot of work ....

----

@dirk-thomas @wjwwood @hidmic @jacobperron This needs a first review.

rmw_fastrtps_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,169 @@
...
+          ss << common_context->graph_cache;
+          RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_cpp", "%s", ss.str().c_str());
+        }
+      }
+    }
+    if (RMW_RET_OK != rmw_destroy_wait_set(wait_set)) {
+      terminate("failed to destroy wait set");
+    }
+  }
+}
```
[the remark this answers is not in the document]
wjwwood: I mean, the other error conditions in this function use terminate, so maybe doing something different is out of scope?

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -1,3 +1,4 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
ivanpauno: Yeah, IDK.
Maybe we should just left the old copyright.

rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -37,18 +38,22 @@ rmw_create_node(
   const char * name,
   const char * namespace_,
   size_t domain_id,
-  const rmw_node_security_options_t * security_options,
+  const rmw_security_options_t * security_options,
   bool localhost_only)
 {
+  (void)domain_id;
+  (void)security_options;
+  (void)localhost_only;
```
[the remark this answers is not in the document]
ivanpauno: I moved localhost logic to the participant, and I haven't modified node API yet.

rmw_fastrtps_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,169 @@
...
+static
+void
+terminate(const char * error_message)
+{
+  RCUTILS_LOG_ERROR_NAMED(log_tag, "%s, terminating ...", error_message);
+  std::terminate();
+}
+
+void
+node_listener(rmw_context_t * context)
```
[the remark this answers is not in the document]
ivanpauno: It's forward declared as `static` in line 44.

rmw_fastrtps_cpp/src/listener_thread.cpp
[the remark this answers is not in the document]
ivanpauno: There's not much to do when this thread fails, though it's true that the "correct thing" to do is shutting down the context corresponding to this listener thread, and not terminating everything.
But currently, shutting down a context is not trivial from `rmw` (adding a interrupt guard condition to the context would be a good solution).

Another solution is to just log an error instead of terminating.
If this thread stops working, the graph is not going to be correctly updated, but other things should still work correctly.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -41,6 +41,7 @@ typedef struct CustomSubscriberInfo : public CustomEventInfo
   eprosima::fastrtps::Subscriber * subscriber_;
   SubListener * listener_;
   rmw_fastrtps_shared_cpp::TypeSupport * type_support_;
+  rmw_gid_t subscription_gid;
```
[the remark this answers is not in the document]
ivanpauno: structs shouldn't use trailing underscore convention.
I can add it for consistency with the rest of the members of this class, but it's against the [GSG](https://google.github.io/styleguide/cppguide.html#Variable_Names).

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/namespace_prefix.hpp
```diff
@@ -29,6 +29,16 @@ RMW_FASTRTPS_SHARED_CPP_PUBLIC extern const char * const ros_service_response_pr
 RMW_FASTRTPS_SHARED_CPP_PUBLIC extern const std::vector<std::string> _ros_prefixes;
 }  // extern "C"
 
+/// Returns `name` stripped of `prefix`.
+/**
+ * \param[in] name string that will be stripped from prefix
+ * \param[in] prefix prefix to be stripped
+ * \return name stripped of prefix, or
+ * \return "" if name doesn't starti with prefix
```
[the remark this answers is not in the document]
ivanpauno: It's stripping the prefix, and if not returning `""` to indicate an error.
IMO, `_resolve_name` isn't clearer.

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -0,0 +1,262 @@
...
+      participantAttrs.rtps.builtin.domainId = static_cast<uint32_t>(number);
+    }
+  } else {
+    participantAttrs.rtps.builtin.domainId = static_cast<uint32_t>(domain_id);
+  }
+
+  const char * ros_local_host_env_val = NULL;
+  bool localhost_only =
+    rcutils_get_env("ROS_LOCALHOST_ONLY", &ros_local_host_env_val) == NULL &&
+        ros_local_host_env_val != NULL && strcmp(ros_local_host_env_val, "1") == 0;
```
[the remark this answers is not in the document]
ivanpauno: yeah, I moved it here temporally. I'm going to clean this.

rmw_fastrtps_cpp/src/listener_thread.cpp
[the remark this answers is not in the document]
hidmic: It's fine to `std::terminate` for now, let's just log enough for the user to know why his/her process aborted. 

Ideally, yes, I think it'd be great for this thread to just shutdown the context it belongs to, but as @wjwwood says it's out of scope for this PR.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/namespace_prefix.hpp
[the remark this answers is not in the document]
hidmic: It's a super minor detail. But the current name suggests something else (i.e. strip/trim prefix/characters from string if present, like every other strip/trim function out there) and not what this particular function actually does (unless you read the docs or the source code).

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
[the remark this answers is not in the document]
wjwwood: Maybe the case here is that this class should be a `class` and not a `struct`... If it has methods and private members.

ivanpauno:
It's more stable now locally, checking CI ...

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8798/)

ivanpauno:
Yet another try:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8847/)

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
```diff
@@ -50,41 +47,32 @@ __rmw_destroy_publisher(
...
 
-  if (!publisher) {
-    RMW_SET_ERROR_MSG("publisher handle is null");
-    return RMW_RET_ERROR;
-  }
-
-  if (publisher->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("publisher handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
```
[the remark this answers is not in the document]
ivanpauno: Moved https://github.com/ros2/rmw_fastrtps/blob/a542af36cf1cf4b6003a73b59d3363a432b271ce/rmw_fastrtps_shared_cpp/src/publisher.cpp#L38-L49.
Those arguments are passed without being used, so it's fine to check them in the called function.

rmw_fastrtps_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,181 @@
...
+        if (std::strncmp(
+            reinterpret_cast<char *>(common_context->gid.data),
+            reinterpret_cast<char *>(&msg.gid.data),
+            RMW_GID_STORAGE_SIZE) == 0)
+        {
+          // ignore local messages
+          continue;
+        }
+        common_context->graph_cache.update_participant_entities(msg);
+        if (rcutils_logging_logger_is_enabled_for("rmw_dds_common",
```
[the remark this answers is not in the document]
ivanpauno: Yeah, that's true.
Maybe using `fprintf` directly?
IMO, logging should be thread safe. I'm sure we can currently hit problems in other places.

rmw_fastrtps_cpp/src/rmw_node.cpp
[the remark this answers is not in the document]
ivanpauno: Localhost logic is now a context configuration, not a node one.
I will not change the API as part of the first iteration, to avoid modifying others rmw implementations.

rmw_fastrtps_shared_cpp/src/rmw_node_info_and_types.cpp
```diff
@@ -140,240 +93,93 @@ rmw_ret_t __validate_input(
...
-  auto impl = static_cast<CustomParticipantInfo *>(node->data);
-
-  __log_debug_information(*impl);
+  auto common_context = static_cast<rmw_dds_common::Context *>(node->context->impl->common);
 
-  GUID_t guid;
-  rmw_ret_t valid_guid = __get_guid_by_name(node, node_name, node_namespace, guid);
-  if (valid_guid != RMW_RET_OK) {
-    return valid_guid;
+  DemangleFunction no_op{[](const std::string & x) {return x;}};
```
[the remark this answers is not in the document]
ivanpauno: It's defined just in two places in different files, ant it's a quite trivial lambda.
I can add an `identity_demangle` function in `demangle.hpp` as an alternative.

ivanpauno:
Now including `rmw_fastrtps_dynamic`:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8928/)

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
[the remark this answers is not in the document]
hidmic: Right, but you dereference `publisher` in [line 51](https://github.com/ros2/rmw_fastrtps/pull/312/files#diff-1623b37ee69464abc9f38e1afc71215bR51) in hopes it's not `nullptr` but a Fast-RTPS publisher so that the cast makes sense.

rmw_fastrtps_shared_cpp/src/rmw_service.cpp
```diff
@@ -58,8 +59,37 @@ __rmw_destroy_service(
     RMW_SET_ERROR_MSG("publisher handle not from this implementation");
     return RMW_RET_ERROR;
   }
+  if (!node) {
+    RMW_SET_ERROR_MSG("service handle is null");
```
hidmic: @ivanpauno bad copy paste?

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
```diff
@@ -53,42 +49,30 @@ __rmw_destroy_subscription(
...
 
-  if (!subscription) {
-    RMW_SET_ERROR_MSG("subscription handle is null");
-    return RMW_RET_ERROR;
-  }
-
-  if (subscription->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("node handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
```
hidmic: @ivanpauno same about not removing the two checks above.

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -199,15 +202,54 @@ rmw_create_client(
...
+    common_context->graph_cache.associate_writer(
+      gid,
+      common_context->gid,
+      node->name,
+      node->namespace_);
+    gid = rmw_fastrtps_shared_cpp::create_rmw_gid(
+      eprosima_fastrtps_identifier, info->response_subscriber_->getGuid());
+    rmw_dds_common::msg::ParticipantEntitiesInfo msg =
+      common_context->graph_cache.associate_reader(
+      gid, common_context->gid, node->name, node->namespace_);
```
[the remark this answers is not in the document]
ivanpauno: Yeah, I don't like that style.
The linter is not only happy, it forces me to use that identation in L219.

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
ivanpauno: Moved to https://github.com/ros2/rmw_fastrtps/blob/96ad1d3ad6c764dd45dadf3fc3a7592fb51c58c4/rmw_fastrtps_shared_cpp/src/subscription.cpp#L47-L54.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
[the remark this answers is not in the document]
ivanpauno: It doesn't have `private` members, they are all `public`.
I added the trailing underscore for consistency. We can decide in a follow up if we want to remove all the trailing underscores of the members, switch from `struct` to `class` and make the members private, etc.

rmw_fastrtps_dynamic_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,181 @@
...
+          ss << common_context->graph_cache;
+          RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_dynamic_cpp", "%s", ss.str().c_str());
+        }
+      }
+    }
+    if (RMW_RET_OK != rmw_destroy_wait_set(wait_set)) {
+      terminate("failed to destroy wait set");
+    }
+  }
+}
```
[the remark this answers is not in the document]
ivanpauno: The problem is that I'm using `rmw` api directly in `listener_thread.cpp`, so I can't move it to shared.

ivanpauno:
CI after rebasing:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9170/)

SidFaber:
Can you also update the [DDS-Security design document](https://github.com/ros2/design/blob/gh-pages/articles/ros2_dds_security.md#security-files-for-each-domain-participant) on how the security files are now located?

ivanpauno:
> Can you also update the [DDS-Security design document](https://github.com/ros2/design/blob/gh-pages/articles/ros2_dds_security.md#security-files-for-each-domain-participant) on how the security files are now located?

I'm not planning to do so now. I can do that after all the opened PRs get reviewed and merged, but I think people in the security working group is more suitable than me for doing that task.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
```diff
@@ -50,41 +47,30 @@ __rmw_destroy_publisher(
...
-    if (info->listener_ != nullptr) {
-      delete info->listener_;
+  auto common_context = static_cast<rmw_dds_common::Context *>(node->context->impl->common);
+  auto info = static_cast<const CustomPublisherInfo *>(publisher->data);
+  {
+    // Update graph
+    std::lock_guard<std::mutex> guard(common_context->node_update_mutex);
+    rmw_dds_common::msg::ParticipantEntitiesInfo msg =
+      common_context->graph_cache.dissociate_writer(
+      info->publisher_gid, common_context->gid, node->name, node->namespace_);
```
hidmic: @ivanpauno indentation is atypical here.

rmw_fastrtps_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,170 @@
...
+          &taken,
+          nullptr))
+      {
+        TERMINATE("rmw_take failed");
+      }
+      if (taken) {
+        if (std::memcmp(
+            reinterpret_cast<char *>(common_context->gid.data),
+            reinterpret_cast<char *>(&msg.gid.data),
+            RMW_GID_STORAGE_SIZE) == 0)
```
wjwwood: Maybe there should be a function for comparing the gid, rather than just memcmp being used everywhere.

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -0,0 +1,180 @@
...
+  eprosima::fastrtps::PublisherAttributes publisherParam;
+
+  if (!is_valid_qos(*qos_policies)) {
+    return nullptr;
+  }
+
+  // Load default XML profile.
+  Domain::getDefaultPublisherAttributes(publisherParam);
+
+  // TODO(karsten1987) Verify consequences for std::unique_ptr?
```
wjwwood: @Karsten1987 what is this about? I assume this was copied from somewhere, but still I'm curious.

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+  participant_info(
+    rmw_fastrtps_shared_cpp::create_participant(
+      eprosima_fastrtps_identifier,
+      context->options.domain_id,
+      &context->options.security_options,
+      (context->options.localhost_only == RMW_LOCALHOST_ONLY_ENABLED) ? 1 : 0,
+      common_context.get()),
+    [&](CustomParticipantInfo * participant_info) {
+      if (RMW_RET_OK != rmw_fastrtps_shared_cpp::destroy_participant(participant_info)) {
+        fprintf(
```
wjwwood: And here, avoid `fprintf`. Probably other places.

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+          RCUTILS_STRINGIFY(__function__) "' failed.\n");
+      }
+    });
+  if (!participant_info) {
+    return RMW_RET_BAD_ALLOC;
+  }
+
+  rmw_qos_profile_t qos = rmw_qos_profile_default;
+
+  qos.avoid_ros_namespace_conventions = false;  // Change this to true after testing
```
wjwwood: What is this about?

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  common_context->gid = rmw_fastrtps_shared_cpp::create_rmw_gid(
+    eprosima_fastrtps_identifier, participant_info->participant->getGuid());
+  common_context->pub = publisher.get();
+  common_context->sub = subscription.get();
+
+  context->impl->common = common_context.get();
+  context->impl->participant_info = participant_info.get();
+
+  if (RMW_RET_OK != rmw_fastrtps_cpp::run_listener_thread(context)) {
```
wjwwood: Why not store and return the ret code from this function?

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+    rmw_ret_t ret;
+    ret = rmw_fastrtps_cpp::join_listener_thread(context);
+    if (RMW_RET_OK != ret) {
+      return ret;
+    }
+
+    auto common_context = static_cast<rmw_dds_common::Context *>(context->impl->common);
+    auto participant_info = static_cast<CustomParticipantInfo *>(context->impl->participant_info);
+
+    common_context->graph_cache.remove_participant(common_context->gid);
```
wjwwood: Can this throw or does it return something? If it does throw, is that ok?

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+    ret = rmw_fastrtps_shared_cpp::destroy_subscription(
+      eprosima_fastrtps_identifier,
+      participant_info,
+      common_context->sub);
+    if (RMW_RET_OK != ret) {
+      return ret;
```
wjwwood: Should any attempt be made to destroy the other entities before returning?

rmw_fastrtps_cpp/src/rmw_init.cpp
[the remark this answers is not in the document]
wjwwood: If you're modifying the file it's always ok to add your copyright. If the change is trivial you may leave it off. In this case I think it is ok to add it.

rmw_fastrtps_dynamic_cpp/src/listener_thread.cpp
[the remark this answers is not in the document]
wjwwood: But this part of the API is implemented in shared anyways right? You could use the shared API instead of the rmw API maybe. That would let you move it. I agree this is a lot of duplication.

rmw_fastrtps_dynamic_cpp/src/rmw_service.cpp
```diff
@@ -215,16 +218,55 @@ rmw_create_service(
   }
   memcpy(const_cast<char *>(rmw_service->service_name), service_name, strlen(service_name) + 1);
 
+  {
+    // Update graph
+    std::lock_guard<std::mutex> guard(common_context->node_update_mutex);
+    rmw_gid_t gid = rmw_fastrtps_shared_cpp::create_rmw_gid(
+      eprosima_fastrtps_identifier, info->request_subscriber_->getGuid());
+    common_context->graph_cache.associate_reader(
```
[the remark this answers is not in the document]
wjwwood: @hidmic what do you mean? You mean to (re)publish on the participant info topic (or whatever it is called) when this service is added?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/names.hpp
```diff
@@ -15,27 +15,27 @@
...
-  * \param[in] qos_profile The QoS profile for the topic.
   * \param[in] prefix Required prefix for topic name.
   * \param[in] base Required name of the topic.
   * \param[in] suffix Optional suffix for topic name.
   */
 inline
-eprosima::fastrtps::string_255
-_create_topic_name(
-  const rmw_qos_profile_t * qos_profile,
+std::string
```
wjwwood: Why change away from the string_255? I think that was done to avoid heap allocations and for performance.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/register_node.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace rmw_fastrtps_cpp
+{
+
+/// Register node in context.
+/**
+ * Function that should be called when creating a node,
+ * before using `context->impl`.
+ */
+rmw_ret_t
+register_node(rmw_context_t * context);
```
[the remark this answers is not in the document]
ivanpauno: Yeah, sounds like a bad name.
I will rename to `increase_context_impl_ref_count`.
And unregister to `decrease_context_impl_ref_count`.

It's basically doing that, creating the context impl on first use, and destroying after last usage.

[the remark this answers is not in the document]
wjwwood: 👍

rmw_fastrtps_cpp/src/gid__type_support.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+RMW_FASTRTPS_CPP_PUBLIC
+const rosidl_message_type_support_t *
+ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME(
+  rosidl_typesupport_cpp, rmw_dds_common, msg, Gid)()
```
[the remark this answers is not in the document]
ivanpauno: I'm not calling that generator, because it introduces a circular dependency.

To be more specific, I avoided depending on `rosidl_typesupport_cpp`.
It has a build export dependency on [rmw_implementation](https://github.com/ros2/rosidl_typesupport/blob/4847322f6a03e5a2ec5cf0498ec8d93d0be8f7b4/rosidl_typesupport_cpp/package.xml#L25), which makes imposible to depend on that in `rmw_fastrtps_*`.
That implementation is needed because the generated typesupport code is dependent of the available typesupports,  and actually, of the typesupports [registered by the rmw implementations](https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_cpp/CMakeLists.txt#L117).

----

If you have any suggestion to avoid duplication, it's welcomed.
But I didn't see a way of reusing that generator without generating a circular dependency, and I didn't want to also refactor how typesupports are generated to avoid the circular dependency (it would make this impossible to review).

rmw_fastrtps_cpp/src/publisher.cpp
ivanpauno: I moved that from `rmw_publisher.cpp`.
I needed to be able to create a publisher without a `node`, that's why the moved code.

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: Do we want the topic where the "ros discovery information" is published to be accessible by the user?
- Should it be a ROS hidden topic?
- Should it be a DDS topic name (no mangling applied, i.e. no `rt/...`)?
- Should it be just a normal ROS topic?

I'm inclined to option 1 or option 2, and I slightly prefer option 2, as not all the implementations will use this topic.

rmw_fastrtps_cpp/src/gid__type_support.cpp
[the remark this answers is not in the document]
wjwwood: I see, I think that duplicating it in that case is ok, but please put a big notice about why that is there.

rmw_fastrtps_cpp/src/register_node.cpp
wjwwood: We discussed this didn't we?

I would say no to the normal or hidden ROS topic, because if that were the case we should do it in `rcl` instead. We chose not to do that (if I remember correctly) because it would prevent middlewares from doing it in a more optimized way.

I think option 2 as well, but I think all the implementations should use this topic if they want to be cross-vendor compatible.

----

So my comment is more like, "well did we test it and are we going to flip it to true or not"?

I think we should flip it, seems like that would be option 2. But either way that comment should be removed, or changed. As-is it is written like a TODO.

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+rmw_ret_t
+rmw_fastrtps_cpp::unregister_node(rmw_context_t * context)
+{
+  assert(context);
+  assert(context->impl);
+
+  std::lock_guard<std::mutex> guard(context->impl->mutex);
+  if (0u == --context->impl->count) {
```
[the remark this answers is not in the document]
ivanpauno: The count is just increased and decreased by an internal API, which always first increase and then decrease, so I don't think that can happen.
I can add an assertion above:
```cpp
assert(context->impl->count);
```

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -32,6 +56,17 @@ rmw_init_options_init(rmw_init_options_t * init_options, rcutils_allocator_t all
...
+  if (!init_options->name) {
+    RMW_SET_ERROR_MSG("failed to copy context name");
+    return RMW_RET_BAD_ALLOC;
+  }
+  init_options->namespace_ = rcutils_strdup("", allocator);
+  if (!init_options->namespace_) {
+    allocator.deallocate(init_options->name, allocator.state);
+    RMW_SET_ERROR_MSG("failed to copy context namespace");
+    return RMW_RET_BAD_ALLOC;
+  }
```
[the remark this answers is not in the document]
ivanpauno: Not yet. I think we should update security documentation when everything else is ready.

rmw_fastrtps_dynamic_cpp/src/rmw_service.cpp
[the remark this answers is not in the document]
ivanpauno: He means that [associate_reader](https://github.com/ivanpauno/rmw_dds_common/blob/c56f0492191f0f4eb8d08ce9cdeb7cbabf4ba106/rmw_dds_common/include/rmw_dds_common/graph_cache.hpp#L262-L268) returns an updated `ParticipantEntitiesInfo` that I'm not using here.

When I wrote `associate_reader`, I didn't thought about services, where I'm not using the return value once.
I think the method has to be split in two:
- `associate_reader`
- `get_participant_info_msg`

> are we expecting the compiler to optimize it out?

I don't think the compiler would be able to do such an optimization, considering that `rmw_dds_common` is compiled as a separated library.

rmw_fastrtps_cpp/src/register_node.cpp
```diff
@@ -0,0 +1,217 @@
...
+      eprosima_fastrtps_identifier,
+      context->options.domain_id,
+      &context->options.security_options,
+      (context->options.localhost_only == RMW_LOCALHOST_ONLY_ENABLED) ? 1 : 0,
+      common_context.get()),
+    [&](CustomParticipantInfo * participant_info) {
+      if (RMW_RET_OK != rmw_fastrtps_shared_cpp::destroy_participant(participant_info)) {
+        fprintf(
+          stderr, "Failed to destroy participant after function: '"
+          RCUTILS_STRINGIFY(__function__) "' failed.\n");
```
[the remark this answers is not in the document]
ivanpauno: That cleanup code is only called if the function (`init_context_impl`) fails.
The intent, is to be specific about in what situation failed:
It failed when executing cleanup code due to a function failing.

Maybe:
```cpp
"'destroy_participant' failed while being executed due to 'init_context_impl' failing"
```
is clearer.

rmw_fastrtps_shared_cpp/src/demangle.cpp
```diff
@@ -52,59 +55,63 @@ _demangle_if_ros_type(const std::string & dds_type_string)
...
+  if (suffix_position != std::string::npos) {
+    if (service_name.length() - suffix_position - suffix.length() != 0) {
+      RCUTILS_LOG_WARN_NAMED("rmw_fastrtps_shared_cpp",
+        "service topic has service prefix and a suffix, but not at the end"
+        ", report this: '%s'", topic_name.c_str());
+      return "";
     }
   }
+
   if (std::string::npos == suffix_position) {
```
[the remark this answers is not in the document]
ivanpauno: Yes, don't know what I was thinking about.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
ivanpauno: I agree it's atypical, but uncrustify complains if I add an extra tab in L57.
Do you have any suggestion to make identation less atypical while pleasing uncrustify?

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: agreed

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: It has a return value that indicates if a change took place or not.
I'm unsure if logging a warning when false is desired, but I guess it might make sense.

It doesn't explicitly `throw` an exception, though it doesn't catch exceptions that may be thrown from the stdlib. I don't think that's ok, as throwing an exception throw a C boundary is UB.
Do you think that `GraphCache` itself should have all `nothrow` methods, or the caller should be responsable of doing a `try ... catch`?

----

Unrelated, but I noted some silmilar problems in the pre-existing codebase:
- There are some "potentially throwing" places without a try catch block. e.g.: [here](https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_shared_cpp/src/rmw_service_names_and_types.cpp#L68) an `std::map` is used and all methods call don't have a try/catch block.
- All `std` containers aren't using an allocator. e.g.: [here](https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp#L192-L197) (and that one can cause an allocation anytime during the execution).
- Though using `goto` to handle errors in `c` is fine, in `cpp` that might be problematic. If an exception is thrown, the cleanup code after the `fail` label will never be executed. e.g.: [here](https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L146), where [a later call](https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L171) might potentially throw.
- There are some `memcpy` usage, where a plain `c` array is copied into an object, and doing the opposite too. See https://github.com/ros2/rmw_fastrtps/pull/346.

If you agree with some of those problems, I will create a ticket to iterate on this later.

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: That's a good question. I will modify the code, so the rest of the cleanup is executed.

ivanpauno: The one I will left a return, is `join_listener_thread`. If it fails, there might be still a thread running that's using the other objects, so destroying them might lead to UB.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/names.hpp
ivanpauno: I needed `_mangle_topic_name` to return an string, as some `GraphCache` methods expect that.

----

Bottom line:
I actually regret not having had adding a template parameter for that in [GraphCache](https://github.com/ivanpauno/rmw_dds_common/blob/c56f0492191f0f4eb8d08ce9cdeb7cbabf4ba106/rmw_dds_common/include/rmw_dds_common/graph_cache.hpp#L48).
Probably, I could have also added a template parameter for the GUID, instead of transforming everything in a `rmw_gid_t`.
Moreover, it would make sense to make `GraphCache` implementation header only, so it can be optimized when compiled with an specific rmw implementation.

Having say that, I think that it makes sense to not worry about that right now, and iterate on those details right after merging this (and before refactoring another rmw implementation).

rmw_fastrtps_dynamic_cpp/src/listener_thread.cpp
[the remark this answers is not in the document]
ivanpauno: Moved to `rmw_fastrtps_shared_cpp`. I also did the same with some other duplicated code.

I couldn't do that in other places, e.g.: `rmw_init` calls `create_publisher`, which is different for each kind of typesupport.

----

Bottom line: AFAIS, only a small part of the code is different in each rmw implementation. Particularly, how the typesupport is created/registered when creating a publisher/subscription/client/service.
If it's just that, wouldn't make more sense to have just one implementation and use macros to compile both libraries from the same source?
I think that that solves the dependency inversion problem (because there's not `rmw_fastrtps_shared_cpp` library at all), it avoids boilerplate code (wrapping the `rmw_fastrtps_shared_cpp` function), and it makes the code more easy to maintain (about half the lines of code).
If we want to have one package for each of the libraries, we could still have the source code in a third package, but which just "exports" the location of the source without building the library (or another similar trick).

Sorry for the unrelated comment, I just wanted to share my thoughts about this problem.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/register_node.hpp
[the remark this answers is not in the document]
ivanpauno: Renamed

rmw_fastrtps_cpp/src/listener_thread.cpp
```diff
@@ -0,0 +1,170 @@
...
+    } catch (...) {
+      RMW_SET_ERROR_MSG("Failed to create std::thread");
+    }
+  } else {
+    RMW_SET_ERROR_MSG("Failed to create guard condition");
+  }
+  common_context->thread_is_running.store(false);
+  if (common_context->listener_thread_gc) {
+    if (RMW_RET_OK != rmw_destroy_guard_condition(common_context->listener_thread_gc)) {
+      fprintf(
```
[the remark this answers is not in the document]
ivanpauno: Fixed

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: done

ivanpauno:
@wjwwood @hidmic I think I've addressed or replied every comment.
I will rebase with master ASAP (lookly, it's just this https://github.com/ros2/rmw_fastrtps/commit/bd5507b61c7285d6083e514ccd49f00b2945baeb).

rmw_fastrtps_dynamic_cpp/src/listener_thread.cpp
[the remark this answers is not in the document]
wjwwood: Those both sound like reasonable things to do. Also, we could consider getting rid of dynamic, if we have another set of things that use the introspection type support.

But all of that is out of scope for this one, I guess. Thanks for trying to de-duplicate more.

rmw_fastrtps_dynamic_cpp/src/rmw_service.cpp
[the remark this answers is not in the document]
wjwwood: Ok, I'll leave it to @hidmic to resolve this thread or ask for you to change it in some specific way.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
wjwwood: Can 55 and 56 not fit on the same line?

wjwwood: I guess, not, you could do this:

```c++
    rmw_dds_common::msg::ParticipantEntitiesInfo msg;
    msg = common_context->graph_cache.dissociate_writer(
      ...
```

rmw_fastrtps_cpp/src/gid__type_support.cpp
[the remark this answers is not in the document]
wjwwood: You added a notice I presume?

rmw_fastrtps_cpp/src/publisher.cpp
wjwwood: Still curious what this TODO is about. @Karsten1987

rmw_fastrtps_cpp/src/register_node.cpp
wjwwood: What happened here? Looks like nothing changed.

rmw_fastrtps_cpp/src/register_node.cpp
wjwwood: > I'm unsure if logging a warning when false is desired, but I guess it might make sense.

I would never advise suppressing an error case, but if you do, you should have a comment explaining why. Silently ignoring when something can fail is always bad, imo.

----

> Do you think that `GraphCache` itself should have all `nothrow` methods, or the caller should be responsible of doing a `try ... catch`?

Uh, that depends. Both can be correct to do, it just depends on what produces the least boilerplate and/or which is easiest to understand. I don't have specific guidance on this case.

-----

> There are some "potentially throwing" places without a try catch block. e.g.: here an std::map is used and all methods call don't have a try/catch block.

Yeah, that's unfortunate, but as we didn't write this originally there are lots of issues like that... If you compare it to the Connext rmw implementation, by comparison, we're very careful about exceptions, including try-catching constructors and using macros like this:

https://github.com/ros2/rmw/blob/35ae2ff54317e8e6f30acace2fbaccd57fb2c566/rmw/include/rmw/impl/cpp/macros.hpp#L29

But that doesn't mean we shouldn't try to be careful with new code.

> All std containers aren't using an allocator. e.g.: here (and that one can cause an allocation anytime during the execution).

Yup, some work was done in this vein and that's why I asked about the replacing of the `string_255` with `std::string` in some of your changes. It seemed like a backslide to me. But as with the exceptions, we should be careful with new code but not necessarily stop to fix everything you see. You could try to collect them into a list with links, which might help someone down the line.

> Though using `goto` to handle errors in c is fine, in cpp that might be problematic. If an exception is thrown, the cleanup code after the fail label will never be executed. e.g.: here, where a later call might potentially throw.

Sure, but if the `goto` is use in the implementation of an `rmw_*` function then throwing isn't allowed anyways.

But if its a different function, then yeah you could use the scope exit pattern instead.

> There are some memcpy usage, where a plain c array is copied into an object, and doing the opposite too. See #346.

Again, bad, but really doesn't change that we should be careful with error handling here.

> If you agree with some of those problems, I will create a ticket to iterate on this later.

Yeah, they all seem like good things to at least document for later.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/names.hpp
wjwwood: I don't follow exactly what doesn't work in this case, but maybe it's worth following up on.

Can you not do `std::string s = <some func that returns string_255>`? i.e. is there no assignment conversion for `std::string`?

wjwwood:
I think I remarked on everything too.

---

> I will rebase with master ASAP (lookly, it's just this bd5507b).

Oh good, at least it was for something important 🙃.

rmw_fastrtps_cpp/src/gid__type_support.cpp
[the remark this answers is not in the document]
ivanpauno: No, I didn't see your reply. Updating now.

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: > I think we should flip it, seems like that would be option 2. But either way that comment should be removed, or changed. As-is it is written like a TODO.

Flipped.

rmw_fastrtps_cpp/src/register_node.cpp
wjwwood: This one still needs resolving too. (check return code or change error handling I think)

rmw_fastrtps_cpp/src/publisher.cpp
Karsten1987: I should have written a bit more context to this for my future self. I don't have a real good answer to do, but I assume it was trying to use `unique_ptr` instead of `new` in the code below. I guess that could eliminate the `goto fail` part.

wjwwood: Since it isn't using `unique_ptr`, yet could we just remove the TODO? @ivanpauno I'll let you decide and act on it or resolve this.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
ivanpauno: Solved in https://github.com/ros2/rmw_fastrtps/pull/312/commits/ec4a3f3e5304881aa607dd1153d0815b7509b9a6.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/names.hpp
ivanpauno: Applied that in https://github.com/ros2/rmw_fastrtps/pull/312/commits/332c7bb7956522c7cb78573b7b2451b7354cfcc5.

----

I can't do the same change in the demangle functions https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_shared_cpp/src/demangle.hpp#L21 (they were already returning a `std::string`).

The reason is that `GraphCache` expects a demangle function to be called with the following signature: https://github.com/ivanpauno/rmw_dds_common/blob/c56f0492191f0f4eb8d08ce9cdeb7cbabf4ba106/rmw_dds_common/include/rmw_dds_common/graph_cache.hpp#L326.
I used `std::string` because the idea is to reuse `GraphCache` in others rmw implementations, but of course, it can be solved by templetazing the class.

rmw_fastrtps_cpp/src/register_node.cpp
ivanpauno: > I would never advise suppressing an error case, but if you do, you should have a comment explaining why. Silently ignoring when something can fail is always bad, imo.

Updated.

> Uh, that depends. Both can be correct to do, it just depends on what produces the least boilerplate and/or which is easiest to understand. I don't have specific guidance on this case.

I will make `GraphCache` methods `noexcept`, to avoid adding `try...catch` in every call in every rmw implementation that uses it.

rmw_fastrtps_cpp/src/publisher.cpp
ivanpauno: Deleted those TODOs.
I will open an issue explaining some possible "cpp style" enhacements that includes that, as I described in another comment https://github.com/ros2/rmw_fastrtps/pull/312#discussion_r375378060.

rmw_fastrtps_cpp/src/gid__type_support.cpp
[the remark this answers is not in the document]
ivanpauno: Updated in https://github.com/ros2/rmw_fastrtps/pull/312/commits/e6c618a6a40543800123c291fd2fd2ccb4ffd600

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -12,64 +13,69 @@
...
+  return rmw_fastrtps_shared_cpp::rmw_init_options_fini(
+    eprosima_fastrtps_identifier, init_options);
 }
 
+using rmw_dds_common::msg::ParticipantEntitiesInfo;
+
 rmw_ret_t
 rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 {
+  std::unique_ptr<rmw_context_t, void (*)(rmw_context_t *)> clean_when_fail(
```
hidmic: @ivanpauno meta, not for now: this seems like a duplicate of the one in `rmw_fastrtps_dynamic_cpp`

rmw_fastrtps_dynamic_cpp/src/rmw_service.cpp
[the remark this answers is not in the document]
hidmic: > I think the method has to be split in two:

I agree. Entity association should be decoupled from message generation. But I'm fine with doing so in a follow-up PR.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/init_rmw_context_impl.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "rmw/init.h"
+#include "rmw/types.h"
+
+namespace rmw_fastrtps_shared_cpp
+{
+
+/// Increment `rmw_context_impl_t` reference count, destroying it if the count reaches zero.
+/**
+ * Function that should be called when destroying a node.
```
hidmic: @ivanpauno nit: 
```suggestion
 * Should be called when destroying a node.
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/participant.hpp
```diff
@@ -0,0 +1,41 @@
...
+  size_t domain_id,
+  const rmw_security_options_t * security_options,
+  bool localhost_only,
+  const char * context_name,
+  const char * context_namespace,
+  rmw_dds_common::Context * common_context);
+
+rmw_ret_t
+destroy_participant(CustomParticipantInfo * info);
+}  // namespace rmw_fastrtps_shared_cpp
```
hidmic: @ivanpauno nit: add space between function and closing namespace bracket.

rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
```diff
@@ -73,246 +77,38 @@ _validate_params(
...
-        _handle_topic_endpoint_info_fini(&tinfo, allocator);
-      }
-      return ret;
-    }
-    for (auto i = 0u; i < count; i++) {
-      participants_info->info_array[i] = topic_endpoint_info_vector.at(i);
-    }
-    participants_info->count = count;
+  auto common_context = static_cast<rmw_dds_common::Context *>(node->context->impl->common);
+  rmw_dds_common::GraphCache & graph_cache = common_context->graph_cache;
```
hidmic: @ivanpauno can be made `const`?

rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
```diff
@@ -325,13 +121,28 @@ __rmw_get_subscriptions_info_by_topic(
...
     allocator,
     topic_name,
-    no_mangle,
-    false, /*is_publisher*/
+    subscriptions_info);
+  if (ret != RMW_RET_OK) {
+    return ret;
+  }
+  auto common_context = static_cast<rmw_dds_common::Context *>(node->context->impl->common);
+  rmw_dds_common::GraphCache & graph_cache = common_context->graph_cache;
```
hidmic: @ivanpauno can be made `const`?

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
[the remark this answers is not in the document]
hidmic: This still applies.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
hidmic: Won't this interfere with copy elision?

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
hidmic: Same as before, `subscription` is being used in https://github.com/ros2/rmw_fastrtps/pull/312/files#diff-57b111a3ec82a425bffeb7e3633c762eR53 already.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/names.hpp
wjwwood: Ok sounds good.

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -0,0 +1,173 @@
...
+  }
+  rmw_subscription->implementation_identifier = eprosima_fastrtps_identifier;
+  rmw_subscription->data = info;
+  rmw_subscription->topic_name =
+    reinterpret_cast<const char *>(rmw_allocate(strlen(topic_name) + 1));
+  if (!rmw_subscription->topic_name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for subscription topic name");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(rmw_subscription->topic_name), topic_name, strlen(topic_name) + 1);
```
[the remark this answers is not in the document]
wjwwood: Actually, even better would be to use `rcutils_ret_t rcutils_ret = rcutils_format_string(allocator, "%s", topic_name)`.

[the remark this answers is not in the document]
wjwwood: Or `rcutils_strdup`: https://github.com/ros2/rcutils/blob/bd590e5253672957e8c7e070523fa725a4187db2/include/rcutils/strdup.h#L44

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/init_rmw_context_impl.hpp
```diff
@@ -0,0 +1,34 @@
...
+#ifndef RMW_FASTRTPS_CPP__INIT_RMW_CONTEXT_IMPL_HPP_
+#define RMW_FASTRTPS_CPP__INIT_RMW_CONTEXT_IMPL_HPP_
+
+#include "rmw/init.h"
+#include "rmw/types.h"
+
+namespace rmw_fastrtps_cpp
+{
+
+/// Increment `rmw_context_impl_t` reference count, initing it necessary.
```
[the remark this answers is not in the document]
ivanpauno: I will update using ` initializing it if necessary.`.
I find "zero initializing" not correct in this context.

rmw_fastrtps_cpp/src/subscription.cpp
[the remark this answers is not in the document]
ivanpauno: I think this is code I just moved, but I will update it.

rmw_fastrtps_cpp/src/rmw_init.cpp
ivanpauno: Yes, the problem is that I can't move it to `rmw_fastrtps_cpp_shared`, because it uses functions defined in `rmw_fastrtps_cpp`.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
ivanpauno: > Won't this interfere with copy elision?

Probably yes, as copy-assignment can't be elided AFAIK.
I will revert the change and left the not nice identing.

rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/init_rmw_context_impl.hpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_cpp/src/rmw_init.cpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_cpp/src/subscription.cpp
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/init_rmw_context_impl.hpp
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Function that should be called when creating a node,
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/init_rmw_context_impl.hpp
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/41977050b94d1e1498a35ba3f7c393048f0da8d5.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/e85759abf12d488f9a139ea3aab6efc66057e2b1.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/participant.hpp
ivanpauno: done

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/9fd7fe605ca749201371a4636274c2bfdb8ec1d0

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rmw_fastrtps/pull/312/commits/9fd7fe605ca749201371a4636274c2bfdb8ec1d0.

ivanpauno:
@hidmic I've addressed comments and rebased after https://github.com/ros2/rmw_fastrtps/pull/348.

ivanpauno:
Solving again conflicts after https://github.com/ros2/rmw_fastrtps/pull/350.

ivanpauno:
Full CI with all rmw implementation and security on (this will take pretty long):
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9482/)

I expect some cross vendor tests and security tests to fail.

ivanpauno:
> Full CI with all rmw implementation and security on (this will take pretty long):
> 
> * Linux [Build Status](https://ci.ros2.org/job/ci_linux/9482/)
> 
> I expect some cross vendor tests and security tests to fail.

Unfortunatelly, lots of cross-vendor tests between cyclonedds and fastrtps started failing.

I also tried a new build, with only `rmw_connext_cpp` and `rmw_fastrtps_cpp`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9486/)

ivanpauno:
Pinned to v1.9.3:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9500/)

ivanpauno:
Another one, to check if updates to test security worked:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9636/)

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -73,22 +73,18 @@ class ParticipantListener : public eprosima::fastrtps::ParticipantListener
       case eprosima::fastrtps::rtps::ParticipantDiscoveryInfo::DISCOVERED_PARTICIPANT:
         {
           auto map = rmw::impl::cpp::parse_key_value(info.info.m_userData);
-          auto name_found = map.find("name");
-          auto ns_found = map.find("namespace");
+          auto name_found = map.find("contextname");
```
hidmic: @ivanpauno nit: 
```suggestion
          auto name_found = map.find("context_name");
```

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
```diff
@@ -71,25 +64,19 @@ rmw_init_options_copy(
   const rcutils_allocator_t * allocator = &src->allocator;
   rmw_ret_t ret = RMW_RET_OK;
 
+  allocator->deallocate(dst->name, allocator->state);
   *dst = *src;
   dst->name = NULL;
-  dst->namespace_ = NULL;
   dst->security_options = rmw_get_zero_initialized_security_options();
 
   dst->name = rcutils_strdup(src->name, *allocator);
```
hidmic: @ivanpauno nit^2: consider attempting all allocations before actually writing `dst` so that an error leaves does not leave `dst` in a bad state.

rmw_fastrtps_shared_cpp/src/listener_thread.cpp
```diff
@@ -100,13 +100,14 @@ rmw_fastrtps_shared_cpp::join_listener_thread(rmw_context_t * context)
...
+#define TERMINATE_THREAD(msg) \
+  { \
     RCUTILS_SAFE_FWRITE_TO_STDERR( \
       RCUTILS_STRINGIFY(__FILE__) ":" RCUTILS_STRINGIFY(__function__) ":" \
-      RCUTILS_STRINGIFY(__LINE__) RCUTILS_STRINGIFY(msg) ": terminating ..."); \
-    std::terminate(); \
-  } while (0)
+      RCUTILS_STRINGIFY(__LINE__) RCUTILS_STRINGIFY(msg) \
+      ": ros discovery info listener thread will shutdown ...\n"); \
+    break; \
```
hidmic: @ivanpauno meta: this is certainly better than aborting, but I'm not thrilled with the idea of relying on a log to be seen for the user to do something about it i.e. client code will never know this happened.

ivanpauno:
Retriggering, after changes in https://github.com/ros2/sros2/pull/177:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9706/)

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
ivanpauno: I started using that. The function parsing the key doesn't accept an underscore in it :rofl::
https://github.com/ros2/rmw/blob/f76511333709a6f5d1e4543129ec73f2cd2b4f06/rmw/include/rmw/impl/cpp/key_value.hpp#L70

After I realized the root cause of the bug I was experimenting, I just dropped the underscore from the key name.

If you want, I can fix `parse_key`.

rmw_fastrtps_shared_cpp/src/listener_thread.cpp
ivanpauno: I think that's the trade-of, either you finish abruptly or you log and continue.
The main reason I prefer logging and continue, is that everything else can actually continue working correctly without this thread running. Some graph introspection methods will not return correct information, but everything else will continue working correctly.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
ivanpauno: I realized this wasn't  in sync with https://github.com/ros2/rmw_fastrtps/blob/f10cc19644cb1763d7b9db2d8cecc42e6f3a4051/rmw_fastrtps_shared_cpp/src/participant.cpp#L174

Fixed in https://github.com/ros2/rmw_fastrtps/pull/312/commits/f10cc19644cb1763d7b9db2d8cecc42e6f3a4051.

ivanpauno:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9735/)
- AArch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5369/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/7917/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/9610/)

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -0,0 +1,265 @@
...
+    participantAttrs.rtps.builtin.initialPeersList.push_back(local_network_interface_locator);
+  }
+
+  participantAttrs.rtps.builtin.domainId = static_cast<uint32_t>(domain_id);
+
+  size_t length = strlen(security_context) + strlen("securitycontext=;") + 1;
+  participantAttrs.rtps.userData.resize(length);
+  int written = snprintf(
+    reinterpret_cast<char *>(participantAttrs.rtps.userData.data()),
+    length, "securitycontext=%s;", security_context);
```
hidmic: @ivanpauno nit^2: calling `snprintf` with a `null` buffer of zero `length` will get you the required size, avoiding the need for clever computations.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -159,43 +121,39 @@ class ParticipantListener : public eprosima::fastrtps::ParticipantListener
...
+          rmw_fastrtps_shared_cpp::create_rmw_gid(
+            identifier_,
+            proxyData.guid()),
+          is_reader);
       }
     }
-    if (trigger) {
-      rmw_fastrtps_shared_cpp::__rmw_trigger_guard_condition(
-        graph_guard_condition_->implementation_identifier,
-        graph_guard_condition_);
```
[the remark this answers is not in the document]
ivanpauno: Replaced with this https://github.com/ros2/rmw_fastrtps/blob/a94234cc623d0f0d2c4321411029297be14f03a4/rmw_fastrtps_cpp/src/init_rmw_context_impl.cpp#L140-L159.

The purpose is to make the user know that a change happened in the graph.

rmw_fastrtps_shared_cpp/src/participant.cpp
ivanpauno: Nice trick, I checked cppreference to confirm https://en.cppreference.com/w/cpp/io/c/fprintf.
See https://github.com/ros2/rmw_fastrtps/pull/312/commits/f043aaa8eeec2471fa7a16e9dc3f143d77eca0ff.

ivanpauno: See also https://github.com/ros2/rmw_fastrtps/pull/312/commits/fa959e1d8c387e657b127e5ae93de5eac4202a31.

ivanpauno:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9794/)
- AArch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5416/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/7960/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/9654/)

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -0,0 +1,181 @@
...
+    eprosima_fastrtps_identifier, info->publisher_->getGuid());
+
+  rmw_publisher = rmw_publisher_allocate();
+  if (!rmw_publisher) {
+    RMW_SET_ERROR_MSG("failed to allocate publisher");
+    goto fail;
+  }
+  rmw_publisher->implementation_identifier = eprosima_fastrtps_identifier;
+  rmw_publisher->data = info;
+  rmw_publisher->topic_name = reinterpret_cast<char *>(rmw_allocate(strlen(topic_name) + 1));
```
wjwwood: Static cast instead?

ivanpauno: Done in https://github.com/ros2/rmw_fastrtps/pull/312/commits/0bf81122039925b1fac3690a0d0b4494d5734de4.

ivanpauno:
Yet another round of CI:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9809/)
- AArch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5436/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/7979/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/9674/)

ivanpauno:
Checking if I'm more lucky after https://github.com/ros2/rmw_fastrtps/pull/312/commits/47abe69f76aa53906fd33ef49312ae31f7bb5630:
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/7989/)

ivanpauno:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9879/)
- AArch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5498/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/8039/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/9738/)

ivanpauno:
- Focal [Build Status](https://ci.ros2.org/job/ci_linux/9889/)

ivanpauno:
@hidmic @wjwwood I will wait for the focal CI, but this looks like ready to be merged.
Complete list of PRs:

(with new fixes)
- https://github.com/ros2/rmw_fastrtps/pull/312
- https://github.com/ros2/rmw_dds_common/pull/4
- https://github.com/ros2/rmw/pull/189
- https://github.com/ros2/rmw_cyclonedds/pull/106
- https://github.com/ros2/rcl/pull/515
- https://github.com/ros2/system_tests/pull/406
- https://github.com/ros2/sros2/pull/177
- https://github.com/ros2/rclpy/pull/535

(rebased, no new changes)
- https://github.com/ros2/rmw_connext/pull/392

(already approved)
- https://github.com/ros2/rmw_implementation/pull/77

About the failures in the jobs, I see no regressions except some in `ros2param`, but those are just because the test is flaky. The test [creates a node](https://github.com/ros2/ros2cli/blob/02ae480e7aee8aea61803666ff5e10ac3a420ec5/ros2param/test/test_verb_dump.py#L52) and then run the cases, without any wait or retry on failure.
I've checked locally both on Linux and Windows, and the test passes.
@wjwwood @hidmic If you don't mind, I will make that test more reliable in a immediate follow-up.

PR #318, opened 2019-09-13
Correct linter failure

ivanpauno:
I guess this will solve this failure in the nightlies:
https://ci.ros2.org/view/nightly/job/nightly_win_rel/1310/testReport/junit/rmw_fastrtps_shared_cpp/cppcheck/error__syntaxError__src_rmw_wait_set_cpp_50_/

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_wait_set.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_wait_set.cpp
@@ -47,7 +47,8 @@ __rmw_create_wait_set(const char * identifier, rmw_context_t * context, size_t m
   wait_set->data = rmw_allocate(sizeof(CustomWaitsetInfo));
   // This should default-construct the fields of CustomWaitsetInfo
   wait_set_info = static_cast<CustomWaitsetInfo *>(wait_set->data);
-  RMW_TRY_PLACEMENT_NEW(wait_set_info, wait_set_info, goto fail, CustomWaitsetInfo)
+  // cppcheck-suppress syntaxError
+  RMW_TRY_PLACEMENT_NEW(wait_set_info, wait_set_info, goto fail, CustomWaitsetInfo, )
   if (!wait_set_info) {
     RMW_SET_ERROR_MSG("failed to construct wait set info struct");
     goto fail;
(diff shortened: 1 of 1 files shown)


ivanpauno:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/8133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4121/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6623/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8016/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8020/)

dirk-thomas:
We use this macro in numerous places already. Why is the suppression necessary for that line?

Also the CI builds still show warnings - do they need a new run?

ivanpauno:
> We use this macro in numerous places already. Why is the suppression necessary for that line?
> 
> Also the CI builds still show warnings - do they need a new run?

The warnings were with the first [commit](https://github.com/ros2/rmw_fastrtps/pull/318/commits/46930d8aaea9bfd5d43ccb794ffe86ef7f07dc54), where I deleted the comma at the end of the macro call.
After https://github.com/ros2/rmw_fastrtps/pull/318/commits/0f2babbeb01f37a93a994fa4adfa2b4c8d01860c, there is no more warning. We didn't have a warning before, only a linter failure on Windows, and now the diff with master is just the suppress comment.

There's no way of satisfy both linux/mac compiler and cppcheck on Windows.
I preferred to keep the old style (that doesn't generates a warning) and suppress the cppcheck failure than doing the opposite.

We could have another macro called `RMW_TRY_PLACEMENT_NEW_NO_ARGUMENTS`, but I think that's more confusing than adding a suppress line on top of each call with no arguments.

dirk-thomas:
> There's no way of satisfy both linux/mac compiler and Windows' cppcheck.

Maybe I am missing it but what makes this line different from all the other lines in the code base which use the macro with the same number of arguments and a trailing comma?

ivanpauno:
> Maybe I am missing it but what makes this line different from all the other lines in the code base which use the macro with the same number of arguments and a trailing comma (without needing the suppression)?

They will need it too, see https://ci.ros2.org/view/nightly/job/nightly_win_rel/1310/#showFailuresLink (probably something changed recently in `cppcheck`).
And I didn't find a CI job testing opensplice, but there are some calls with a trailing comma there too that will need a suppression.

dirk-thomas:
> probably something changed recently in `cppcheck`

It would good to know for certain and then also fill a ticket upstream about the regression.

> I didn't find a CI job testing opensplice

On ci.ros2.org there are nightly jobs which contain "extra_rmw" in their name which also build against OpenSplice. Beside that you can trigger any of CI builds and manually select OpenSplice.

And there are CI builds on build.ros2.org specifically for OpenSplice: http://build.ros2.org/view/Eci/job/Eci__nightly-opensplice_ubuntu_bionic_amd64/

ivanpauno:
> It would good to know for certain and then also fill a ticket upstream about the regression.

I will open a ticket when I get the version of cppcheck https://github.com/ros2/build_cop/issues/234#issuecomment-531387124.

PR #320, opened 2019-09-19
Correct error message

ivanpauno:
Follow-up of https://github.com/ros2/rmw_fastrtps/pull/311/files, after reading the error messages reported in https://github.com/ros2/build_cop/issues/241.


ivanpauno:
Sorry, something was not up to data when I first pushed, and the diff was wrong.
Now it's fine.

PR #322, opened 2019-09-21
Zero copy api

Karsten1987:
connected to https://github.com/ros2/ros2/issues/785


rmw_fastrtps_cpp/src/rmw_take.cpp
```diff
@@ -70,6 +70,52 @@ rmw_take_serialized_message_with_info(
...
+rmw_ret_t
+rmw_release_loaned_message(
+  const rmw_subscription_t * subscription,
+  void * loaned_message)
+{
+  (void) subscription;
+  (void) loaned_message;
+
+  RMW_SET_ERROR_MSG("rmw_release_loaned_message_with_info not implemented for rmw_fastrtps_cpp");
+  return RMW_RET_ERROR;
```
mjcarroll: There is also RMW_RET_UNSUPPORTED, would that make more sense here?

PR #331, opened 2019-10-11
Restrict traffic to localhost only if env var is provided

BMarchi:
PR to implement [network traffic restriction for localhost](https://github.com/ros2/ros2/issues/798). There are no test yet since this is a mock open for implementation discussion. Connected to [rmw](https://github.com/ros2/rmw/pull/190)

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8441/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4343/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8352/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/6866/)

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_node.cpp
@@ -37,7 +37,8 @@ rmw_create_node(
   const char * name,
   const char * namespace_,
   size_t domain_id,
-  const rmw_node_security_options_t * security_options)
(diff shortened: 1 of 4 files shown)


dirk-thomas:
How about passing the flag through the `create node` API instead? Similar to how the domain id is passed - rcl and above check the env var ans pass the value to the rmw interface.

PR #335, opened 2019-10-24
Load profiles based on topic names

EduPonz:
When loading configuration from a Fast DDS XML file, the current code uses the same publisher profile for all publishers and the same subscriber profile for all subscribers.

This PR aims to change this behavior, allowing the user to specify different settings on each publisher or subscriber, according to the following rules:

#### When creating a ROS2 publisher

If a `<publisher>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<publisher>` profile with `is_default_profile="true"`

#### When creating a ROS2 subscription

If a `<subscriber>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<subscriber>` profile with `is_default_profile="true"`

#### When creating a ROS2 service

For the creation of the request subscriber, if a `<subscriber>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, if a `<subscriber>` profile with `profile_name="service"` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<subscriber>` profile with `is_default_profile="true"`

For the creation of the reply publisher, if a `<publisher>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, if a `<publisher>` profile with `profile_name="service"` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<publisher>` profile with `is_default_profile="true"`

#### When creating a ROS2 client

For the creation of the request publisher, if a `<publisher>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, if a `<publisher>` profile with `profile_name="client"` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<publisher>` profile with `is_default_profile="true"`

For the creation of the reply subscriber, if a `<subscriber>` profile with `profile_name=topic_name` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, if a `<subscriber>` profile with `profile_name="client"` exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the `<subscriber>` profile with `is_default_profile="true"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,28 +28,25 @@ You can however set it to `rmw_fastrtps_dynamic_cpp` using the environment varia
 
 ## Advance usage
 
-`rmw_fastrtps` sets some of the Fast DDS configurable parameters:
+[`rclcpp`](https://github.com/ros2/rclcpp) and [`rclpy`](https://github.com/ros2/rclpy) only allow for the configuration of certain middleware QoS (see [ROS 2 QoS policies](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-policies)).
+In addition to ROS 2 QoS policies, `rmw_fastrtps` sets two more Fast DDS configurable parameters:
+
 * History memory policy: `PREALLOCATED_WITH_REALLOC_MEMORY_MODE`
 * Publication mode: `ASYNCHRONOUS_PUBLISH_MODE`
 
-However, it is possible to fully configure Fast DDS (including the history memory policy and the publication mode) using an XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
-Bear in mind that if you want to modify the history memory policy or the publication mode you must set environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` to 1 (it is set to 0 by default) besides defining the XML file.
-This tells `rmw_fastrtps` that it should override both the history memory policy and the publication mode using the XML.
-Bear in mind that if you set this environment variable but do not give a value to either of these policies, defaults will be used.
-Current Fast-DDS defaults are:
-* [History memory policy](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos): `PREALLOCATED_MEMORY_MODE`.
-* [Publication mode](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy): `SYNCHRONOUS_PUBLISH_MODE`.
+However, `rmw_fastrtps` offers the possibility to further configure Fast DDS:
 
-You have two ways of telling you ROS 2 application which XML to use:
-1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
-2. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to your XML file.
+* [Change publication mode](#change-publication-mode)
+* [Full QoS configuration](#full-qos-configuration)
 
 ### Change publication mode
 
-Another way to change easily the publication mode without the need of defining a XML file is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
-This variable has lower precedence than `RMW_FASTRTPS_USE_QOS_FROM_XML`.
-Therefore, it is only taken into account when `RMW_FASTRTPS_USE_QOS_FROM_XML` is not set (or given a value different than `1`).
+Fast DDS feats two different [publication modes](https://fast-dds.docs.eprosima.com/en/v2.1.0/fastdds/dds_layer/core/policy/eprosimaExtensions.html?highlight=synchronous#publishmodeqospolicykind): synchronous and asynchronous.
+To learn more about the implications of choosing one mode over the other, please refer to [DDS: Asynchronous vs Synchronous Publishing](https://www.eprosima.com/index.php/resources-all/performance/dds-asynchronous-vs-synchronous-publishing):
+
+`rmw_fastrtps` offers an easy way to change Fast DDS' publication mode without the need of defining a XML file. That is environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
 The admissible values are:
+
 * `ASYNCHRONOUS`: asynchronous publication mode.
 Setting this mode implies that when the publisher invokes the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
 A background thread (asynchronous thread) is in turn in charge of consuming the queue and sending the data to every matched reader.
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Please provide a detailed description what this patch is about and what the references in the title are referring to.

EduPonz:
Done

claireyywang:
@EduPonz looks like there are some build errors related to `rmw_fastrtps_shared_cpp` in the PR check. Can you please look into that? You may want to rebase onto `master` first to incorporate the latest changes. Thanks!

EduPonz:
@claireyywang I've taken a look at the console output from the CI and have found this error message.
```
02:47:01 In file included from /tmp/ws/src/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/rmw_node.cpp:49:0:
02:47:01 /opt/ros/eloquent/include/fastrtps/xmlparser/XMLProfileManager.h:96:37: note: candidate: static eprosima::fastrtps::xmlparser::XMLP_ret eprosima::fastrtps::xmlparser::XMLProfileManager::fillParticipantAttributes(const string&, eprosima::fastrtps::ParticipantAttributes&)
02:47:01          RTPS_DllAPI static XMLP_ret fillParticipantAttributes(
02:47:01                                      ^~~~~~~~~~~~~~~~~~~~~~~~~
02:47:01 /opt/ros/eloquent/include/fastrtps/xmlparser/XMLProfileManager.h:96:37: note:   candidate expects 2 arguments, 3 provided
```

In Fast-RTPS v1.9.3, we have added an optional argument to `XMLProfileManager::fillParticipantAttributes()`, `XMLProfileManager::fillPublisherAttributes()`, and `XMLProfileManager::fillSubscriberAttributes()` so that the user can prevent an error level log if the profile is not found in the XML. We have added this option in relation with this PR, because the intention is for the rmw to look for some default profiles, but we don't want to show a log if they are not found. Which version of Fast-RTPS are you using for this build?

Thanks!

claireyywang:
@EduPonz Thanks for the quick response! I believe we are using version `1.9.x` of Fast-RTPS for eloquent builds.

EduPonz:
@claireyywang I'm looking at the [header](https://github.com/eProsima/Fast-RTPS/blob/1.9.x/include/fastrtps/xmlparser/XMLProfileManager.h) and [source](https://github.com/eProsima/Fast-RTPS/blob/1.9.x/src/cpp/xmlparser/XMLProfileManager.cpp) files of `XMLProfileManager` in `1.9.x` and I can see the methods are correctly defined there (meaning including the `log_error` flag). Is it possible that your Fast-RTPS is not up-to-date or that you are linking with and older version?

claireyywang:
@ros-pull-request-builder retest this please.

EduPonz:
@claireyywang Actually, we now believe that the PR came before the update of the `1.9.x`, so the necessary changes were actually not there. Re-running the build should suffice in that case. can you do that or should I do a force push to force the build? Thanks!

claireyywang:
@ros-pull-request-builder retest this please

claireyywang:
@ros-pull-request-builder retest this please

claireyywang:
@ros-pull-request-builder retest this please

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
We can't run the PR builder atm since Eloquent has been branched of and F-turtle jobs aren't available yet. @claireyywang Please trigger jobs on ci.ros2.org instead.

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8716/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4548/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/7119/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8630/)

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -85,9 +89,39 @@ rmw_create_client(
...
+  // If FASTRTPS_DEFAULT_PROFILES_FILE defined, fill subscriber attributes with a subscriber
+  // profile profile located based of topic name defined by _create_topic_name(). If no profile is
+  // found, a search with profile_name "service"
+  // is attempted. Else, use the default attributes.
+  std::string topic_name_fallback = "client";
   eprosima::fastrtps::SubscriberAttributes subscriberParam;
+  eprosima::fastrtps::fixed_string<255> sub_topic_name = _create_topic_name(
+    qos_policies, ros_service_response_prefix, service_name, "Reply");
+  Domain::getDefaultSubscriberAttributes(subscriberParam);
+  if (std::getenv("FASTRTPS_DEFAULT_PROFILES_FILE") != nullptr) {
```
claireyywang: `std::getenv` triggered warnings on Windows. Compiler suggests using `_dupenv_s` instead. Same for below.

dirk-thomas: This should probably use the platform independent function `rcutils_get_env()`.

EduPonz:
I have removed the `std::getenv("FASTRTPS_DEFAULT_PROFILES_FILE")` called, since the `getDefault<>Attributes()` methods already attempt to load the profiles file defined by environment variable `FASTRTPS_DEFAULT_PROFILES_FILE`

claireyywang:
CI 
Linux  [Build Status](https://ci.ros2.org/job/ci_linux/8770/)
Linux-aarc64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4596/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/7170/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8694/)

ivanpauno:
Thanks for the reminder @claireyywang.

----

> If a <participant> profile with profile_name=node_name exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the <participant> profile with is_default_profile="true"

This rule will not have more sense when we finish with https://github.com/ros2/rmw/issues/183.
And I don't know what rule would be used after that change.

----

In case of publishers/subscribers, the ros topic name is being used.
In case of clients/services, the demangled topic name is being used.
Maybe just using the ros client/service name in the last case would be more clearer.
Though that would enforce to use the same profile for the two entities a client/service create (which might be fine).

----

About the other rules, though it sounds reasonable to load a profile based on the topic name, I'm not sure if that will satisfy every use case. I think that it would be better if we could pass the profile file and the profile name in a higher level API, than having some magic rules in the implementation.

As exposing DDS specific API is usually not desired, I think we should provide a mechanism to pass rmw specific options in the creation of a publisher/subscriber/node/context, that could have this information. Maybe, just extending the idea @wjwwood applied [here](https://github.com/ros2/rclcpp/pull/882). I think we need a way to extend the higher level API, in an 
(rmw) implementation specific way (see also, https://discourse.ros.org/t/a-more-unified-and-standard-way-of-configuring-the-dds-layer/11372).

EduPonz:
Hi all!

I have corrected the problematic line lengths.

Maybe the rule will not apply anymore after ros2/rmw#183, but it will be useful now in eloquent, for example for the iRobot performance test case that initiates the ROS discourse you linked.

This feature is motivated for the use-case where a lot of nodes are spin (such as iRobot case of any other scalability tests for example), which makes the Fast-RTPS configuration using XML just not convenient (because of the need of many XML files). With the change, there could be just one XML, which profiles are selected dynamically depending on the node and topic name. Furthermore, it is not incompatible with extending the higher level API, since this only applies if the XML file is present.

As for the client/service topic names, the demangled ones are used precisely to not enforcing the use  of the same profile for different entities.

ivanpauno:
> Maybe the rule will not apply anymore after ros2/rmw#183, but it will be useful now in eloquent, for example for the iRobot performance test case that initiates the ROS discourse you linked.

If it's merged now, it will be useful in `Foxy` and for people using `master`.
Features aren't usually backported.
And merging a feature that will be broken in the near feature, without a plan for working it around, it's not a good idea (a name may be added to the context as part of `ros2/rmw#183` to allow configuring the security options, it could be used here too in that case).

IMHO, we need a better approach to configure middleware specific options from the higher level API, and not just applying some magic rules under the hood.

@dirk-thomas @wjwwood @claireyywang @hidmic opinions?

dirk-thomas:
> IMHO, we need a better approach to configure middleware specific options from the higher level API

We all agree on that I think.

Since it is unclear when that becomes available I think it is reasonable to use RMW specific configuration options until then. It allows the user to configure what they need right now with the trade off that it locks them to one vendor and the general approach might be superseded in the future.

ivanpauno:
>>If a profile with profile_name=node_name exists in FASTRTPS_DEFAULT_PROFILES_FILE, then load it.
Otherwise, use the profile with is_default_profile="true"
>
> This rule will not have more sense when we finish with ros2/rmw#183.

How to extend the rule after ros2/rmw#183 gets merged is still unclear.

> We all agree on that I think.
>
> Since it is unclear when that becomes available I think it is reasonable to use RMW specific configuration options until then. It allows the user to configure what they need right now with the trade off that it locks them to one vendor and the general approach might be superseded in the future.

If these environments variables aren't considered part of the API, and they can be broken without any deprecation cycle, I don't care if this is merged as-is. But I don't see the point of merging and breaking something, except if we want to backport the feature.

EduPonz:
Hi all, and again thanks for taking the time to discuss this!

First of all, I have re-based with master to get the README and have updated the PR documentation there.

For sure the user will benefit from a unified way of configuring. However, not all the vendors expose the same configuration features, and I don't see a reason for limiting user options if the middleware has them.

> If these environments variables aren't considered part of the API, and they can be broken without any deprecation cycle, I don't care if this is merged as-is. But I don't see the point of merging and breaking something, except if we want to backport the feature.

Regarding the environment variables `FASTRTPS_DEFAULT_PROFILES_FILE` is used by Fast-RTPS and will not be deprecated, and `RMW_FASTRTPS_USE_QOS_FROM_XML` could, in my opinion, stay after ros2/rmw#183. I mean, even if there was a way of configuring from upper layers, this could still be in place.

As for the plan for when there is only one participant per context, the rule should just change to load participant profile using context name, which as I understand will exist (as a requirement for security).

We are currently using this PR for the case of a ROS 2 application with a known topology, where we know beforehand which nodes (participants) will communicate to which. In this scenario we set (using XML), for each participant, a white-list of participants, so they only discover the ones they are interested in. This, for instance, cannot be done in eloquent, which carries a serious impact in performance when dealing with big topologies with a lot of nodes that are not necessarily interested in each other.

agoeckner:
I have a situation where using static EDP is desirable for performance. However, static EDP requires that each publisher/subscriber have a unique "user defined ID" and "entity ID".

Am I correct in thinking that this would allow me to set those IDs for any chosen ROS2 topic?

dirk-thomas:
With the goal of merging the feature described in ros2/rmw#183 soon (definitely for the next release Foxy) the currently proposed patch will not work anymore since multiple nodes share the participant.

@EduPonz So the question is how should this proposal work in the context of the DDS participant being associated to the context (rather than the node)?

EduPonz:
@agoeckner Yes and no. You can set "user defined ID" and "entity ID" in Fast-RPTS XML profiles ([publisher profile example](https://eprosima-fast-rtps.readthedocs.io/en/latest/xmlprofiles.html#publisher-profiles)). There, you can name the profile using your topic name, and the profile will be loaded. As long as you have one publisher/subscriber, per topic, you will have unique "user defined ID" and "entity ID". However, if you create more than one publisher/subscriber in a given topic, they will all load the same profile.

@dirk-thomas In my opinion, the participant profile should be loaded using the [context name](https://github.com/ros2/design/blame/72ad69028917a391715f40d644c9c85f9ada337b/articles/151_node_to_participant_mapping.md#L190) (rather than the node's)

EduPonz:
I have updated this PR cause I think the content is very much relevant and users would definitely benefit from it. It enables loading different QoS settings for different publishers/subscriptions/clients/services based on topic names. Participant profiles are not loaded since, to my knowledge, there is no context identifier that the user can be aware of.

MiguelCompany:
> Participant profiles are not loaded since, to my knowledge, there is no context identifier that the user can be aware of.

Perhaps you should update the PR description, then

EduPonz:
> Perhaps you should update the PR description, then

I had done that, but I never clicked on update. Thanks!

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -161,8 +166,20 @@ rmw_create_client(
...
+  // with profile_name "client" is attempted. Else, use the default attributes.
+  std::string topic_name_fallback = "client";
   eprosima::fastrtps::SubscriberAttributes subscriberParam;
-  eprosima::fastrtps::PublisherAttributes publisherParam;
+  eprosima::fastrtps::fixed_string<255> sub_topic_name = _create_topic_name(
+    qos_policies, ros_service_response_prefix, service_name, "Reply");
+  Domain::getDefaultSubscriberAttributes(subscriberParam);
+
+  if (XMLProfileManager::fillSubscriberAttributes(
+      sub_topic_name.to_string(), subscriberParam, false) != XMLP_ret::XML_OK)
```
ivanpauno, line 179: is there a specific return code for "profile not found"?
If there is, it would be good to check in all calls to `fillSubscriberAttributes()` that the return code is either "ok" or "profile not found" and not a different error.

ivanpauno (approved):
@EduPonz could you add documentation about this in the README?

ivanpauno:
It would also be great if we would have ROS2 documentation pointing out when it's recommended to use the new "qos parameter overrides" and when vendor specific profiles. The general guideline would be use parameters for qos settings available in ROS 2, and to use vendor specific files for everything else.

We actually should add a ROS 2 tutorial about that, but if you can briefly mention both in the readme that would be great.

> Participant profiles are not loaded since, to my knowledge, there is no context identifier that the user can be aware of.

Yeah, we don't have an identifier for that.

Most applications only have one participant per process.
If you want to share the same XML file between different processes but use different participant profiles for each of them, maybe the profile to use could be specified in an environment variable (?), but I'm not sure if that's a good idea or just using different files is better.

rmw_fastrtps_cpp/src/rmw_client.cpp
EduPonz: `fill*Attributes()` consists in a find in a map. The method returns `XMLP_ret::XML_OK` if found, and `XMLP_ret::XML_ERROR` if not. In that sense, `XMLP_ret::XML_ERROR` is the same as profile not found

EduPonz:
> If you want to share the same XML file between different processes but use different participant profiles for each of them, maybe the profile to use could be specified in an environment variable (?), but I'm not sure if that's a good idea or just using different files is better.

We can discuss that for a different PR. I think that we have enough interdependent environment variable as it is, adding more might just be too confusing. I'll extend the README and then we can talk about how to proceed with the ROS 2 tutorial. Thanks for the review!

rmw_fastrtps_cpp/src/rmw_client.cpp
ivanpauno: sounds good

EduPonz:
@ivanpauno I've extended the README to document the new functionality. It'd be great if you could take a look at it.

README.md
```diff
@@ -59,9 +56,65 @@ This entails that any blocking call occurring during the write operation would b
...
+1. To configure the reply publisher, define a `<publisher>` profile with attribute `profile_name=topic_name`, where topic name is the name of the service after mangling.
+If such profile is not defined, `rmw_fastrtps` attempts to load a `<publisher>` profile with attribute `profile_name="service"`.
+If neither of the previous profiles exist, `rmw_fastrtps` attempts to load the `<publisher>` profile with attribute `is_default_profile="true"`.
+
+##### Creating clients with different profiles
+
+ROS 2 clients contain a publisher to send requests, and a subscription to receive the service's replies.
+`rmw_fastrtps` allows for configuring each of these endpoints separately in the following manner:
+
+1. To configure the requests publisher, define a `<publisher>` profile with attribute `profile_name=topic_name`, where topic name is the name of the service after mangling.
```
ivanpauno, line 110: it would be great to explain how the topic names after mangling look

README.md
```diff
@@ -59,9 +56,65 @@ This entails that any blocking call occurring during the write operation would b
...
+1. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to contain the path to your XML file (relative to the working directory).
+
+#### Applying different profiles to different entities
+
+`rmw_fastrtps` allows for the configuration of different entities with different QoS using the same XML file.
+For doing so, `rmw_fastrtps` locates profiles in the XML based on topic names abiding to the following rules:
+
+##### Creating publishers/subscriptions with different profiles
+
+To configure a publisher/subscription, define a `<publisher>`/`<subscriber>` profile with attribute `profile_name=topic_name`.
```
ivanpauno, line 90: it would be great to clarify if this is the topic name after mangling or not

ivanpauno (approved):
Maybe an example of how to use different profiles for different topics would be good.

This looks great to me, thanks!

ivanpauno:
Waiting for an extra approval from a maintainer, maybe @hidmic or @mjeronimo ?

JaimeMartin (approved):
LGTM

README.md
hidmic: +1, but since some of the mangling is ROS 2 specific, let's not repeat existing documentation.

hidmic (approved):
Overall LGTM. On a related note, why is this functionality not available in `rmw_fastrtps_dynamic_cpp`?

hidmic:
This needs a CI run too.

hidmic:
CI up to `rmw_fastrps_cpp`, `test_rmw_implementation`, `test_communication`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13307/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8240/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11040/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13350/)

EduPonz:
@hidmic @ivanpauno I'm adding the functionality to `rmw_fastrtps_dynamic_cpp` and extending the README in a follow up PR.

PR #340, opened 2019-12-04
Create README.md

EduPonz:
Signed-off-by: EduPonz <eduardoponz@eprosima.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,40 +1,51 @@
 # ROS2 Middleware Implementation for eProsima's Fast-RTPS
+
 `rmw_fastrtps` constitutes __[ROS2](https://index.ros.org/doc/ros2/) default middleware implementation__, providing an interface between ROS2 and [eProsima's](https://www.eprosima.com/index.php) [Fast-RTPS](https://github.com/eProsima/Fast-RTPS) middleware.
 
 ## Getting started
-This implementation is available in all ROS2 distributions, both from binaries and from sources. You do not need to do anything in order to use Fast-RTPS as your ROS2 middleware layer (since it is the default implementation). However, you can still specify it in two different ways:
+This implementation is available in all ROS2 distributions, both from binaries and from sources.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please start each sentence on a new line (following the [styleguide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#markdown-restructured-text-docblocks)).

EduPonz:
Thanks @dirk-thomas! Did not know about that

README.md
```diff
@@ -0,0 +1,80 @@
...
+1. When launching your ROS2 application:
+    ```bash
+    RMW_IMPLEMENTATION=rmw_fastrtps_cpp ros2 run <your_package> <your application>
+    ```
+
+## Two different RMW implementations
+
+`rmw_fastrtps` actually provides not one but two different ROS2 middleware implementations, both of them using Fast-RTPS as middleware layer: `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` (note that directory `rmw_fastrtps_shared_cpp` just contains the code that the two implementations share, and does not constitute a layer on its own).
+
+The main difference between the two is that `rmw_fastrtps_dynamic_cpp` uses ROS2 Introspection at run time to decide on the serialization/deserialization mechanism, meaning that the application can be agnostic of the message types.
```
dirk-thomas: Instead of:

> ... ROS2 Introspection ...

I would suggest:

> ... introspection typesupport ...

The "ROS2" seems unnecessary here and the typesupport is the important missing term.

README.md
```diff
@@ -0,0 +1,80 @@
...
+On the other hand, `rmw_fastrtps_cpp` relies on message types known at build time.
```
dirk-thomas: Even the introspection typesupport requires the message types to be known at build time. So I think this should be rephrased to something like: "rmw_fastrtps_cpp uses its own typesupport which generated the mapping for each message type at build time".

EduPonz:
I have applied the suggested changes

dirk-thomas:
Thanks for this contribution.

PR #342, opened 2019-12-20
Passing down type support information

### LINKED ISSUE #265: Using different typesupports on same node and topic ends in segfault
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - Head
- DDS implementation:
  - Fast RTPS
  - N/A

#### Steps to reproduce issue
Create a node that registers to an already registered topic but with a different typesupport. If we assume that the node is using rosout log, and that rosout regsiters a publisher in the /rosout topic with C typesupport. You can subscribe to the /rosout topic with a CPP typesupport.

```
#include <iostream>
#include "rclcpp/rclcpp.hpp"
#include <rcl_interfaces/msg/log.hpp>

class MinimalSubscriber : public rclcpp::Node
{
public:
  MinimalSubscriber()
  : Node("minimal_subscriber")
  {
    subscription_ = this->create_subscription<rcl_interfaces::msg::Log>(
      "rosout",
      [this](rcl_interfaces::msg::Log::UniquePtr msg) {
      RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg->msg.c_str());
    });
  }

private:
  rclcpp::Subscription<rcl_interfaces::msg::Log>::SharedPtr subscription_;
};

int main(int argc, char * argv[])
{
  rclcpp::init(argc, argv);
  rclcpp::spin(std::make_shared<MinimalSubscriber>());
  rclcpp::shutdown();
  return 0;
}

```

#### Expected behavior
Both registered publisher and subscribers should be able to publish and receive messages for the topic.

#### Actual behavior
The process exits with a segfault when receiving a message on the subscribed topic.

#### Additional information
Two different typesupports cannot be in the same topic at the same time.
This is due to the fact that rmw_fastrtps uses the same name to register both typesupports in the FastRTPS Domain. See [code](https://github.com/ros2/rmw_fastrtps/blob/c9e496560d098b0de2e12aee9af31b1b990156ce/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L108)

Disambiguate with a different name (appending or prepending a string) is not possible because the name of the typesupport is directly related to the name of the topic. See [code](https://github.com/ros2/rmw_fastrtps/blob/c9e496560d098b0de2e12aee9af31b1b990156ce/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L108) and [fastrtps TopicData](https://github.com/eProsima/Fast-RTPS/blob/bfbe04e0974758fab68c377800b5fbf2741949b9/include/fastrtps/TopicDataType.h#L87)

Probably happening in other implementations.

MiguelCompany:
This PR fixes #265.

The idea is to use a mechanism similar to the one used for already serialized types, passing an additional field on the `SerializedData` structure with a pointer to the corresponding rosidl_typesupport_fastrtps entity.

Tested with a modified version of [this example](https://github.com/ros2/examples/blob/master/rclcpp/minimal_subscriber/lambda.cpp) that subscribes to `rosout`


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -41,6 +41,7 @@ typedef struct CustomSubscriberInfo : public CustomEventInfo
   eprosima::fastrtps::Subscriber * subscriber_;
   SubListener * listener_;
   rmw_fastrtps_shared_cpp::TypeSupport * type_support_;
+  const void * type_support_impl_;
```
hidmic: @MiguelCompany here and elsewhere, out of curiosity, why not storing a pointer to the type support struct instead of those of its individual members?

MiguelCompany: @hidmic Avoiding an indirection every time we publish or take improves performance, IMHO

hidmic: Alright, that sounds reasonable.

dirk-thomas:
Will this fix be backported to the 1.9.x branch?

hidmic:
CI up to `rmw_fastrtps` and `test_communication`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8988/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4755/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7358/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8919/)

MiguelCompany:
> Will this fix be backported to the 1.9.x branch?

No changes have been required on Fast RTPS, just on this repo, so no need to backport.

MiguelCompany:
@hidmic I solved the issues with the linters. Would you mind launching CI once more?

hidmic:
Alright, CI's green. Moving forward. Thanks for your contribution @MiguelCompany !

dirk-thomas:
@hidmic Should this be backported to Eloquent and/or Dashing? (@mjcarroll FYI)

hidmic:
@dirk-thomas Yes, it probably should, but let me investigate CI failures on https://github.com/ros2/ros1_bridge/pull/233 first.

dirk-thomas:
I added this to both projects to track the backporting.

PR #343, opened 2020-01-13
Do not force asynchronous publishing

MiguelCompany:
Since eProsima/Fast-RTPS#898, asynchronous publishing is not necessary by default. This PR lets the default value for the publishing mode (which is synchronous), which usually has better latency results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,11 @@ You can however set it to `rmw_fastrtps_dynamic_cpp` using the environment varia
 
 ## Advance usage
 
-`rmw_fastrtps` sets some of the Fast-RTPS configurable parameters:
-* History memory policy: `PREALLOCATED_WITH_REALLOC_MEMORY_MODE`
-* Publication mode: `ASYNCHRONOUS_PUBLISH_MODE`
+`rmw_fastrtps` sets Fast-RTPS history memory policy to `PREALLOCATED_WITH_REALLOC_MEMORY_MODE`
 
-However, it is possible to fully configure Fast-RTPS (including the history memory policy and the publication mode) using an XML file as described in [Fast-RTPS documentation](https://eprosima-fast-rtps.readthedocs.io/en/latest/xmlprofiles.html).
+However, it is possible to fully configure Fast-RTPS (including the history memory policy) using an XML file as described in [Fast-RTPS documentation](https://eprosima-fast-rtps.readthedocs.io/en/latest/xmlprofiles.html).
 Then, you just need to set environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` to 1 (it is set to 0 by default).
-This tells `rmw_fastrtps` that it should not override neither the history memory policy nor the publication mode.
+This tells `rmw_fastrtps` that it should not override the history memory policy.
 
 You have two ways of telling you ROS 2 application which XML to use:
 1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -28,21 +28,19 @@ You can however set it to `rmw_fastrtps_dynamic_cpp` using the environment varia
 
 ## Advance usage
 
-`rmw_fastrtps` sets some of the Fast-RTPS configurable parameters:
-* History memory policy: `PREALLOCATED_WITH_REALLOC_MEMORY_MODE`
-* Publication mode: `ASYNCHRONOUS_PUBLISH_MODE`
```
ivanpauno: Before, asynchronous publish mode was forced, and the readme documented how to configure fastrtps in synchronous publish mode.
Does make sense now to include in the readme how to configure fastrtps in asynchronous mode?
Just wondering, I don't know in which cases the user would be interested in doing that.

MiguelCompany: We are only documenting how to avoid the parameters that are overridden by the code in this repo. Configuring fastrtps for asynchronous mode will now be done as with the rest of the parameters, i.e. with the XML file.

Now that I read it in detail, I think the whole section may need some rework, but the changes will not be related to this PR, so I propose to make them on a separate one.

ivanpauno: Sounds reasonable

MiguelCompany:
> Can you apply the same three changes to `rmw_fastrpts_dynamic_cpp` (`rmw_publisher`, `rmw_client`, `rmw_service`)?

ups! done!

ivanpauno:
CI including test_rclcpp, test_communication:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8995/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4763/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8931/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/7364/)

JaimeMartin:
@ivanpauno is ok with you if we merge this? We need this change for the ongoing performance tests.

ivanpauno:
> @ivanpauno is ok with you if we merge this? We need this change for the ongoing performance tests.

I haven't merged yet, because I'm not sure of all the implications this PR has.
About performance tests, I think the order is the opposite. First, have a performance test framework that can measure performance of what we currently have. Then do performance improvements, and use the performance tests as a proof. Without performance tests, I have to blindy trust that performance does improve with this PR.

----

I still think the changes are reasonable. Adding @wjwwood for a second approval before merging.

JaimeMartin:
The performance improves, yes. We have your performance tests replicated on our performance test machines. You can safely merge this.

wjwwood:
Please do not merge this until I have time to review it. 

I think this may be changing behavior in a significant way and is also causing the behavior of publish to be different from other rmw implementations. 

Also I think that allowing the publishing mode to be changed via external xml is a very bad idea as I detailed in this discussion:

https://github.com/ros2/ros2/issues/255

tfoote:
I agree that we should stick with asynchronous across all implementations for consistency between implementations as well as because it's what's expected of the ROS API. We were actually inconsistent between python and c++ originally in ROS 1 and converted python to asynchronous several years ago too.

From the comment on the Opensplice ticket it appears that it's also asynchronous and doesn't have support for synchronous publishing https://github.com/ros2/rmw_opensplice/issues/301#issuecomment-577076387

MiguelCompany:
Closing this, as we already provided a way of avoiding async publishing on #470

PR #346, opened 2020-01-31
Fix build warning in Ubuntu Focal

ivanpauno:
Fixes https://ci.ros2.org/job/ci_linux/9175/warnings23Result/package.936823267/.

```
‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘struct eprosima::fastrtps::rtps::GUID_t’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
--
```

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
@@ -23,10 +23,12 @@
 #include "rmw/topic_endpoint_info_array.h"
 #include "rmw/topic_endpoint_info.h"
 
-#include "demangle.hpp"
 #include "rmw_fastrtps_shared_cpp/custom_participant_info.hpp"
+#include "rmw_fastrtps_shared_cpp/guid_utils.hpp"
 #include "rmw_fastrtps_shared_cpp/namespace_prefix.hpp"
 #include "rmw_fastrtps_shared_cpp/rmw_common.hpp"
+
+#include "demangle.hpp"
 
 namespace rmw_fastrtps_shared_cpp
 {
@@ -144,8 +146,10 @@ _set_rmw_topic_endpoint_info(
   }
   // set endpoint gid
   uint8_t rmw_gid[RMW_GID_STORAGE_SIZE];
-  memset(&rmw_gid, 0, RMW_GID_STORAGE_SIZE);
-  memcpy(&rmw_gid, &topic_data.entity_guid, sizeof(eprosima::fastrtps::rtps::GUID_t));
+  // memset(&rmw_gid, 0, RMW_GID_STORAGE_SIZE);
+  rmw_fastrtps_shared_cpp::copy_from_GUID_t_to_byte_array(
+    topic_data.entity_guid,
+    rmw_gid);
   ret = rmw_topic_endpoint_info_set_gid(
     topic_endpoint_info,
     rmw_gid,
--- a/rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
@@ -146,7 +146,7 @@ _set_rmw_topic_endpoint_info(
   }
   // set endpoint gid
   uint8_t rmw_gid[RMW_GID_STORAGE_SIZE];
-  rmw_fastrtps_shared_cpp::copy_from_GUID_t_to_byte_array(
+  rmw_fastrtps_shared_cpp::copy_from_fastrtps_guid_to_byte_array(
     topic_data.entity_guid,
     rmw_gid);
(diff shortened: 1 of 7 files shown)


rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -92,7 +92,9 @@ __rmw_send_response(
   assert(info);
 
   eprosima::fastrtps::rtps::WriteParams wparams;
-  memcpy(&wparams.related_sample_identity().writer_guid(), request_header->writer_guid,
+  memcpy(
+    static_cast<void *>(&wparams.related_sample_identity().writer_guid()),
+    request_header->writer_guid,
```
hidmic: @ivanpauno it seems to me that these structures have a copy assignment operator, why not using that instead?

hidmic: Hmm, after a quick sync with @ivanpauno, this is just another case of a `memcpy` from a primitive array into a `struct` (!?). It's interesting the compiler didn't complain before. Why is this being done across `rmw_fastrtps` instead of defining an adequate copy-assignment operator for `GUID_t`? @richiware @MiguelCompany by any chance, do you happen to know/recall?

ivanpauno: I solved the problem in https://github.com/ros2/rmw_fastrtps/pull/346/commits/5ebac808ea5e89e7bc8e174e267eb79014fec88f, instead of ignoring the warning as before.

rmw_fastrtps_shared_cpp/src/rmw_get_topic_endpoint_info.cpp
```diff
@@ -144,8 +146,10 @@ _set_rmw_topic_endpoint_info(
   }
   // set endpoint gid
   uint8_t rmw_gid[RMW_GID_STORAGE_SIZE];
-  memset(&rmw_gid, 0, RMW_GID_STORAGE_SIZE);
-  memcpy(&rmw_gid, &topic_data.entity_guid, sizeof(eprosima::fastrtps::rtps::GUID_t));
+  // memset(&rmw_gid, 0, RMW_GID_STORAGE_SIZE);
```
jacobperron: Should this be uncommented or removed?

ivanpauno:
* Focal [Build Status](http://ci.ros2.org/job/ci_linux/9454/)
* Bionic [Build Status](http://ci.ros2.org/job/ci_linux/9455/)
* AArch64 Bionic [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5172/)
* OsX [Build Status](http://ci.ros2.org/job/ci_osx/7733/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9385/)

ivanpauno:
* Focal [Build Status](http://ci.ros2.org/job/ci_linux/9469/)
* Bionic [Build Status](http://ci.ros2.org/job/ci_linux/9470/)
* AArch64 Bionic [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5199/)
* OsX [Build Status](http://ci.ros2.org/job/ci_osx/7743/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9396/)

ivanpauno:
@hidmic @jacobperron I was copying in the opposite direction I should [here](https://github.com/ros2/rmw_fastrtps/pull/346/commits/581f80f69ca8b039ee243ea36fdfb3cb44bfc54a#diff-03f98ff242b054bf5cd49a872185b58d).
Now tests should pass.

hidmic:
Let's not pay attention to us two approving as it was :zipper_mouth_face:

ivanpauno:
> There are some OSX failures.

I'm surprised that clang did not allow that usage of `constexpr`, it seems to be a bug in the compiler to me. Another try:

* OsX [Build Status](http://ci.ros2.org/job/ci_osx/7744/)

MiguelCompany:
@ivanpauno 
> I'm surprised that clang did not allow that usage of `constexpr`, it seems to be a bug in the compiler to me. 

The compiler complains when accessing the argument `guid` (which is not a `constexpr`) to construct another `constexpr`, so I don't see it as a bug. You could use `eprosima::fastrtps::rtps::GuidPrefix_t::size`

ivanpauno:
> The compiler complains when accessing the argument guid (which is not a constexpr) to construct another constexpr, so I don't see it as a bug.

I'm not sure what the standard says, but other compilers can solve it (and it's definitly solvable).
For example, `sizeof(guid.guidPrefix.value)` evaluates the accessing of an argument in compile time. In the case of `guid.guidPrefix.size`, being the last argument a `static constexpr`, the compiler should be able to solve it at compile time (I don't know if it's a requirement of the standard or not, but some compilers can).

> You could use eprosima::fastrtps::rtps::GuidPrefix_t::size. Or even better decltype(guid->guidPrefix)::size

I prefer the `sizeof` version, but those are fine too.

PR #350, opened 2020-02-24
Fixing type support C/CPP mix on rmw_fastrtps_dynamic_cpp

### LINKED ISSUE #265: Using different typesupports on same node and topic ends in segfault

MiguelCompany:
This should fix #265 with `rmw_fastrtps_dynamic_cpp`.

### DIFF
--- a/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
@@ -135,8 +135,7 @@ void TypeSupportRegistry::return_response_type_support(
 
 static TypeSupportRegistry type_registry_instance;
 
-TypeSupportRegistry & get_type_support_registry(const rmw_node_t * node)
+TypeSupportRegistry & get_type_support_registry()
 {
-  (void)node;
   return type_registry_instance;
 }
--- a/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rcutils/logging_macros.h"
+
 #include "rmw/error_handling.h"
 
 #include "type_support_common.hpp"
@@ -41,15 +43,31 @@ void return_type_support(
   auto it = map().find(ros_type_support);
   assert(it != map().end());
   if (0 == --it->second.ref_count) {
+    delete it->second.type_support;
     map().erase(it);
+  }
+}
+
+template<typename map_type>
+void cleanup(map_type & map, const char * msg)
+{
+  std::lock_guard<std::mutex> guard(map.getMutex());
+  if (!map().empty()) {
+    RCUTILS_LOG_DEBUG_NAMED(
+      "rmw_fastrtps_dynamic_cpp",
+      "TypeSupportRegistry %s is not empty. Cleaning it up...", msg);
+    for (auto it : map() ) {
+      delete it.second.type_support;
+    }
+    map().clear();
   }
 }
 
 TypeSupportRegistry::~TypeSupportRegistry()
 {
-  assert(message_types_().empty());
-  assert(request_types_().empty());
-  assert(response_types_().empty());
+  cleanup(message_types_, "message_types_");
+  cleanup(request_types_, "request_types_");
+  cleanup(response_types_, "response_types_");
 }
 
 type_support_ptr TypeSupportRegistry::get_message_type_support(
--- a/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
@@ -70,6 +70,12 @@ TypeSupportRegistry::~TypeSupportRegistry()
   cleanup(response_types_, "response_types_");
 }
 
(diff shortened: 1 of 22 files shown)


MiguelCompany:
@dirk-thomas This is ready for review!

dirk-thomas:
Packaging job using this patch as well as ros2/ros1_bridge#233: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/298/)

dirk-thomas:
Non-packaging CI builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9437/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5124/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9363/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/209/)

MiguelCompany:
Will address failures today

MiguelCompany:
@dirk-thomas I think I have solved the issue.

I was assuming that the entities were always being destroyed, and had an assertion to check all the objects registered on the type registry were unregistered before destroying the registry itself.

But this is not the case on the failing tests, so I have changed the mechanism on the type registry destructor from an assertion to a cleanup.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9457/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5187/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7734/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9386/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/227/)

rmw_fastrtps_dynamic_cpp/src/type_support_registry.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+class TypeSupportRegistry
+{
+private:
+  LockedObject<msg_map_t> message_types_;
+  LockedObject<srv_map_t> request_types_;
+  LockedObject<srv_map_t> response_types_;
+
+public:
+  ~TypeSupportRegistry();
```
hidmic: @MiguelCompany nit: consider making the (default) constructor private.

rmw_fastrtps_dynamic_cpp/src/type_support_registry.cpp
```diff
@@ -0,0 +1,159 @@
...
+  return_type_support(ros_type_support, request_types_);
+}
+
+void TypeSupportRegistry::return_response_type_support(
+  const rosidl_service_type_support_t * ros_type_support)
+{
+  return_type_support(ros_type_support, response_types_);
+}
+
+static TypeSupportRegistry type_registry_instance;
```
hidmic: @MiguelCompany consider moving this into the function below to avoid having a non trivially constructed object in global scope.

MiguelCompany:
@hidmic I have just addressed your review comments

PR #354, opened 2020-03-17
Add rmw_*_event_init() functions

mm318:
Related to https://github.com/ros2/ros2/issues/822

This adds the `rmw_publisher_event_init()` and `rmw_subscription_event_init()` functions to `rmw_fastrtps`, moved from `rmw`, so that these functions will be able to check for whether or not the given type of `rmw_event_type_t` is supported by the chosen middleware.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_event.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_event.cpp
@@ -26,7 +26,9 @@ static const std::unordered_set<rmw_event_type_t> g_rmw_event_type_set{
   RMW_EVENT_OFFERED_DEADLINE_MISSED
 };
 
-namespace rmw_fastrtps_shared_cpp_internal
+namespace rmw_fastrtps_shared_cpp
+{
+namespace internal
 {
 
 bool is_event_supported(rmw_event_type_t event_type)
@@ -34,10 +36,7 @@ bool is_event_supported(rmw_event_type_t event_type)
   return g_rmw_event_type_set.count(event_type) == 1;
 }
 
-}  // namespace rmw_fastrtps_shared_cpp_internal
-
-namespace rmw_fastrtps_shared_cpp
-{
+}  // namespace internal
 
 rmw_ret_t
 __rmw_init_event(
@@ -56,7 +55,7 @@ __rmw_init_event(
     topic_endpoint_impl_identifier,
     identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
-  if (!rmw_fastrtps_shared_cpp_internal::is_event_supported(event_type)) {
+  if (!internal::is_event_supported(event_type)) {
     RMW_SET_ERROR_MSG("provided event_type is not supported by rmw_fastrtps_cpp");
     return RMW_RET_UNSUPPORTED;
   }
--- a/rmw_fastrtps_shared_cpp/src/rmw_event.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_event.cpp
@@ -26,10 +26,15 @@ static const std::unordered_set<rmw_event_type_t> g_rmw_event_type_set{
   RMW_EVENT_OFFERED_DEADLINE_MISSED
 };
 
+namespace rmw_fastrtps_shared_cpp_internal
+{
+
 bool is_event_supported(rmw_event_type_t event_type)
 {
-  return g_rmw_event_type_set.count(event_type) > 0;
+  return g_rmw_event_type_set.count(event_type) == 1;
 }
+
+}  // namespace rmw_fastrtps_shared_cpp_internal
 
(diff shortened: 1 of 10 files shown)


rmw_fastrtps_shared_cpp/src/rmw_event.cpp
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+#include <unordered_set>
+
+#include "rmw/impl/cpp/macros.hpp"
+
+#include "rmw_fastrtps_shared_cpp/rmw_common.hpp"
+#include "types/event_types.hpp"
+
+static const std::unordered_set<rmw_event_type_t> __rmw_event_type_set{
```
wjwwood: Don't use `__`, it is reserved for the standard library.
Use `g_` instead since this is a global.

rmw_fastrtps_shared_cpp/src/types/event_types.hpp
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TYPES__EVENT_TYPES_HPP_
+#define TYPES__EVENT_TYPES_HPP_
+
+#include "rmw/event.h"
+
+bool is_event_supported(rmw_event_type_t event_type);
```
wjwwood: Why is this not in a namespace?

mm318: APIs in `rmw_fastrtps_shared_cpp` that aren't called from `rmw_fastrtps_cpp` don't seem to be put into the `rmw_fastrtps_shared_cpp` namespace.

rmw_fastrtps_shared_cpp/src/rmw_event.cpp
mm318: Done.

mm318:
What is the state of `rmw_fastrtps_dynamic_cpp`? Is it still supported and need these functions to be added for it?

wjwwood:
> What is the state of rmw_fastrtps_dynamic_cpp? Is it still supported and need these functions to be added for it?

I'm not 100% sure, but I think that they are needed. Looks like you added them already though.

rmw_fastrtps_shared_cpp/src/rmw_event.cpp
```diff
@@ -0,0 +1,66 @@
...
+static const std::unordered_set<rmw_event_type_t> g_rmw_event_type_set{
+  RMW_EVENT_LIVELINESS_CHANGED,
+  RMW_EVENT_REQUESTED_DEADLINE_MISSED,
+  RMW_EVENT_LIVELINESS_LOST,
+  RMW_EVENT_OFFERED_DEADLINE_MISSED
+};
+
+bool is_event_supported(rmw_event_type_t event_type)
+{
+  return g_rmw_event_type_set.count(event_type) > 0;
```
ivanpauno: I would be more precise, and use `== 1`. If it's greater than 1, there is something wrong with those macros.

rmw_fastrtps_shared_cpp/src/types/event_types.hpp
ivanpauno: Though that might be the case, I think every function should be defined in a namespace and not be global (except functions marked as `static`, which is not the case here).
No-polluting the global namespace is also recommended by the [GSG](https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions).

ivanpauno:
> What is the state of rmw_fastrtps_dynamic_cpp? Is it still supported and need these functions to be added for it?

AFAIK, it's still a tier 2 rmw implementation.

rmw_fastrtps_shared_cpp/src/rmw_event.cpp
mm318: Done.

rmw_fastrtps_shared_cpp/src/types/event_types.hpp
mm318: I have moved this function into a namespace called `rmw_fastrtps_shared_cpp_internal`.

wjwwood: I'd suggest `rmw_fastrtps_shared_cpp::internal` instead.

wjwwood: But I don't think you have to change it to move forward, only if you'd like to.

mm318: I have changed to the namespace to `rmw_fastrtps_shared_cpp::internal`. Thanks.

PR #358, opened 2020-03-31
Correct error message when event is not supported

ivanpauno:
Follow up of https://github.com/ros2/rmw_fastrtps/pull/354.

In this way, it prints `rmw_fastrtps_cpp` or `rmw_fastrtps_dynamic_cpp` accordingly.


ivanpauno:
CI: https://github.com/ros2/rmw_connext/pull/400#issuecomment-606706513

PR #361, opened 2020-04-02
Add support for security logging plugin

kyrofa:
If a `logging.xml` file is included in the security files, enable the security logging plugin in Fast RTPS using the properties contained in said file. Also reorganize existing security logic to be more testable, and add some tests for it.

This PR is ready for review, but requires https://github.com/eProsima/Fast-RTPS/pull/1125, thus the draft status.


kyrofa:
Let's go a simpler direction for Foxy.

sloretz:
Replaced by https://github.com/ros2/rmw_fastrtps/pull/362

PR #366, opened 2020-04-10
Add support for taking a sequence of messages

mjcarroll:
FastRTPS doesn't have built-in support for taking multiple samples, so
this implementation utilizes a loop over the `take` function multiple
times.

Implementation of ros2/rmw#212.

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -195,13 +195,13 @@ __rmw_take_sequence(
   rmw_subscription_allocation_t * allocation)
 {
   RCUTILS_CHECK_FOR_NULL_WITH_MSG(
-    subscription, "subscription pointer is null", return RMW_RET_ERROR);
+    subscription, "subscription pointer is null", return RMW_RET_INVALID_ARGUMENT);
   RCUTILS_CHECK_FOR_NULL_WITH_MSG(
-    message_sequence, "message_sequence pointer is null", return RMW_RET_ERROR);
+    message_sequence, "message_sequence pointer is null", return RMW_RET_INVALID_ARGUMENT);
   RCUTILS_CHECK_FOR_NULL_WITH_MSG(
-    message_info_sequence, "message_info_sequence pointer is null", return RMW_RET_ERROR);
+    message_info_sequence, "message_info_sequence pointer is null", return RMW_RET_INVALID_ARGUMENT);
   RCUTILS_CHECK_FOR_NULL_WITH_MSG(
-    taken, "size_t flag for count is null", return RMW_RET_ERROR);
+    taken, "size_t flag for count is null", return RMW_RET_INVALID_ARGUMENT);
 
   if (count > message_sequence->capacity) {
     RMW_SET_ERROR_MSG("Insufficient capacity in message_sequence");
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
+
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/serialized_message.h"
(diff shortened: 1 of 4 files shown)


rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -86,6 +86,41 @@ _take(
...
+  size_t * taken,
+  rmw_subscription_allocation_t * allocation)
+{
+  *taken = 0;
+  bool taken_flag = false;
+  rmw_ret_t ret = RMW_RET_OK;
+
+  for (size_t ii = 0; ii < count; ++ii) {
+    taken_flag = false;
+    ret = _take(
```
wjwwood, line 105: Is this safe with respect to asynchronous incoming data? Since we're not calling a Fast-RTPS function that does the taking, can new data slip in while taking here? Is that a problem?

mjcarroll: As far as I can tell, there isn't an equivalent FastRTPS method for retrieving a sequence of samples as there is in Connext and Cyclone.  

I think in this case, it would be possible for additional data to slip in asynchronously while we are servicing this request.  If this is a problem (I'm not sure that it is?) then we could mark this function unsupported for FastRTPS.  

Alternatively, we could bound the `count` by the number of unread messages that the subscription has available when this method is called, and not allow any additional ones to be read beyond that.  I think this could still be susceptible to async issues depending on the ordering that messages get returned from the `takeNextData` call.

wjwwood: We talked about this offline, and we decided that since we're taking no more than number available at the beginning of this function (bounded by `count`), the negative issues of this are minimal. However, if would be better if we could have a take sequence provided by fast-rtps or a way to "lock" the history cache while we're taking things, and then unlock it afterwards. @mjcarroll is going to ping people to see if we can get one of the two aforementioned features at some point to make this better.

mjcarroll: @MiguelCompany perhaps you can weigh in here.

Since there isn't currently a mechanism for locking and then taking multiple sames at the same time, this simply iterates over the `takeNextData` multiple times.  There are two potential consequences to this though:

1. After the `_take_sequence` call begins, new data becomes available, causing more data to be returned than would have been otherwise expected.  This could be particularly dicey in the case that a queue depth was set to `1`, and after `takeNextData` a new piece of data immediately becomes available, causing this method to return 2 samples, which would be unexpected given the queue depth.  I believe that this can be solved by at least capping the maximum number of samples to be read to be equal to the number currently unread at the beginning of the call (as implemented in da7b410)

2. After the `_take_sequence` call begins, new data becomes available and changes the original ordering of the underlying queue or store that `takeNextData` is reading from.  This is similar to (1), but without a substantial workaround.

I think the ideal scenario is that FastRTPS could either expose a sufficient API to be able to either:

1) take a sequence of data via an API call (perhaps a `takeNextData(std::vector<SampleT>)` or equivalent)
2) Allow the buffers that are to be read from to be locked, so that we could emulate this behavior as a downstream user of the library.

Would appreciate your thoughts here.

MiguelCompany: @mjcarroll I agree that the ideal scenario is to have an API to take a sequence of data. We are going to work on that after we release Fast RTPS 2.0

In the meantime, I think there is something that can be done without that kind of API available. The idea is to make the [SubListener](https://github.com/ros2/rmw_fastrtps/blob/86ae164414ee78a9f0d60caf463a31c6bb26631f/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp#L81) and `_take_sequence` mutually exclusive.

Thoughts?

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -86,6 +88,55 @@ _take(
   return RMW_RET_OK;
 }
 
+rmw_ret_t
```
jacobperron, line 91: `static` ?

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -131,6 +182,40 @@ __rmw_take(
...
+  rmw_subscription_allocation_t * allocation)
+{
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    subscription, "subscription pointer is null", return RMW_RET_ERROR);
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    message_sequence, "message_sequence pointer is null", return RMW_RET_ERROR);
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    message_info_sequence, "message_info_sequence pointer is null", return RMW_RET_ERROR);
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    taken, "size_t flag for count is null", return RMW_RET_ERROR);
```
jacobperron, line 202: I think the above error codes should be changed to `RMW_RET_INVALID_ARGUMENT`, right?

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
jacobperron: I think implementing option (2) makes sense for now.

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
mjcarroll: Why is that?

jacobperron: It looks like the definition of this function should only be used within this translation unit. But I also just realized this is C++ and we have a namespace too, so it's not a big deal.

jacobperron: I think also the `_take` and `_assign_message_info` functions could also be `static`. But it's a bit outside the scope of this PR.

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
mjcarroll: c493ccc

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
mjcarroll: After discussing with @wjwwood and @jacobperron, I would like to get this merged as-is before API freeze, and then add the locking before feature freeze.

While the locking looks straightforward to implement, I would like to insure that I don't introduce any deadlocks.

PR #371, opened 2020-04-22
export targets in a addition to include directories / libraries

dirk-thomas:
Related to ros2/ros2#904.


dirk-thomas:
Linux sanity build up to `rmw_fastrtps_cpp`, only testing the touched packages for CMake linting: [Build Status](https://ci.ros2.org/job/ci_linux/10406/)

PR #376, opened 2020-04-28
Fixed rmw_fastrtps_dynamic_cpp package description

ahcorde:

### DIFF
--- a/rmw_fastrtps_dynamic_cpp/package.xml
+++ b/rmw_fastrtps_dynamic_cpp/package.xml
@@ -3,7 +3,7 @@
 <package format="3">
   <name>rmw_fastrtps_dynamic_cpp</name>
   <version>0.8.1</version>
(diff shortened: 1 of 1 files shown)


dirk-thomas:
There is no "dynamic code generation". It probably should be "introspection type support".

rmw_fastrtps_dynamic_cpp/package.xml
```diff
@@ -3,7 +3,7 @@
 <package format="3">
   <name>rmw_fastrtps_dynamic_cpp</name>
   <version>0.8.1</version>
-  <description>Implement the ROS middleware interface using eProsima FastRTPS  static code generation in C++.</description>
+  <description>Implement the ROS middleware interface using introspection type support in C++.</description>
```
dirk-thomas, line 6: Sorry for only reviewing this incrementally. The typesupport is not limited to C++ - it also supports C types. So I would suggest to also remove the trailing "in C++".

PR #380, opened 2020-05-08
Update quality declarations on feature testing.

hidmic:
Precisely what the title says. As discussed offline, we defer unit testing and rely on integration and system tests up the stack (at least for the time being).

### DIFF
--- a/rmw_fastrtps_cpp/QUALITY_DECLARATION.md
+++ b/rmw_fastrtps_cpp/QUALITY_DECLARATION.md
@@ -85,7 +85,7 @@ The results of the test can be found [here](https://ci.ros2.org/view/nightly/job
 
 ### Feature Testing [4.i]
 
-There is no feature testing in `rmw_fastrtps_cpp`.
+All `rmw_fastrtps_cpp` public features are ROS middleware features. Integration and system tests up the stack provide coverage.
 
 ### Public API Testing [4.ii]
 
--- a/rmw_fastrtps_cpp/QUALITY_DECLARATION.md
+++ b/rmw_fastrtps_cpp/QUALITY_DECLARATION.md
@@ -85,7 +85,7 @@ The results of the test can be found [here](https://ci.ros2.org/view/nightly/job
 
 ### Feature Testing [4.i]
 
-All `rmw_fastrtps_cpp` public features are ROS middleware features. Integration and system tests up the stack provide coverage.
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_cpp/QUALITY_DECLARATION.md
```diff
@@ -85,7 +85,7 @@ The results of the test can be found [here](https://ci.ros2.org/view/nightly/job
 
 ### Feature Testing [4.i]
 
-There is no feature testing in `rmw_fastrtps_cpp`.
+All `rmw_fastrtps_cpp` public features are ROS middleware features. Integration and system tests up the stack provide coverage.
```
brawner, line 88: Be more specific about what "up the stack" means. Also in this case, since this package provides a lot of functionality that's depended on by other packages, I would provide explicit links to the packages and CI links where this feature testing is happening. Essentially, we're testing rmw_fastrtps_cpp all the time by making it a default configuration on ci.ros2.org and build.ros2.org. So it would be good to just state that that's how we know this is very carefully vetted.

rmw_fastrtps_shared_cpp/QUALITY_DECLARATION.md
```diff
@@ -85,7 +85,7 @@ The results of the test can be found [here](https://ci.ros2.org/view/nightly/job
 
 ### Feature Testing [4.i]
 
-Many of the features have testing in `rmw_fastrtps_shared_cpp`.
+Many of the features have testing in `rmw_fastrtps_shared_cpp`. Since `rmw_fastrtps_shared_cpp` supports `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp`, integration and system tests for ROS middlewares up the stack also provide coverage.
```
brawner, line 88: Same comment.

rmw_fastrtps_cpp/QUALITY_DECLARATION.md
hidmic: See https://github.com/ros2/rmw_fastrtps/pull/380/commits/4e7316ad6346f9e7a7253ae1ee81961ac346092c

rmw_fastrtps_shared_cpp/QUALITY_DECLARATION.md
hidmic: See https://github.com/ros2/rmw_fastrtps/pull/380/commits/4e7316ad6346f9e7a7253ae1ee81961ac346092c. I refrained from mentioning CI builds, as `rmw_fastrtps_shared_cpp` isn't an actual implementation and it already mentions `rmw_fastrtps_cpp`.

PR #386, opened 2020-05-18
Drop unused TopicCache sources.

hidmic:
Precisely what the title says. The functionality provided by the `TopicCache` class has been superseded and replaced by `rmw_dds_common::GraphCache`.


hidmic:
I think this is good to go. I don't expect MacOS or Windows CI to show any difference. @jacobperron do I have your approval?

PR #387, opened 2020-05-18
Add test coverage for GUID utilities.

hidmic:
Precisely what the title says. These tests simply verify that conversion in both directions is possible.

### DIFF
--- a/rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
+++ b/rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
@@ -25,6 +25,15 @@ static constexpr size_t byte_array_size =
   eprosima::fastrtps::rtps::GuidPrefix_t::size +
   eprosima::fastrtps::rtps::EntityId_t::size;
 
+TEST(GUIDUtilsTest, bad_arguments) {
+  eprosima::fastrtps::rtps::GUID_t guid;
+  uint8_t byte_array[byte_array_size] = {0};
+  uint8_t * null_byte_array = nullptr;
+  EXPECT_DEATH(copy_from_byte_array_to_fastrtps_guid(byte_array, nullptr), "guid");
+  EXPECT_DEATH(copy_from_byte_array_to_fastrtps_guid(null_byte_array, &guid), "byte_array");
+  EXPECT_DEATH(copy_from_fastrtps_guid_to_byte_array(guid, null_byte_array), "byte_array");
+}
+
 TEST(GUIDUtilsTest, byte_array_to_guid_and_back) {
   uint8_t input_byte_array[byte_array_size] = {0};
   input_byte_array[0] = 0xA5;
--- a/rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
+++ b/rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
@@ -1,0 +1,49 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


hidmic:
CI up to `rmw_fastrtps_shared_cpp`:


* Coverage [Build Status](https://ci.ros2.org/job/ci_linux_coverage/99/)

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "rmw_fastrtps_shared_cpp/guid_utils.hpp"
+
+using rmw_fastrtps_shared_cpp::copy_from_byte_array_to_fastrtps_guid;
+using rmw_fastrtps_shared_cpp::copy_from_fastrtps_guid_to_byte_array;
+
+static constexpr size_t byte_array_size =
+  eprosima::fastrtps::rtps::GuidPrefix_t::size +
+  eprosima::fastrtps::rtps::EntityId_t::size;
+
+TEST(GUIDUtilsTest, byte_array_to_guid_and_back) {
```
brawner, line 28: What do you think about checking for null arrays using EXPECT_DEATH()?

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+  uint8_t input_byte_array[byte_array_size] = {0};
+  input_byte_array[0] = 0xA5;
+  input_byte_array[byte_array_size - 1] = 0x4B;
+  eprosima::fastrtps::rtps::GUID_t guid;
+  copy_from_byte_array_to_fastrtps_guid(input_byte_array, &guid);
```
brawner, line 33: While converting back and forth between the two does ensure that no data is lost, it doesn't really test that it was converted to the correct output. It is also possible that this won't catch errors if they are introduced in the header if somehow both methods correct each other's mistake.

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "gtest/gtest.h"
+
+#include "fastrtps/rtps/common/Guid.h"
+
+#include "rmw_fastrtps_shared_cpp/guid_utils.hpp"
+
+using rmw_fastrtps_shared_cpp::copy_from_byte_array_to_fastrtps_guid;
+using rmw_fastrtps_shared_cpp::copy_from_fastrtps_guid_to_byte_array;
+
+static constexpr size_t byte_array_size =
```
brawner, line 24: Out of curiosity, what does byte_array_size evaluate to?

hidmic: It's 16, but I preferred not to hard-code it.

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
hidmic: Good point.

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
hidmic: That is true, I just didn't want to pollute the test with too much information as to how it gets copied in-n-out.

rmw_fastrtps_shared_cpp/test/test_guid_utils.cpp
hidmic: See https://github.com/ros2/rmw_fastrtps/pull/387/commits/7bcd43457d79a0a3108f25c0d26f15ef604f640b

brawner:
Windows failed the test because it didn't output a message when they died, I think it would be fine if the expected matcher was just an empty string.

Blast545:
Why isn't the output available in Windows as in the other platforms though? AFAIK the `assert` also outputs diagnostic information in Windows, should this be investigated?

> I think it would be fine if the expected matcher was just an empty string.

+1

hidmic:
@brawner PTAL

PR #388, opened 2020-05-18
Add test coverage for name mangling and namespacing-specific API.

hidmic:
Precisely what the title says.

CI up to `rmw_fastrtps_shared_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6207/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8822/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10783/)
* Coverage [Build Status](https://ci.ros2.org/job/ci_linux_coverage/107/)


rmw_fastrtps_shared_cpp/test/test_names.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include "gtest/gtest.h"
+
+#include "rmw/qos_profiles.h"
+
+#include "rmw_fastrtps_shared_cpp/names.hpp"
+#include "rmw_fastrtps_shared_cpp/namespace_prefix.hpp"
+
+TEST(NamespaceTest, get_prefix) {
+  EXPECT_EQ("", _get_ros_prefix_if_exists(""));
```
brawner, line 23: Add tests for:
* non-empty string with no prefix
* string with prefix but not matching '/', (i.e "rr_should_not_match")
* string where the prefix shows up elsewhere, but '/' is in the right place i.e. ("al/so_should_not_match_rr")

Substituting 'rr' with variable 'prefix'  in your loop.

I think these same tests apply to the TEST blocks below as well.

rmw_fastrtps_shared_cpp/test/test_names.cpp
```diff
@@ -0,0 +1,53 @@
...
+  for (const auto & prefix : _get_all_ros_prefixes()) {
+    EXPECT_EQ(prefix, _get_ros_prefix_if_exists(prefix + "/test"));
+  }
+}
+
+TEST(NamespaceTest, strip_prefix) {
+  EXPECT_EQ("/no_prefix/test", _strip_ros_prefix_if_exists("/no_prefix/test"));
```
brawner, line 30: I think strip_prefix and resolve_prefix should also test if its parameters are empty strings.

rmw_fastrtps_shared_cpp/test/test_names.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+TEST(NamespaceTest, resolve_prefix) {
+  EXPECT_EQ("", _resolve_prefix("/test", "some_ros_prefix"));
+  EXPECT_EQ(
+    "/test_some_ros_prefix",
+    _resolve_prefix("some_ros_prefix/test_some_ros_prefix", "some_ros_prefix"));
+}
+
+TEST(NamespaceTest, name_mangling) {
```
brawner, line 43: Add tests for if arguments are nullptrs.

hidmic: Good catch, had to add a few `assert` to avoid segfaults. See https://github.com/ros2/rmw_fastrtps/pull/388/commits/b029a312f4e049efbb609f1b9883650d7b5ffe4f.

rmw_fastrtps_shared_cpp/test/test_names.cpp
hidmic: Done in https://github.com/ros2/rmw_fastrtps/pull/388/commits/b029a312f4e049efbb609f1b9883650d7b5ffe4f. Had to change implementation a tiny bit for it to behave correctly in all cases.

rmw_fastrtps_shared_cpp/test/test_names.cpp
hidmic: Done in https://github.com/ros2/rmw_fastrtps/pull/388/commits/b029a312f4e049efbb609f1b9883650d7b5ffe4f.

rmw_fastrtps_shared_cpp/test/test_names.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "rmw_fastrtps_shared_cpp/names.hpp"
+#include "rmw_fastrtps_shared_cpp/namespace_prefix.hpp"
+
+TEST(NamespaceTest, get_prefix) {
+  EXPECT_EQ("", _get_ros_prefix_if_exists(""));
+  EXPECT_EQ("", _get_ros_prefix_if_exists("not/a_ros_prefix"));
+  for (const auto & prefix : _get_all_ros_prefixes()) {
+    EXPECT_EQ("", _get_ros_prefix_if_exists(prefix + "_should_not_match"));
+    EXPECT_EQ("", _get_ros_prefix_if_exists("this/should_not_match/" + prefix));
```
brawner, line 27: I think forward slash should be at prefix.length(), since that's what all the if statements are checking for. 

```suggestion
    EXPECT_EQ("", _get_ros_prefix_if_exists("th/is_should_not_match/" + prefix));
```

rmw_fastrtps_shared_cpp/src/namespace_prefix.cpp
```diff
@@ -31,8 +31,10 @@ const std::vector<std::string> _ros_prefixes =
 std::string
 _resolve_prefix(const std::string & name, const std::string & prefix)
 {
-  if (name.rfind(prefix, 0) == 0 && name.at(prefix.length()) == '/') {
-    return name.substr(prefix.length());
+  if (!name.empty()) {
```
brawner, line 34: I may be missing something, but why don't you need this check in the other functions?

hidmic: Because the other functions don't use the [`std::string::at()`](https://en.cppreference.com/w/cpp/string/basic_string/at) method, which will throw if its argument is out of range.

brawner:
Looks good, but it looks like there is a strange linker error with windows.

rmw_fastrtps_shared_cpp/src/namespace_prefix.cpp
Blast545: I think the [`std::string::rfind()`](https://en.cppreference.com/w/cpp/string/basic_string/rfind) function will return `npos` if the prefix is not found so the `at` function won't be used if the string is empty. Then the added check won't be needed.

Same below where the `at` function is used after a call to `rfind`.

brawner: Ok, I think I get it now. But it leads to more questions actually. I think this logic needs to be updated and some more tests added. If the name or topic_name is "rr" or "rt" or "rq", these functions will throw. And lines 47 and 59 use std::string.at()

hidmic: All very good points. I've updated both implementation and tests in https://github.com/ros2/rmw_fastrtps/pull/388/commits/564379569c91890136ef6aa44f1a84bbb2237726

hidmic:
> Looks good, but it looks like there is a strange linker error with windows.

Hmm, that's a symbol visibility issue. I'm starting to think this header is only half public. I can't find call sites outside this package.

hidmic:
Alright, I rebased and rewrote the commit history to have https://github.com/ros2/rmw_fastrtps/pull/388/commits/5029e4fdecf7c3e0aed1761caba85a49feeebcba changing API visibility and https://github.com/ros2/rmw_fastrtps/pull/388/commits/2a0acb3e83cded5bcd4261cf8db3a9902a5e0822 adding tests. PTAL !

rmw_fastrtps_shared_cpp/src/namespace_prefix.cpp
```diff
@@ -31,8 +31,10 @@ const std::vector<std::string> _ros_prefixes =
 std::string
 _resolve_prefix(const std::string & name, const std::string & prefix)
 {
-  if (name.rfind(prefix, 0) == 0 && name.at(prefix.length()) == '/') {
-    return name.substr(prefix.length());
+  if (name.length() > prefix.length()) {
```
brawner, line 34: Coming back with fresh eyes on this one, I think this can all be simplified down to:
```
if (name.rfind(prefix + "/", 0)) {
  return name.substr(prefix.length());
}
```

Same thing with the functions below.

hidmic: That is correct. Fixed in https://github.com/ros2/rmw_fastrtps/pull/388/commits/e0fe9ca02c9d9287f990f7855f87b957e6470f48.

hidmic:
Alright! Going in.

ivanpauno:
This seem to have introduced a regression: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1556/testReport/junit/projectroot/test/test_names/

hidmic:
I bet those asserts were removed in Release mode...

ivanpauno:
> I bet those asserts were removed in Release mode...

I think you can make use of `NDEBUG` macro to solve the problem, see https://en.cppreference.com/w/cpp/error/assert

PR #390, opened 2020-05-24
Make service wait for response reader

MiguelCompany:
This addresses ros2/ros2#922

The idea is to send the GUID of the response subscription on each request (using related_sample_identity), so the service could check if that reader is already matched to the response publisher when sending the response.

I've made the changes so they are backwards compatible. The service will only use and check the related guid when received. The client will check the received related guid against both the response reader and the request writer, in case an old service answers with the latter.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
@@ -49,6 +49,7 @@ typedef struct CustomClientInfo
   eprosima::fastrtps::Publisher * request_publisher_;
   ClientListener * listener_;
   eprosima::fastrtps::rtps::GUID_t writer_guid_;
+  eprosima::fastrtps::rtps::GUID_t reader_guid_;
   eprosima::fastrtps::Participant * participant_;
   const char * typesupport_identifier_;
   ClientPubListener * pub_listener_;
@@ -88,7 +89,8 @@ public:
       if (eprosima::fastrtps::rtps::ALIVE == response.sample_info_.sampleKind) {
         response.sample_identity_ = response.sample_info_.related_sample_identity;
 
-        if (response.sample_identity_.writer_guid() == info_->writer_guid_) {
+        if (response.sample_identity_.writer_guid() == info_->reader_guid_ ||
+            response.sample_identity_.writer_guid() == info_->writer_guid_) {
           std::lock_guard<std::mutex> lock(internalMutex_);
 
           if (conditionMutex_ != nullptr) {
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
@@ -90,7 +90,8 @@ public:
         response.sample_identity_ = response.sample_info_.related_sample_identity;
 
         if (response.sample_identity_.writer_guid() == info_->reader_guid_ ||
-            response.sample_identity_.writer_guid() == info_->writer_guid_) {
+          response.sample_identity_.writer_guid() == info_->writer_guid_)
+        {
           std::lock_guard<std::mutex> lock(internalMutex_);
 
           if (conditionMutex_ != nullptr) {
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -34,6 +34,7 @@
 #include "rmw_fastrtps_shared_cpp/TypeSupport.hpp"
 
 class ServiceListener;
+class ServicePubListener;
 
 typedef struct CustomServiceInfo
 {
@@ -44,6 +45,7 @@ typedef struct CustomServiceInfo
   eprosima::fastrtps::Subscriber * request_subscriber_;
   eprosima::fastrtps::Publisher * response_publisher_;
   ServiceListener * listener_;
+  ServicePubListener * pub_listener_;
   eprosima::fastrtps::Participant * participant_;
   const char * typesupport_identifier_;
 } CustomServiceInfo;
@@ -163,6 +165,47 @@ private:
   std::atomic_bool list_has_data_;
   std::mutex * conditionMutex_ RCPPUTILS_TSA_GUARDED_BY(internalMutex_);
   std::condition_variable * conditionVariable_ RCPPUTILS_TSA_GUARDED_BY(internalMutex_);
+};
+
+class ServicePubListener : public eprosima::fastrtps::PublisherListener
+{
+public:
+  explicit ServicePubListener() = default;
+  
+  template< class Rep, class Period >
+  bool wait_for_subscription(
+    const eprosima::fastrtps::rtps::GUID_t & guid,
+    const std::chrono::duration<Rep, Period> & rel_time)
+  {
+    auto guid_is_present = [this, guid]() -> bool
+    {
+      return subscriptions_.find(guid) != subscriptions_.end();
+    };
+    
+    std::unique_lock<std::mutex> lock(mutex_);
+    return cv_.wait_for(lock, rel_time, guid_is_present);
+  }
+
+  void onPublicationMatched(
+    eprosima::fastrtps::Publisher * pub,
+    eprosima::fastrtps::rtps::MatchingInfo & matchingInfo)
+  {
+    (void) pub;
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (eprosima::fastrtps::rtps::MATCHED_MATCHING == matchingInfo.status) {
+      subscriptions_.insert(matchingInfo.remoteEndpointGuid);
+    } else if (eprosima::fastrtps::rtps::REMOVED_MATCHING == matchingInfo.status) {
+      subscriptions_.erase(matchingInfo.remoteEndpointGuid);
+    } else {
+      return;
+    }
+    cv_.notify_all();
+  }
+
+private:
+  std::mutex mutex_;
+  std::set<eprosima::fastrtps::rtps::GUID_t> subscriptions_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+  std::condition_variable cv_;
 };
 
 #endif  // RMW_FASTRTPS_SHARED_CPP__CUSTOM_SERVICE_INFO_HPP_
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -84,6 +84,12 @@ public:
     if (sub->takeNextData(&data, &request.sample_info_)) {
       if (eprosima::fastrtps::rtps::ALIVE == request.sample_info_.sampleKind) {
         request.sample_identity_ = request.sample_info_.sample_identity;
+        // Use response subscriber guid (on related_sample_identity) when present.
+        const eprosima::fastrtps::rtps::GUID_t& reader_guid =
+          request.sample_info_.related_sample_identity.writer_guid();
+        if (reader_guid != eprosima::fastrtps::rtps::GUID_t::unknown() ) {
+          request.sample_identity_.writer_guid() = reader_guid;
+        }
 
         std::lock_guard<std::mutex> lock(internalMutex_);
 
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -19,6 +19,7 @@
 #include <condition_variable>
 #include <list>
 #include <mutex>
+#include <set>
 
 #include "fastcdr/FastBuffer.h"
 
@@ -87,7 +88,7 @@ public:
       if (eprosima::fastrtps::rtps::ALIVE == request.sample_info_.sampleKind) {
         request.sample_identity_ = request.sample_info_.sample_identity;
         // Use response subscriber guid (on related_sample_identity) when present.
-        const eprosima::fastrtps::rtps::GUID_t& reader_guid =
+        const eprosima::fastrtps::rtps::GUID_t & reader_guid =
           request.sample_info_.related_sample_identity.writer_guid();
         if (reader_guid != eprosima::fastrtps::rtps::GUID_t::unknown() ) {
           request.sample_identity_.writer_guid() = reader_guid;
@@ -170,18 +171,18 @@ private:
 class ServicePubListener : public eprosima::fastrtps::PublisherListener
 {
 public:
-  explicit ServicePubListener() = default;
-  
-  template< class Rep, class Period >
+  ServicePubListener() = default;
+
+  template<class Rep, class Period>
   bool wait_for_subscription(
     const eprosima::fastrtps::rtps::GUID_t & guid,
     const std::chrono::duration<Rep, Period> & rel_time)
   {
     auto guid_is_present = [this, guid]() -> bool
-    {
-      return subscriptions_.find(guid) != subscriptions_.end();
-    };
-    
+      {
+        return subscriptions_.find(guid) != subscriptions_.end();
+      };
+
     std::unique_lock<std::mutex> lock(mutex_);
     return cv_.wait_for(lock, rel_time, guid_is_present);
   }
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -19,7 +19,7 @@
 #include <condition_variable>
 #include <list>
 #include <mutex>
-#include <set>
+#include <unordered_set>
 
 #include "fastcdr/FastBuffer.h"
 
@@ -33,6 +33,7 @@
 #include "rcpputils/thread_safety_annotations.hpp"
 
 #include "rmw_fastrtps_shared_cpp/TypeSupport.hpp"
+#include "rmw_fastrtps_shared_cpp/guid_utils.hpp"
 
 class ServiceListener;
 class ServicePubListener;
@@ -204,8 +205,12 @@ public:
   }
 
 private:
+  using subscriptions_set_t =
+    std::unordered_set<eprosima::fastrtps::rtps::GUID_t,
+      rmw_fastrtps_shared_cpp::hash_fastrtps_guid>;
+
   std::mutex mutex_;
-  std::set<eprosima::fastrtps::rtps::GUID_t> subscriptions_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+  subscriptions_set_t subscriptions_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
   std::condition_variable cv_;
 };
 
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
@@ -62,13 +62,14 @@ struct hash_fastrtps_guid
   {
     union u_convert {
       uint8_t plain_value[sizeof(guid)];
-      uint32_t plain_ints[4];
+      uint32_t plain_ints[sizeof(guid)/sizeof(uint32_t)];
     } u;
 
     static_assert(
+      sizeof(guid) == 16 &&
       sizeof(u.plain_value) == sizeof(u.plain_ints) &&
       offsetof(u_convert, plain_value) == offsetof(u_convert, plain_ints),
-      "ByteT should be either int8_t or uint8_t");
+      "Plain guid should be easily convertible to uint32_t[4]");
 
     copy_from_fastrtps_guid_to_byte_array(guid, u.plain_value);
 
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
@@ -16,6 +16,7 @@
 #define RMW_FASTRTPS_SHARED_CPP__GUID_UTILS_HPP_
 
 #include <cassert>
+#include <cstddef>
 #include <cstring>
 #include <type_traits>
 
@@ -54,6 +55,36 @@ copy_from_fastrtps_guid_to_byte_array(
   memcpy(guid_byte_array, &guid.guidPrefix, prefix_size);
   memcpy(&guid_byte_array[prefix_size], &guid.entityId, guid.entityId.size);
 }
+
+struct hash_fastrtps_guid
+{
+  std::size_t operator()(const eprosima::fastrtps::rtps::GUID_t & guid) const
+  {
+    union u_convert
+    {
+      uint8_t plain_value[sizeof(guid)];
+      uint32_t plain_ints[4];
+    } u;
+
+    static_assert(
+      sizeof(u.plain_value) == sizeof(u.plain_ints) &&
+      offsetof(u_convert, plain_value) == offsetof(u_convert, plain_ints),
+      "ByteT should be either int8_t or uint8_t");
+
+    copy_from_fastrtps_guid_to_byte_array(guid, u.plain_value);
+
+    constexpr std::size_t prime_1 = 7;
+    constexpr std::size_t prime_2 = 31;
+    constexpr std::size_t prime_3 = 59;
+
+    size_t ret_val = prime_1 * u.plain_ints[0];
+    ret_val = prime_2 * (u.plain_ints[1] + ret_val);
+    ret_val = prime_3 * (u.plain_ints[2] + ret_val);
+    ret_val = u.plain_ints[3] + ret_val;
+
+    return ret_val;
+  }
+};
 
 }  // namespace rmw_fastrtps_shared_cpp
 
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
@@ -60,8 +60,7 @@ struct hash_fastrtps_guid
 {
   std::size_t operator()(const eprosima::fastrtps::rtps::GUID_t & guid) const
   {
-    union u_convert
-    {
+    union u_convert {
       uint8_t plain_value[sizeof(guid)];
       uint32_t plain_ints[4];
     } u;
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
@@ -62,7 +62,7 @@ struct hash_fastrtps_guid
   {
     union u_convert {
       uint8_t plain_value[sizeof(guid)];
-      uint32_t plain_ints[sizeof(guid)/sizeof(uint32_t)];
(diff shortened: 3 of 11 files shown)


rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,18 @@ __rmw_send_response(
   wparams.related_sample_identity().sequence_number().low =
     (int32_t)(request_header->sequence_number & 0xFFFFFFFF);
 
+  const eprosima::fastrtps::rtps::GUID_t& related_guid =
```
hidmic, line 107: @MiguelCompany nit: 
```suggestion
  const eprosima::fastrtps::rtps::GUID_t & related_guid =
```

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,18 @@ __rmw_send_response(
...
+    wparams.related_sample_identity().writer_guid();
+  if( (related_guid.entityId.value[3] & 0x04) != 0)
```
hidmic, line 109: @MiguelCompany nit: 
```suggestion
  if((related_guid.entityId.value[3] & 0x04) != 0)
```

hidmic: Also, a comment explaining this bit fiddling would be great.

MiguelCompany:
@hidmic Thanks for your review.

> I wonder, does (ab)using the writer GUID in the related sample identity compromises cross-vendor communication? We'll run full CI anyways, but I figured I'd ask.

Cyclone is not interoperable with Fast RTPS on services, and I'm not sure about RTI Connext. The latter would most probably ignore the related_guid from the request, which is covered by the following lines.
https://github.com/ros2/rmw_fastrtps/blob/fce9022597c373b6ce56bf08e3fef29a5b65db30/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp#L92-L93

hidmic:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10885/)

hidmic:
@MiguelCompany mind to fix linter issues?

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,22 @@ __rmw_send_response(
...
     (int32_t)(request_header->sequence_number & 0xFFFFFFFF);
 
+  // According to the list of possible entity kinds in section 9.3.1.2 of RTPS
+  // readers will have this bit on, while writers will not. We use this to know
+  // if the related guid is the request writer or the response reader.
+  constexpr uint8_t entity_id_is_reader_bit = 0x04;
+  const eprosima::fastrtps::rtps::GUID_t & related_guid =
+    wparams.related_sample_identity().writer_guid();
+  if((related_guid.entityId.value[3] & entity_id_is_reader_bit) != 0)
+  {
```
jacobperron, line 114: I think this will cause a linter error
```suggestion
  if ((related_guid.entityId.value[3] & entity_id_is_reader_bit) != 0) {
```

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,22 @@ __rmw_send_response(
...
+    // Related guid is a reader, so it is the response subscription guid.
+    // Wait for the response writer to be matched with it.
+    if(!info->pub_listener_->wait_for_subscription(related_guid, std::chrono::milliseconds(100))) {
```
jacobperron, line 117:
```suggestion
    if (!info->pub_listener_->wait_for_subscription(related_guid, std::chrono::milliseconds(100))) {
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
```diff
@@ -88,7 +89,8 @@ class ClientListener : public eprosima::fastrtps::SubscriberListener
       if (eprosima::fastrtps::rtps::ALIVE == response.sample_info_.sampleKind) {
         response.sample_identity_ = response.sample_info_.related_sample_identity;
 
-        if (response.sample_identity_.writer_guid() == info_->writer_guid_) {
+        if (response.sample_identity_.writer_guid() == info_->reader_guid_ ||
+            response.sample_identity_.writer_guid() == info_->writer_guid_) {
```
ivanpauno, line 93: I think the linter doesn't accept that style (it the linter is happy, no change needed).

Suggested style:
```cpp
        if (response.sample_identity_.writer_guid() == info_->reader_guid_ ||
          response.sample_identity_.writer_guid() == info_->writer_guid_)
        {
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
```diff
@@ -159,4 +167,45 @@ class ServiceListener : public eprosima::fastrtps::SubscriberListener
...
+      subscriptions_.erase(matchingInfo.remoteEndpointGuid);
+    } else {
+      return;
+    }
+    cv_.notify_all();
+  }
+
+private:
+  std::mutex mutex_;
+  std::set<eprosima::fastrtps::rtps::GUID_t> subscriptions_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
```
ivanpauno, line 207: nit: prefer `std::unordered_set`

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
ivanpauno, line 117: `rmw_send_response` was not blocking, and it should be continue being not blocking.

jacobperron:
> I don't think that blocking in rmw_send_response is acceptable.

As I understand, blocking is the primary patch being proposed. Though it is not documented in the `rmw` API, the [`rcl` API](https://github.com/ros2/rcl/blob/65c2d46ee0d2def3c991ecac6674f0366406900c/rcl/include/rcl/service.h#L271) does make the assumption that it is a non-blocking call (I think we should update the documentation if we merge this PR). At least there is a timeout so we won't block indefinitely, but I agree it would be better if we did not block at all. However, what are the alternatives? Do you think a patch similar to what's being proposed in https://github.com/ros2/rmw_cyclonedds/pull/187 would work? (Though, it looks like the [rmw_cyclonedds patch is still blocking](https://github.com/ros2/rmw_cyclonedds/blob/4c3b8fa72b37dd76ee29006cbe1c93f59b7f1ffb/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3282-L3290) similar to this PR)

ivanpauno:
> As I understand, blocking is the primary patch being proposed.

Yes.

> At least there is a timeout so we won't block indefinitely, but I agree it would be better if we did not block at all

:+1:

> However, what are the alternatives?

A mechanism to ensure that `rmw_is_service_server_available` returns `true` when the client discovered the server and the server the client.

> Do you think a patch similar to what's being proposed in ros2/rmw_cyclonedds#187 would work? (Though, it looks like the rmw_cyclonedds patch is still blocking similar to this PR)

I thought the cyclonedds PR could solve the problem without needing to introduce blocking behavior, but thinking about it again, it isn't true.

----

If many people left a :+1:, I think that introducing the blocking behavior and documenting it is ok.

hidmic:
Same as in https://github.com/ros2/rmw_cyclonedds/pull/187, I think this is OK for a short-term workaround. And while blocking isn't ideal, even if discovery takes up to 100ms to happen we've simply moved the wait from client to server (which isn't strictly worst, performance-wise, since discovery is async).

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
wjwwood: What if we change this to `0` timeout? That way we at least know that it won't work and can react to it.

wjwwood:
If others on the team feel that the urgency in fixing this issue outweighs the concerns about the workaround, then I'll go along with it, but I cannot enthusiastically approve this workaround (nor the cyclone one last I looked, but maybe it was updated in the meantime).

eboasson:
Interesting ... similar approaches to fixing the same problems at the same time 🙂

@wjwwood @hidmic @ivanpauno, I share your dislike about these workarounds. The way I see it, the options are:
1. stick with dashing/eloquent behaviour (broken, but not so badly that it is unusable in practice);
2. accept blocking in `rmw_send_response` as a workaround (with the risk of it becoming permanent);
3. don't block, instead always handle the first *serviceable* request in the queue (but this comes at a significant increase in complexity, which is a risk in itself and consumes even more development effort on a throwaway thing).

Evidently, the proper solution is fixing the deficiency in the DDS spec (or the implementation of it) but at least for Cyclone there are process barriers that prevent doing that prior to the Foxy release. It seems to me 1 and 2 are the only sensible options. Pick your poison ...

MiguelCompany:
@eboasson 
> 1. stick with dashing/eloquent behaviour (broken, but not so badly that it is unusable in practice);
> 2. accept blocking in `rmw_send_response` as a workaround (with the risk of it becoming permanent);
> 3. don't block, instead always handle the first _serviceable_ request in the queue (but this comes at a significant increase in complexity, which is a risk in itself and consumes even more development effort on a throwaway thing).

I think there is a fourth option

4. don't block and return a specific error if we know that the response is not going to be received. This way we let the upper layer decide whether to wait or not.

I can even think of new a new API function allowing the user to wait till the response can be reliably sent.

eboasson:
@MiguelCompany certainly there is that 4th option, or the 5th of an additional interface like you mention, but those make the application responsible for dealing with the problem. I just don't believe this should be the application's problem.

Furthermore, just like @wjwwood I worry about temporary workarounds becoming permanent, and a workaround that involves changes to the application will definitely never go away ...

ivanpauno:
@eboasson @MiguelCompany I would stick with the old poison, i.e. option 2.
The issue is easy to document, and users can work it around if they need (and we can work it around in our tests too).

It would be great in early G-Turtle to iterate on a solution of the problem.

> Evidently, the proper solution is fixing the deficiency in the DDS spec (or the implementation of it) but at least for Cyclone there are process barriers that prevent doing that prior to the Foxy release. It seems to me 1 and 2 are the only sensible options. Pick your poison ...

I think that more than an issue of the DDS spec, we assumed in the service implementation something that is not ensured: if A discovered B, B has already discovered A.
If needed, we could add a side-channel "discovery mechanism", to ensure `rmw_is_service_server_available` returns `true` correctly.

Note: I'm not sure if `rmw_connext` implementation has the same problems, as we are using an implementation of RPC over DDS they provide.

hidmic:
> I think that more than an issue of the DDS spec, we assumed in the service implementation something that is not ensured: if A discovered B, B has already discovered A.

On that note. [OMG DDS-RPC 1.0 spec](https://www.omg.org/spec/DDS-RPC/1.0/PDF) covers the discovery problem for RPC services in section 7.6. It acknowledges that independent discovery isn't robust and can lead to races (_Basic Service Mapping_ under section 7.6.1). It then suggests an extension to the data type sent over discovery built-in topics and an algorithm (_Enhanced Service Mapping_ under section 7.6.2), which basically boils down to the client not sending requests until it has discovered the server's response data writer and the server not picking up those requests until it has discovered the client's response data reader. 

Something like that would make the most sense to me, providing for reliable service operation without cluttering user code with fail-safe/retry logic. But I can understand that's unlikely to land for Foxy. Thus, I'd stick to poison No. 2 too, keeping changes under the `rmw` hood so we can backport a proper fix in a future patch release.

jacobperron:
In lieu of a proper fix (like the one suggested by @hidmic), I'd be okay with a version of option 2 landing for Foxy. I agree we should aim to replace the patch with a better solution early in the G-turtle cycle.

wjwwood:
Ok, option 2 is ok with me, since the caveats are known.

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
MiguelCompany: Fixed on 13c540b

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
MiguelCompany: Fixed on 13c540b

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
MiguelCompany: Fixed on 13c540b

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
MiguelCompany: Done in 01dd156. I had to add a hash function for `GUID_t`.

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
MiguelCompany: Based on the conversation below, I assume the code is ok as it is right now.

If this be changed to an indefinite block, please tell me and I'll do it

MiguelCompany:
@hidmic @jacobperron @ivanpauno @wjwwood This is ready for review.

Just an open question. Should we then block indefinitely?

ivanpauno:
> @hidmic @jacobperron @ivanpauno @wjwwood This is ready for review.
> 
> Just an open question. Should we then block indefinitely?

No

> @eboasson @MiguelCompany I would stick with the old poison, i.e. option 2.

I want to clarify that with "old poison" I meant not doing changes and sticking with the old behavior, i.e. option 1. Sorry for the typo.
I hope not having confused the other people that commented :joy:.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
```diff
@@ -55,6 +56,35 @@ copy_from_fastrtps_guid_to_byte_array(
   memcpy(&guid_byte_array[prefix_size], &guid.entityId, guid.entityId.size);
 }
 
+struct hash_fastrtps_guid
+{
+  std::size_t operator()(const eprosima::fastrtps::rtps::GUID_t & guid) const
+  {
+    union u_convert {
+      uint8_t plain_value[sizeof(guid)];
+      uint32_t plain_ints[4];
```
hidmic, line 65: @MiguelCompany 
```suggestion
      uint32_t plain_ints[sizeof(guid)/sizeof(uint32_t)];
```
?

rmw_fastrtps_shared_cpp/src/rmw_service_server_is_available.cpp
```diff
@@ -100,11 +100,22 @@ __rmw_service_server_is_available(
     return RMW_RET_OK;
   }
 
-  if (0 == client_info->request_publisher_matched_count_.load()) {
+  if (number_of_request_subscribers != number_of_response_publishers) {
```
hidmic, line 103: @MiguelCompany why do still check for the global count? Isn't checking for the matched count enough? Won't this make this check fail for a fully matched client if another client for the same service is still halfway through discovery?

hidmic:
> Just an open question. Should we then block indefinitely?

IMHO indefinitely is not an option. Either you introduce some timeout (like in this patch) or ensure that you'll unblock if the client goes away before discovery completes (like in https://github.com/ros2/rmw_cyclonedds). And even that last bit is debatable. 

At any rate, this is temporary.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
MiguelCompany: Yeah. But we should also check we have 4 integers, as the calculation below is assuming that.

Should I also add a check to the `staic_assert` below for `sizeof(guid) == 16`?

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,22 @@ __rmw_send_response(
   wparams.related_sample_identity().sequence_number().low =
     (int32_t)(request_header->sequence_number & 0xFFFFFFFF);
 
+  // According to the list of possible entity kinds in section 9.3.1.2 of RTPS
+  // readers will have this bit on, while writers will not. We use this to know
+  // if the related guid is the request writer or the response reader.
+  constexpr uint8_t entity_id_is_reader_bit = 0x04;
```
hidmic, line 110: @MiguelCompany mind to a TODO for a proper fix? I opened an https://github.com/ros2/rmw_fastrtps/issues/392 to track that work.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
hidmic: > But we should also check we have 4 integers, as the calculation below is assuming that.

That is a good point. 

> Should I also add a check to the staic_assert below for sizeof(guid) == 16?

Fair enough.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
```diff
@@ -55,6 +56,35 @@ copy_from_fastrtps_guid_to_byte_array(
...
+    } u;
+
+    static_assert(
+      sizeof(u.plain_value) == sizeof(u.plain_ints) &&
+      offsetof(u_convert, plain_value) == offsetof(u_convert, plain_ints),
+      "ByteT should be either int8_t or uint8_t");
```
hidmic, line 71: @MiguelCompany also, `ByteT`?

rmw_fastrtps_shared_cpp/src/rmw_service_server_is_available.cpp
MiguelCompany: I assume we want to ensure that the discovered servers are consistent. This check will fail if there is a **server** halfway through discovery.

I guess it is ok if we just check the matched count for equality. It will also be inexact if the client is halfway matched with two servers, but it is an improvement nonetheless.

MiguelCompany:
> > Just an open question. Should we then block indefinitely?
> 
> IMHO indefinitely is not an option. Either you introduce some timeout (like in this patch) or ensure that you'll unblock if the client goes away before discovery completes (like in https://github.com/ros2/rmw_cyclonedds). And even that last bit is debatable.
> 
> At any rate, this is temporary.

We could then convert it to an unblocking check, removing the condition variable and returning an error directly if the reader has not been matched.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
MiguelCompany: Ups, forgot to change the message.

hidmic:
> We could then convert it to an unblocking check, removing the condition variable and returning an error directly if the reader has not been matched.

But then we're simply deferring the problem to calling code. IMHO 

- [ ] `rmw_is_server_available()` should imply both server and client have seen each other.
- [ ] `rmw_take_request()` shouldn't succeed if the client has not been fully discovered.

Also, ideally I'd like this implementation to be compliant with the spec so that it can interoperate with other vendors. That's what #392 is meant to track.

MiguelCompany:
> * [ ] `rmw_is_server_available()` should imply both server and client have seen each other.
> * [ ]  `rmw_take_request()` shouldn't succeed if the client has not been fully discovered.

I agree this is what it should be. But that will be a follow-up. We leave the current timeout then and add the TODO.

MiguelCompany:
@hidmic only [this comment](https://github.com/ros2/rmw_fastrtps/pull/390#discussion_r431853486) pending

rmw_fastrtps_shared_cpp/src/rmw_service_server_is_available.cpp
hidmic: > It will also be inexact if the client is halfway matched with two servers, but it is an improvement nonetheless.

That is a good point, but AFAIK the semantics of multiple servers for the same service in ROS 2 aren't well defined. I wouldn't worry as much about that use case. @wjwwood @ivanpauno for valuable input.

ivanpauno: Yes, I wouldn't worry about multiple servers behavior now.
I think that's another point where we can iterate in a future release.

MiguelCompany: Shall I revert 2590645a9de94ad5582415f9c7f258505ef9dc74 then?

hidmic: No, that is correct. What seemed redundant were the graph cache checks that were already there. Fine to leave as-is, failure conditions are much more unlikely now and will ultimately get caught by the check in `rmw_send_response`.

MiguelCompany:
@hidmic due to the union on the hashing function for the guid, I would also launch a CI for a 32bit platform if possible

Perhaps not a full CI, just a build so we can ensure the `static_assert` is not thrown

hidmic:
> I would also launch a CI for a 32bit platform if possible

I understand why and I'd agree, but I think we don't have 32 bits CI available. @nuclearsandwich @dirk-thomas to confirm.

ivanpauno:
> @hidmic due to the union on the hashing function for the guid, I would also launch a CI for a 32bit platform if possible
> 
> Perhaps not a full CI, just a build so we can ensure the `static_assert` is not thrown

I didn't know that `GUID_t` wasn't hashable, you could've ignored [my comment](https://github.com/ros2/rmw_fastrtps/pull/390#discussion_r430650118) to avoid adding the hashing function.

MiguelCompany:
> I didn't know that `GUID_t` wasn't hashable, you could've ignored [my comment](https://github.com/ros2/rmw_fastrtps/pull/390#discussion_r430650118) to avoid adding the hashing function.

I prefer addressing optimizations at the same time as some functionality is added. I have taken note to add a standard hash function to `GUID_t` on a future release of Fast RTPS (perhaps based on the one on this PR).

I can remove the hashing function and change the collection back to `std::set`, adding a `// TODO(MiguelCompany) change this to std::unordered_set when GUID_t becomes hashable`

ivanpauno:
@MiguelCompany I think it's ok to keep the hashing function here, whatever you prefer.
We're currently not targeting any 32bit platform AFAIK, so we can't test that case.

MiguelCompany:
@ivanpauno +1 to keep it here. I have put the code from `guid_utils.hpp` and the `GUID_t` code from Fast RTPS and compiled it on `gcc -m32` and the `static_assert` did not throw. The same when compiling with an ARM32 compiler `arm-linux-gnueabihf-g++`

So I think we're good to go

hidmic:
CI test failures are unrelated, this is ready to go.

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
```diff
@@ -104,6 +104,28 @@ __rmw_send_response(
...
+  // with the GUID of the response reader, so we can wait here for it to be matched to
+  // the server response writer. In the future, this should be done with the mechanism
+  // explained on OMG DDS-RPC 1.0 spec under section 7.6.2 (Enhanced Service Mapping)
+
+  // According to the list of possible entity kinds in section 9.3.1.2 of RTPS
+  // readers will have this bit on, while writers will not. We use this to know
+  // if the related guid is the request writer or the response reader.
+  constexpr uint8_t entity_id_is_reader_bit = 0x04;
+  const eprosima::fastrtps::rtps::GUID_t & related_guid =
+    wparams.related_sample_identity().writer_guid();
```
ivanpauno, line 118: isn't line 101 overriding this?
shouldn't it be checked before it's overriden?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
```diff
@@ -88,7 +89,9 @@ class ClientListener : public eprosima::fastrtps::SubscriberListener
       if (eprosima::fastrtps::rtps::ALIVE == response.sample_info_.sampleKind) {
         response.sample_identity_ = response.sample_info_.related_sample_identity;
 
-        if (response.sample_identity_.writer_guid() == info_->writer_guid_) {
+        if (response.sample_identity_.writer_guid() == info_->reader_guid_ ||
+          response.sample_identity_.writer_guid() == info_->writer_guid_)
```
ivanpauno, line 93: IIUC, this will always be the writer guid here (it's one thing or the other, but both aren't allowed).

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
MiguelCompany: Line 101 is just translating from the implementation-agnostic parameter `request_header` into Fast RTPS `wparams`.

I'm getting a const ref here so the check does not become a huge line.

We are then checking if the request came from a client with this PR merged, and only in that case we wait for the response writer to be matched against the corresponding reader.

nuclearsandwich:
> I understand why and I'd agree, but I think we don't have 32 bits CI available.

armhf CI is unavailable right now due to issues with kernel/userland mismatch for 32-bit userlands on our CI. You can build and run locally using the osrf/ubuntu_armhf:focal docker image to get a test via QEMU emulation.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
MiguelCompany: The server may not have the code in this PR, and will therefore answer with the writer guid. This makes the client backwards compatible

jacobperron:
Given that we're scheduled to release this week, I'd be more comfortable holding this PR for the first patch release to give us more time for testing. This means holding until the end of this week.

rmw_fastrtps_shared_cpp/src/rmw_response.cpp
ivanpauno: I misread the patch. LGTM!

ivanpauno:
> Given that we're scheduled to release this week, I'd be more comfortable holding this PR for the first patch release to give us more time for testing. This means holding until the end of this week.

If we do this change in a patch release and somebody updates the machine that has the client side but not the one that has the server side, communication will stop working (@MiguelCompany am I correct?).

IMO, this can't be backported, it's now or no fix for Foxy.

hidmic:
> If we do this change in a patch release and somebody updates the machine that has the client side but not the one that has the server side, communication will stop working (@MiguelCompany am I correct?).

@ivanpauno this patch is backwards compatible IIUC

ivanpauno:
> @ivanpauno this patch is backwards compatible IIUC

The server side for sure, but not the client side.

ivanpauno:
After talking offline with @hidmic this looks backward compatible to me too.

SteveMacenski:
@jacobperron its been a week, would it be OK to merge now?

hidmic:
Going in.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
```diff
@@ -159,4 +169,49 @@ class ServiceListener : public eprosima::fastrtps::SubscriberListener
...
+  ServicePubListener() = default;
+
+  template<class Rep, class Period>
+  bool wait_for_subscription(
+    const eprosima::fastrtps::rtps::GUID_t & guid,
+    const std::chrono::duration<Rep, Period> & rel_time)
+  {
+    auto guid_is_present = [this, guid]() -> bool
+      {
+        return subscriptions_.find(guid) != subscriptions_.end();
```
dirk-thomas, line 184: `clang` warns about this line: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/483/gcc/folder.862079900/

> reading variable 'subscriptions_' requires holding mutex 'mutex_' [-Wthread-safety-analysis] return subscriptions_.find(guid) != subscriptions_.end();

ros-discourse:

https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2020-07-23/15570/2

ros-discourse:

https://discourse.ros.org/t/new-packages-and-patch-release-for-ros-2-foxy-fitzroy-2020-08-07/15818/1

PR #396, opened 2020-06-22
Finalize context if and only if it's shutdown.

hidmic:
This pull request ensures `rmw_context_fini()` fails on a not shutdown context, as its API contract establishes (see https://github.com/ros2/rmw/pull/242).


hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11240/) (canary build)

PR #397, opened 2020-06-22
Finalize init options in context.

hidmic:
Avoid leaking options when finalizing contexts.


hidmic:
This has been superseded by #401.

PR #399, opened 2020-06-23
Ensure compliant init options API implementations.

hidmic:
Complying with https://github.com/ros2/rmw/pull/244.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_init.cpp
@@ -103,11 +103,9 @@ rmw_init_options_fini(const char * identifier, rmw_init_options_t * init_options
   rcutils_allocator_t * allocator = &init_options->allocator;
   RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
 
+  allocator->deallocate(init_options->enclave, allocator->state);
   rmw_ret_t ret = rmw_security_options_fini(&init_options->security_options, allocator);
-  if (RMW_RET_OK == ret) {
-    allocator->deallocate(init_options->enclave, allocator->state);
-    *init_options = rmw_get_zero_initialized_init_options();
-  }
+  *init_options = rmw_get_zero_initialized_init_options();
   return ret;
 }
 
--- a/rmw_fastrtps_shared_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_init.cpp
@@ -29,6 +29,7 @@ rmw_ret_t
 rmw_init_options_init(
   const char * identifier, rmw_init_options_t * init_options, rcutils_allocator_t allocator)
 {
+  assert(identifier != NULL);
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11262/)

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
```diff
@@ -62,39 +68,47 @@ rmw_init_options_copy(
     return RMW_RET_INVALID_ARGUMENT;
   }
```
ivanpauno, line 69: without line 66-69, this function could perfectly work with a non-zero initialized `dst` AFAIS

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
```diff
@@ -62,39 +68,47 @@ rmw_init_options_copy(
...
-  allocator.deallocate(init_options->enclave, allocator.state);
-  rmw_security_options_fini(&init_options->security_options, &allocator);
-  *init_options = rmw_get_zero_initialized_init_options();
-  return RMW_RET_OK;
+  rcutils_allocator_t * allocator = &init_options->allocator;
+  RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
+
+  rmw_ret_t ret = rmw_security_options_fini(&init_options->security_options, allocator);
+  if (RMW_RET_OK == ret) {
+    allocator->deallocate(init_options->enclave, allocator->state);
```
ivanpauno, line 108: should this happen if security options failed to be deallocated?
maybe:
```cpp
allocator->deallocate(init_options->enclave, allocator->state);
rmw_ret_t ret = rmw_security_options_fini(&init_options->security_options, allocator);
if (RMW_RET_OK == ret) {
  *init_options = rmw_get_zero_initialized_init_options();
}
return ret;
```

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
hidmic: I know, I prepared it for it. But it's lacking a finalization halfway through.

We can discuss guarantees later on and apply changes as needed.

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
hidmic: So, this is something I wasn't clear about in the API contract. I'm actively trying to ensure a failed finalization doesn't affect state. Alternatively, we could take the finalize-as-much-as-you-can approach AND prepare finalization functions for multiple calls on partially finalized state. WDYT?

hidmic:
One last sanity CI run after API contract updates:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11281/)

rmw_fastrtps_shared_cpp/src/rmw_init.cpp
hidmic: Alright, fini API contract has been updated.

PR #403, opened 2020-06-26
Update Quality Declaration to QL3.

hidmic:
Precisely what the title says. I may or may not update the dependencies' listing based on https://github.com/ros2/rosidl_typesupport/pull/77#discussion_r445662985 outcome.


hidmic:
Ok, considering that discussion is still on in https://github.com/ros2/rosidl_typesupport/pull/77#discussion_r445662985, I'll proceed and merge this.

PR #410, opened 2020-07-17
Set context actual domain id

ivanpauno:
See https://github.com/ros2/rcl/pull/718

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -100,6 +100,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eprosima_fastrtps_identifier;
+  context->actual_domain_id =
+    RMW_DEFAULT_DOMAIN_ID == options->domain_id ? 0uL : options->domain_id;
 
   context->impl = new (std::nothrow) rmw_context_impl_t();
   if (nullptr == context->impl) {
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -93,10 +93,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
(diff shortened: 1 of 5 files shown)


rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -166,11 +166,9 @@ rmw_fastrtps_shared_cpp::create_participant(
 
   // No custom handling of RMW_DEFAULT_DOMAIN_ID. Simply use a reasonable domain id.
 #if FASTRTPS_VERSION_MAJOR < 2
-  participantAttrs.rtps.builtin.domainId =
-    static_cast<uint32_t>(domain_id != RMW_DEFAULT_DOMAIN_ID ? domain_id : 0u);
+  participantAttrs.rtps.builtin.domainId = domain_id;
```
hidmic, line 169: @ivanpauno here and below, consider keeping those `static_cast`s.

ivanpauno: Good point, I didn't see them.

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -93,13 +93,13 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
-  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
```
hidmic, line 97: @ivanpauno also unrelated, and can be backtracked to the discussion that took place about UB in `rmw_init()`.

rmw_fastrtps_dynamic_cpp/src/rmw_init.cpp
```diff
@@ -93,13 +93,13 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
-  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
```
hidmic, line 97: @ivanpauno same as above.

rmw_fastrtps_cpp/src/rmw_init.cpp
ivanpauno: See https://github.com/ros2/rmw_connext/pull/443#discussion_r458314262

PR #409, opened 2020-07-15
Add missing thread-safety annotation in ServicePubListener

hidmic:
Addresses https://github.com/ros2/rmw_fastrtps/pull/390#discussion_r455136217. Looks like thread safety analysis isn't smart enough.

PR #412, opened 2020-07-21
Make service wait for response reader (#390)

jacobperron:
Backport #390 to Foxy.

Note, this breaks ABI compatibility, but only below the RMW layer. IMO, the breakage is acceptable considering the improvement to service discovery.

I'm still checking if this change is backports compatible over the wire.

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -192,6 +192,7 @@ rmw_create_client(
   }
 
   info->writer_guid_ = info->request_publisher_->getGuid();
+  info->reader_guid_ = info->response_subscriber_->getGuid();
 
   rmw_client = rmw_client_allocate();
   if (!rmw_client) {
--- a/rmw_fastrtps_cpp/src/rmw_service.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_service.cpp
@@ -192,8 +192,9 @@ rmw_create_service(
     RMW_SET_ERROR_MSG("failed to get datawriter qos");
     goto fail;
   }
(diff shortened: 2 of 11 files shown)


jacobperron:
> - A client of the new version can make requests to a service of the old version.
> - A client of the old version can make requests to a service of the new version.
> - Service and clients of the old version are listed correctly by a ros2cli tool of the new version.
> - Service and clients of the new version are listed correctly by a ros2cli tool of the old version.

All of these checks LGTM; I did checks on my local machine and between two hosts over a VPN.

jacobperron:
Note, I've also backported #409 which fixes a compiler warning introduced by this change (https://github.com/ros2/rmw_fastrtps/pull/413)

MiguelCompany:
> We should comment the breakage in the release notes, clarifying the it only affects users making use of FastRTPS native handles.

This is only affecting rmw, so users making use of native handles will NOT be affected, right?

jacobperron:
> This is only affecting rmw, so users making use of native handles will NOT be affected, right?

I think what @ivanpauno meant, is it only affects users making use of the handles in rmw_fastrtps (instead of the generic `rmw` structs). IIUC, only people using the structs `CustomClientInfo` and/or `CustomServiceInfo` will be affected. Does that sound right?

MiguelCompany:
> Does that sound right?

Perfect!

ros-discourse:

https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2020-07-23/15570/2

ros-discourse:

https://discourse.ros.org/t/new-packages-and-patch-release-for-ros-2-foxy-fitzroy-2020-08-07/15818/1

PR #414, opened 2020-07-22
Ensure compliant publisher API.

hidmic:
Precisely what the title says. Connected to https://github.com/ros2/rmw/pull/252.

### DIFF
--- a/rmw_fastrtps_cpp/src/publisher.cpp
+++ b/rmw_fastrtps_cpp/src/publisher.cpp
@@ -21,6 +21,9 @@
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
+#include "rmw/validate_full_topic_name.h"
+
+#include "rcpputils/scope_exit.hpp"
 
 #include "rmw_fastrtps_shared_cpp/create_rmw_gid.hpp"
 #include "rmw_fastrtps_shared_cpp/custom_participant_info.hpp"
@@ -49,20 +52,27 @@ rmw_fastrtps_cpp::create_publisher(
   bool keyed,
   bool create_publisher_listener)
 {
-  if (!participant_info) {
-    RMW_SET_ERROR_MSG("participant_info is null");
-    return nullptr;
+  RMW_CHECK_ARGUMENT_FOR_NULL(participant_info, nullptr);
+  RMW_CHECK_ARGUMENT_FOR_NULL(type_supports, nullptr);
+  RMW_CHECK_ARGUMENT_FOR_NULL(topic_name, nullptr);
+  RMW_CHECK_ARGUMENT_FOR_NULL(qos_policies, nullptr);
+  if (!qos_policies->avoid_ros_namespace_conventions) {
+    int validation_result = RMW_TOPIC_VALID;
+    rmw_ret_t ret = rmw_validate_full_topic_name(topic_name, &validation_result, nullptr);
+    if (RMW_RET_OK != ret) {
+      return nullptr;
+    }
+    if (RMW_TOPIC_VALID != validation_result) {
+      const char * reason = rmw_full_topic_name_validation_result_string(validation_result);
+      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("invalid topic name: %s", reason);
+      return nullptr;
+    }
   }
-  if (!topic_name || strlen(topic_name) == 0) {
-    RMW_SET_ERROR_MSG("publisher topic is null or empty string");
-    return nullptr;
-  }
-  if (!qos_policies) {
-    RMW_SET_ERROR_MSG("qos_policies is null");
-    return nullptr;
-  }
-  if (!publisher_options) {
-    RMW_SET_ERROR_MSG("publisher_options is null");
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher_options, nullptr);
+
+  Participant * participant = participant_info->participant;
+  if (!participant) {
+    RMW_SET_ERROR_MSG("participant handle is null");
     return nullptr;
   }
 
@@ -93,7 +103,14 @@ rmw_fastrtps_cpp::create_publisher(
     RMW_SET_ERROR_MSG("failed to allocate CustomPublisherInfo");
     return nullptr;
   }
-
+  auto cleanup_info = rcpputils::make_scope_exit(
+    [info, participant]() {
+      if (info->type_support_) {
+        _unregister_type(participant, info->type_support_);
+      }
+      delete info->listener_;
+      delete info;
+    });
   info->typesupport_identifier_ = type_support->typesupport_identifier;
   info->type_support_impl_ = type_support->data;
 
@@ -101,15 +118,15 @@ rmw_fastrtps_cpp::create_publisher(
   std::string type_name = _create_type_name(callbacks);
   if (
     !Domain::getRegisteredType(
-      participant_info->participant, type_name.c_str(),
+      participant, type_name.c_str(),
       reinterpret_cast<TopicDataType **>(&info->type_support_)))
   {
     info->type_support_ = new (std::nothrow) MessageTypeSupport_cpp(callbacks);
     if (!info->type_support_) {
-      RMW_SET_ERROR_MSG("Failed to allocate MessageTypeSupport");
-      goto fail;
+      RMW_SET_ERROR_MSG("failed to allocate MessageTypeSupport");
+      return nullptr;
     }
-    _register_type(participant_info->participant, info->type_support_);
+    _register_type(participant, info->type_support_);
   }
 
   if (!participant_info->leave_middleware_default_qos) {
@@ -124,8 +141,7 @@ rmw_fastrtps_cpp::create_publisher(
   publisherParam.topic.topicName = _create_topic_name(qos_policies, ros_topic_prefix, topic_name);
 
   if (!get_datawriter_qos(*qos_policies, publisherParam)) {
-    RMW_SET_ERROR_MSG("failed to get datawriter qos");
-    goto fail;
+    return nullptr;
   }
 
   info->listener_ = nullptr;
@@ -133,17 +149,17 @@ rmw_fastrtps_cpp::create_publisher(
     info->listener_ = new (std::nothrow) PubListener(info);
     if (!info->listener_) {
       RMW_SET_ERROR_MSG("create_publisher() could not create publisher listener");
-      goto fail;
+      return nullptr;
     }
   }
 
   info->publisher_ = Domain::createPublisher(
-    participant_info->participant,
+    participant,
     publisherParam,
     info->listener_);
   if (!info->publisher_) {
     RMW_SET_ERROR_MSG("create_publisher() could not create publisher");
-    goto fail;
+    return nullptr;
   }
 
   info->publisher_gid = rmw_fastrtps_shared_cpp::create_rmw_gid(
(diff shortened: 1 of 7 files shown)


rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -101,8 +100,18 @@ rmw_create_publisher(
...
       nullptr);
     if (RMW_RET_OK != rmw_ret) {
-      rmw_fastrtps_shared_cpp::__rmw_destroy_publisher(
+      rmw_error_state_t error_state = *rmw_get_error_state();
+      rmw_reset_error();
+      static_cast<void>(common_context->graph_cache.dissociate_writer(
+        info->publisher_gid, common_context->gid, node->name, node->namespace_));
+      rmw_ret = rmw_fastrtps_shared_cpp::__rmw_destroy_publisher(
         eprosima_fastrtps_identifier, node, publisher);
+      if (RMW_RET_OK != rmw_ret) {
```
hidmic, line 109: I don't like having to do this to ensure the first error that occurs is the one that pops up and that no subsequent error goes unnoticed.

If we had a way to tell `rcutils` that we're in error handling or finalization/cleanup mode so that it doesn't complain about error states being overwritten (and prints the new one to `stderr`, or aborts the program, or whatever), this code would simpler (and a bit more neat). What do reviewers think?

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -164,6 +173,19 @@ rmw_return_loaned_message_from_publisher(
...
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    node,
+    node->implementation_identifier,
+    eprosima_fastrtps_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
+    publisher->implementation_identifier,
+    eprosima_fastrtps_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
```
ivanpauno, line 187: why doing this checks here.
If they are done in `__rmw_destroy_publisher`, both rmw_fastrtps_cpp and rmw_fastrtps_dynamic_cpp share the same code.

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -49,20 +52,27 @@ rmw_fastrtps_cpp::create_publisher(
...
+      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("invalid topic name: %s", reason);
+      return nullptr;
+    }
   }
-  if (!publisher_options) {
-    RMW_SET_ERROR_MSG("publisher_options is null");
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher_options, nullptr);
+
+  Participant * participant = participant_info->participant;
+  if (!participant) {
```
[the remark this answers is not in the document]
hidmic: Good catch.

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -124,26 +141,25 @@ rmw_fastrtps_cpp::create_publisher(
   publisherParam.topic.topicName = _create_topic_name(qos_policies, ros_topic_prefix, topic_name);
 
   if (!get_datawriter_qos(*qos_policies, publisherParam)) {
-    RMW_SET_ERROR_MSG("failed to get datawriter qos");
```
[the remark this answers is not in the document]
hidmic: `get_datawriter_qos` already does (see [here](https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_shared_cpp/src/qos.cpp)).

rmw_fastrtps_cpp/src/rmw_publisher.cpp
hidmic: That is true, but `rmw_fastrtps_dynamic_cpp` has to do these checks earlier (see [here](https://github.com/ros2/rmw_fastrtps/blob/ece181fa17f3cbcba0bbfccfbd1491b79d972922/rmw_fastrtps_dynamic_cpp/src/rmw_publisher.cpp#L170-L187)). Between having duplicate checks in `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp`, or checking the same thing twice in run-time for `rmw_fastrtps_dynamic_cpp`, I chose the former.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
hidmic: That'd be simpler, yeah, but I think we would be trading usability in the process. To the point it makes me wonder what 's the point of returning an generic error code vs. not returning anything.

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -93,23 +103,30 @@ rmw_fastrtps_cpp::create_publisher(
     RMW_SET_ERROR_MSG("failed to allocate CustomPublisherInfo");
     return nullptr;
   }
-
+  auto cleanup_info = rcpputils::make_scope_exit(
+    [info, participant]() {
+      if (info->type_support_) {
+        _unregister_type(participant, info->type_support_);
+      }
+      delete info->listener_;
```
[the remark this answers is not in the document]
clalancette: Agreed.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
clalancette: I agree that this is a bit unfortunate.  The big problem is that you have to be careful on *every* error path to do the checking and not overwrite the error.  It will always be a game of whack-a-mole (or we have to have really good tests).

> That'd be simpler, yeah, but I think we would be trading usability in the process. To the point it makes me wonder what 's the point of returning an generic error code vs. not returning anything.

To be fair, the generic error code at least distinguishes between "succeeded" and "failed".  The question on whether the individual errors are worthwhile comes down to whether the calling code can do anything reasonable in response to the error.  If that caller can do something specific in response to some errors, then it makes sense to have individual errors.  If the caller can never do anything about the situation, the generic error is enough.

The problem in my mind with the `rcutils` flag is that it is not generic enough; you also need that flag at the rcl layer, rmw layer, and potentially the rclcpp/rclpy layers as well.

Overall, I think the error situation isn't perfect.  But I also think it is a large topic, and we don't necessarily have to solve it here.  I'd say that we continue with this PR assuming the current situation.  We can open up a separate issue and potentially have a meeting where we talk about the broader issue of errors.  Does that sound reasonable?

rmw_fastrtps_shared_cpp/src/publisher.cpp
```diff
@@ -35,33 +35,17 @@ destroy_publisher(
...
-  if (publisher->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("publisher handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
-  if (!participant_info) {
-    RMW_SET_ERROR_MSG("participant_info is null");
-    return RMW_RET_ERROR;
-  }
+  assert(publisher->implementation_identifier == identifier);
+  static_cast<void>(identifier);
```
clalancette, line 39: In my view, there are 2 cases here:

* `rmw_fastrtps_shared_cpp` is considered a private implementation detail of `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp`.  If that is the case, then we don't need to check the identifier here at all since we are checking at the "public" API layer before calling this.  We could also not pass the identifier into this method.
* `rmw_fastrtps_shared_cpp` is considered a public library.  In that case, we should keep the checks in place and return the error message (leaving the code that was here, or maybe using the macro instead).

The same goes for the asserts below.

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -49,22 +52,26 @@ rmw_fastrtps_cpp::create_publisher(
...
-  if (!qos_policies) {
-    RMW_SET_ERROR_MSG("qos_policies is null");
-    return nullptr;
-  }
-  if (!publisher_options) {
-    RMW_SET_ERROR_MSG("publisher_options is null");
-    return nullptr;
+  RMW_CHECK_ARGUMENT_FOR_NULL(participant_info, nullptr);
+  RMW_CHECK_ARGUMENT_FOR_NULL(type_supports, nullptr);
+  RMW_CHECK_ARGUMENT_FOR_NULL(topic_name, nullptr);
```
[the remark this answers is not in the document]
hidmic: Yes, we should for the non-ROS use case. I dropped it because topic validation would caught it, but that's not happening for native publishers.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
hidmic: > To be fair, the generic error code at least distinguishes between "succeeded" and "failed".

True, but we don't need specific return codes for that.

> The problem in my mind with the rcutils flag is that it is not generic enough; you also need that flag at the rcl layer, rmw layer, and potentially the rclcpp/rclpy layers as well.

Why? It's the same flag everywhere. It doesn't have to be a flag either, it's just that building anything bigger than this seems overkill.

> We can open up a separate issue and potentially have a meeting where we talk about the broader issue of errors. Does that sound reasonable?

Fair enough. Let's roll as-is.

rmw_fastrtps_shared_cpp/src/publisher.cpp
hidmic: So, I'll being saying that we don't consistently use either approach. It's currently a bit of both.

Considering that `rmw_fastrtps_shared_cpp` is not a standalone `rmw` implementation but the parts of one, I'm inclined towards using preconditions as much as it is possible i.e. to assert instead of checking. But sometimes, some checks can be pushed down to `rmw_fastrtps_shared_cpp` to avoid duplication. All this essentially couples all three packages' implementations for the sake of run-time efficiency and code reuse.

If we don't like that mix and match, that's fine. We can settle for either: all preconditions OR all checks in `rmw_fastrtps_shared_cpp`. Personally, I'd stick to all preconditions.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
clalancette: > Why? It's the same flag everywhere. It doesn't have to be a flag either, it's just that building anything bigger than this seems overkill.

I'm thinking of this situation:

```
int rcl_do_thing() {
  rcl_subthing_init();  // Succeeds
  if (rcutils_do_something() != OK) {
    // Failed, rcutils_do_something() set error code/message
    rcl_subthing_cleanup();  // Also failed, reset error code/message
    return NOT_OK;
  }
}
```

But its possible that I don't have enough context on how the error handling works.

hidmic: So, all layers use the same thread local error state and API that's defined in `rcutils` but through aliases. In your example, `rcl_do_thing` would enter this error handling mode for the extent of that `if` block statement. In a way, it's a poor man's try-catch block. 

But I see now that arbitrarily nested error handling may occur (a try-catch in the catch block of a try-catch in the catch block of...). We would have to make it reentrant and a boolean flag isn't enough. An integer count could be used instead.

rmw_fastrtps_shared_cpp/src/publisher.cpp
ivanpauno: > Considering that rmw_fastrtps_shared_cpp is not a standalone rmw implementation but the pieces of one, I'm inclined towards using preconditions as much as it is possible i.e. to assert instead of checking. But sometimes, some checks can be pushed down to rmw_fastrtps_shared_cpp to avoid code duplication.

:+1:

rmw_fastrtps_cpp/src/publisher.cpp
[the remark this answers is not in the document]
hidmic: Looking at `CustomPublisherInfo`, I'd expect the default constructor to value-initialize each of its members (which for integral types such as pointers that is zero-initialization).

hidmic:
CI up `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11630/)

rmw_fastrtps_cpp/src/publisher.cpp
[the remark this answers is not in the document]
ivanpauno: > Looking at `CustomPublisherInfo`, I'd expect the default constructor to value-initialize each of its members (which for integral types such as pointers that is zero-initialization). So I think we're covered.

The default constructor doesn't do that for built-in types.
AFAIK, it does nothing.

If aggregate initialization or value initialization is used, that would zero initialize built-in data members.

---

I would add a `= nullptr` in the struct definition, so the default is explicit and we don't have to double check what kind of initialization is being used.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
```diff
@@ -37,25 +37,11 @@ __rmw_destroy_publisher(
...
-
-  if (!publisher) {
-    RMW_SET_ERROR_MSG("publisher handle is null");
-    return RMW_RET_ERROR;
-  }
-  if (publisher->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("publisher handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
+  assert(node->implementation_identifier == identifier);
```
ivanpauno, line 40: I think you need to add:

```cpp
(void)identifier;
```

If you want to avoid a warning in release mode.

rmw_fastrtps_cpp/src/publisher.cpp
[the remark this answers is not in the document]
ivanpauno: This was solved in https://github.com/ros2/rmw_fastrtps/pull/414/commits/66bf0688e27b2d9bfe9ebed7411f6fd575b672d0.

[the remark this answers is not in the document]
hidmic: > If aggregate initialization or value initialization is used, that would zero initialize built-in data members.

Which is the case here. But fair enough, explicit is often better than implicit.

rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp
hidmic: It's used below.

hidmic:
`valgrind` shows no `new` leak directly or indirectly bound to this PR. It does show a leak in the node listener thread. We can follow-up with it in a separate PR.

PR #416, opened 2020-07-29
Unregister typesupport rather than delete

brawner:
There was a use-after-free that I discovered while adding fault injection tests to rcl_action over at https://github.com/ros2/rcl/pull/730. Somehow rcl_action's typesupport for its client and subscription were being duplicated. When it was cleaned up during this fail block in create_subscription, it would try to delete it again over in rmw_client_fini. Instead, I believe it needs to unregister it, which also deletes it, so that it won't get deleted twice. `rmw_client.cpp` already does this with unregister rather than deleting the pointer directly.

Before-and-after unit tests will have to wait until tomorrow.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/rmw_fastrtps_cpp/src/subscription.cpp
+++ b/rmw_fastrtps_cpp/src/subscription.cpp
@@ -164,7 +164,9 @@ create_subscription(
 
 fail:
   if (info != nullptr) {
-    delete info->type_support_;
+    if (info->type_support_ != nullptr) {
+      _unregister_type(participant, info->type_support_);
+    }
     delete info->listener_;
     delete info;
   }


brawner:
Builds and tests of the following packages `rcutils`, `rcl_logging_spdlog`, `rosidl_typesupport_c`, `rosidl_typesupport_cpp`, `rcl`, `rcl_action`, `rcl_lifecycle`, `rosidl_runtime_c`, `rmw_dds_common`, `rmw`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11916/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9709/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11850/)

hidmic:
@brawner I think https://github.com/ros2/rmw_fastrtps/pull/419 fixed this (among other things), no?

brawner:
I think you're correct.

hidmic:
@brawner are you keeping the branch for anything in particular?

brawner:
No, deletebot didn't want to delete it I guess.

PR #419, opened 2020-07-31
Ensure compliant subscription API.

hidmic:
Depends on https://github.com/ros2/rmw/pull/256.

### DIFF
--- a/rmw_fastrtps_cpp/src/subscription.cpp
+++ b/rmw_fastrtps_cpp/src/subscription.cpp
@@ -21,6 +21,9 @@
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
+#include "rmw/validate_full_topic_name.h"
+
+#include "rcpputils/scope_exit.hpp"
 
 #include "rmw_fastrtps_shared_cpp/custom_participant_info.hpp"
 #include "rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp"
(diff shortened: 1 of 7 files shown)


hidmic:
CI up to `test_rmw_implementation` and `rcl`, against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11736/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6817/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9574/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/11676/)

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -128,47 +148,50 @@ create_subscription(
...
   }
+  auto cleanup_subscription = rcpputils::make_scope_exit(
+    [rmw_subscription]() {
+      rmw_free(const_cast<char *>(rmw_subscription->topic_name));
+      rmw_subscription_free(rmw_subscription);
+    });
+
   rmw_subscription->implementation_identifier = eprosima_fastrtps_identifier;
   rmw_subscription->data = info;
-  rmw_subscription->topic_name = rcutils_strdup(topic_name, rcutils_get_default_allocator());
```
wjwwood, line 156: Why are you replacing the user of `rcutils_strdup`? It seems better than manually allocating and memcpy'ing.

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
```diff
@@ -39,25 +39,12 @@ __rmw_destroy_subscription(
...
-  if (!subscription) {
-    RMW_SET_ERROR_MSG("subscription handle is null");
-    return RMW_RET_ERROR;
-  }
-  if (subscription->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("subscription handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
+  assert(node->implementation_identifier == identifier);
+  assert(subscription->implementation_identifier == identifier);
```
wjwwood, line 43: An `assert` is not the same as the checks it is replacing... Why is this being changed?

rmw_fastrtps_cpp/src/subscription.cpp
hidmic: To at least be consistent with the rest of this implementation and others. We still owe ourselves a discussion on how allocations should be made.

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
hidmic: This was discussed elsewhere (https://github.com/ros2/rmw_fastrtps/pull/414#discussion_r460270498 comes to mind). Long story short, it doesn't make much sense to perform the same checks twice. So here those (redundant) checks in `rmw_fastrtps_shared_cpp` are replaced by assertions, and both `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp`  are expected to make proper use of this public but internal API.

wjwwood: Makes sense, but wouldn't it make more sense to check here once? Rather than checking in each and only asserting here? I'm guessing that wasn't possible.

rmw_fastrtps_cpp/src/subscription.cpp
wjwwood: Unless this unifies how we do string duplication everywhere in the package, then I think this is a step backwards and kind of unrelated to this pull request right? I just don't see why we're doing this in this pull request. The previous state wasn't wrong was it?

hidmic: It does for API that does not explicitly take an `rcutils_allocator_t` argument (see https://github.com/ros2/rmw_fastrtps/search?q=rcutils_strdup&unscoped_q=rcutils_strdup). 

But provided that this isn't part of the API spec, you're right in that it doesn't have to be in this PR. I'll remove it.

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
hidmic: We do that in many cases, but sometimes `rmw_fastrtps_cpp` or `rmw_fastrtps_dynamic_cpp` don't just forward the call to `rmw_fastrtps_shared_cpp` and thus validation has to happen there before doing anything vendor specific.

rmw_fastrtps_cpp/src/subscription.cpp
hidmic: See https://github.com/ros2/rmw_fastrtps/pull/419/commits/791a3308eb4aa92fc90d9b24ef60a24b636ce210

rmw_fastrtps_shared_cpp/src/subscription.cpp
```diff
@@ -44,35 +44,23 @@ destroy_subscription(
...
-  }
-  if (subscription->implementation_identifier != identifier) {
-    RMW_SET_ERROR_MSG("subscription handle not from this implementation");
-    return RMW_RET_ERROR;
-  }
-  if (!participant_info) {
-    RMW_SET_ERROR_MSG("participant_info is null");
-    return RMW_RET_ERROR;
-  }
+  assert(subscription->implementation_identifier == identifier);
```
dirk-thomas, line 47: @hidmic In non-debug builds the `identifier` variable is now unused - see new compiler warnings: https://ci.ros2.org/view/packaging/job/packaging_linux/1969/gcc/

jacobperron: Fixed in https://github.com/ros2/rmw_fastrtps/pull/422

PR #417, opened 2020-07-29
Fix trying to get topic data that was already removed.

iuhilnehc-ynos:
This PR is related to https://github.com/ros2/rclcpp/issues/1242

Signed-off-by: Chen.Lihui <lihui.chen@sony.com>

PR #421, opened 2020-08-06
[backport foxy] Fix trying to get topic data that was already removed

dirk-thomas:
Backport of #417.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
@@ -67,34 +67,23 @@ public:
   // SubscriberListener implementation
   void
   onSubscriptionMatched(
-    eprosima::fastrtps::Subscriber * /*sub*/, eprosima::fastrtps::rtps::MatchingInfo & info) final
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -142,7 +131,7 @@ class SubListener : public EventListenerInterface, public eprosima::fastrtps::Su
   }
 
   void
-  data_taken(eprosima::fastrtps::Subscriber * sub)
+  update_unread_count(eprosima::fastrtps::Subscriber * sub)
```
jacobperron, line 134: Is it possible to make this change in an ABI compatible way (e.g. leave `data_taken` and call `update_unread_count` from it, or avoid the rename entirely)?

dirk-thomas: Forwarding the request to @iuhilnehc-ynos @fujitatomoya @MiguelCompany which iterated on the original patch.

dirk-thomas: Since avoiding the renaming completely I reverted it in 648f9b0d159590c4a70adfeeb5f8aaf7bead24ca.

PR #443, opened 2020-09-23
Updated rmw_* return codes

ahcorde:
Updated return codes for unimplemented rmw_*_*_allocation functions. Related with this PR https://github.com/ros2/rmw_implementation/pull/137

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
@@ -90,7 +90,7 @@ namespace rmw_fastrtps_shared_cpp
 {
 rmw_ret_t
 __rmw_wait(
-  const char* identifier,
+  const char * identifier,
   rmw_subscriptions_t * subscriptions,
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
@@ -107,7 +107,8 @@ __rmw_wait(
 
   // If wait_set_info is ever nullptr, it can only mean one of three things:
   // - Wait set is invalid. Caller did not respect preconditions.
-  // - Implementation is logically broken. Definitely not something we want to treat as a normal error.
+  // - Implementation is logically broken. Definitely not something we want to treat as a normal
+  // error.
   // - Heap is corrupt.
   // In all three cases, it's better if this crashes soon enough.
   CustomWaitsetInfo * wait_set_info = static_cast<CustomWaitsetInfo *>(wait_set->data);
--- a/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
@@ -90,6 +90,7 @@ namespace rmw_fastrtps_shared_cpp
 {
 rmw_ret_t
 __rmw_wait(
+  const char* identifier,
   rmw_subscriptions_t * subscriptions,
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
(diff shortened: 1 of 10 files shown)


hidmic (approved):
LGTM, but before merging, would you mind opening a PR against `rmw` to add this error code to the API spec?

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -100,8 +101,13 @@ __rmw_wait(
 {
   if (!wait_set) {
     RMW_SET_ERROR_MSG("wait set handle is null");
-    return RMW_RET_ERROR;
+    return RMW_RET_INVALID_ARGUMENT;
```
hidmic, line 104: @ahcorde FYI `RMW_CHECK_ARGUMENT_FOR_NULL` is your friend here.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -100,8 +101,13 @@ __rmw_wait(
...
   }
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    wait set handle,
+    wait_set->implementation_identifier, identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION)
+
   CustomWaitsetInfo * wait_set_info = static_cast<CustomWaitsetInfo *>(wait_set->data);
   if (!wait_set_info) {
```
hidmic, line 112: @ahcorde I think we can remove this check. If `wait_set_info` is ever `nullptr`, it can only mean one of three things:

1. Wait set is invalid. Caller did not respect preconditions. We cannot check for pointer validity in general, checking for `nullptr` is near useless. 
2. Implementation is logically broken. Definitely not something we want to treat as a normal error.
3. Heap is corrupt.

In all three cases, it's better if this crashes soon enough.

hidmic: On that same note, [lines 118 to 125](https://github.com/ros2/rmw_fastrtps/pull/443/files#diff-6c98126e644ee46f5d86c7fffa3da911R118-R125) are nonsense. Those cannot ever be `nullptr` (see [CustomWaitSetInfo](https://github.com/ros2/rmw_fastrtps/blob/8f6a13c5995b13a8bd9a0d147851ab5754bdf104/rmw_fastrtps_shared_cpp/src/types/custom_wait_set_info.hpp#L21)).

rmw_fastrtps_shared_cpp/src/rmw_wait_set.cpp
```diff
@@ -80,7 +80,7 @@ __rmw_destroy_wait_set(const char * identifier, rmw_wait_set_t * wait_set)
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     wait set handle,
     wait_set->implementation_identifier, identifier,
-    return RMW_RET_ERROR)
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION)
 
   auto result = RMW_RET_OK;
   auto wait_set_info = static_cast<CustomWaitsetInfo *>(wait_set->data);
```
hidmic, line 86: @ahcorde I think this fini function needs a refactor. The two checks right below make no sense for the same reasons exposed above. And I'd go as far as to say that those if clauses below make no sense either -- if you got this far in the function and one of those pointers is `nullptr` something's very wrong.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12470/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7437/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10183/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12390/)

hidmic:
CI up to `test_rmw_implementation` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12476/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7443/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10189/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12395/)

PR #447, opened 2020-09-24
Improved return error in __rmw_destroy_client

ahcorde:
Check node var in __rmw_destroy_client and return an error if it's a nullptr


hidmic:
This may not be necessary once https://github.com/ros2/rmw_fastrtps/pull/445 goes in.

hidmic:
@ahcorde I think this patch is no longer necessary.

PR #457, opened 2020-09-30
Handle too large QoS queue depths. 

hidmic:
A bug I came across while looking for more lines to cover.

CI up to `rmw_fastrtps_shared_cpp` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7539/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10282/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12496/)

### DIFF
--- a/rmw_fastrtps_shared_cpp/test/test_rmw_qos_to_dds_attributes.cpp
+++ b/rmw_fastrtps_shared_cpp/test/test_rmw_qos_to_dds_attributes.cpp
@@ -108,10 +108,15 @@ TEST_F(GetDataReaderQoSTest, large_depth_conversion) {
   EXPECT_EQ(
     eprosima::fastrtps::KEEP_LAST_HISTORY_QOS,
     subscriber_attributes_.topic.historyQos.kind);
-  EXPECT_GE(depth, static_cast<size_t>(subscriber_attributes_.topic.historyQos.depth));
+  EXPECT_LE(depth, static_cast<size_t>(subscriber_attributes_.topic.historyQos.depth));
 
   using depth_type = decltype(subscriber_attributes_.topic.historyQos.depth);
-  size_t max_depth = static_cast<size_t>(std::numeric_limits<depth_type>::max());
+  constexpr size_t max_depth = static_cast<size_t>(std::numeric_limits<depth_type>::max());
+
+  qos_profile_.depth = max_depth;
+  EXPECT_TRUE(get_datareader_qos(qos_profile_, subscriber_attributes_));
+  EXPECT_LE(depth, static_cast<size_t>(subscriber_attributes_.topic.historyQos.depth));
+
   if (max_depth < std::numeric_limits<size_t>::max()) {
     qos_profile_.depth = max_depth + 1;
     EXPECT_FALSE(get_datareader_qos(qos_profile_, subscriber_attributes_));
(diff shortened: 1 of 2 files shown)


rmw_fastrtps_shared_cpp/test/test_rmw_qos_to_dds_attributes.cpp
```diff
@@ -94,7 +95,27 @@ TEST_F(GetDataReaderQoSTest, nominal_conversion) {
   EXPECT_EQ(
     eprosima::fastrtps::KEEP_LAST_HISTORY_QOS,
     subscriber_attributes_.topic.historyQos.kind);
-  EXPECT_EQ(10, subscriber_attributes_.topic.historyQos.depth);
+  EXPECT_GE(10, subscriber_attributes_.topic.historyQos.depth);
+}
+
+TEST_F(GetDataReaderQoSTest, large_depth_conversion) {
+  size_t depth = subscriber_attributes_.topic.historyQos.depth + 1;
```
brawner, line 102: What is the general type of depth here? Are you relying on integer overflowing?

brawner: Specifically, "signed integer overflowing"

hidmic: Type is `size_t` for `rmw` and `int32_t` for `fastrtps`. This isn't relying on overflow (which are UB in C, IIRC), simply ensuring the configured `depth` is larger than the one already configured. 

I don't think there's a practical risk of overflow here anyways, unless we start running ROS 2 on 16 bits architectures or eProsima starts using `std::numeric_limits<int32_t>::max()` as default depth.

brawner: Ok, I think I understand better what's being tested.

rmw_fastrtps_shared_cpp/test/test_rmw_qos_to_dds_attributes.cpp
```diff
@@ -94,7 +95,27 @@ TEST_F(GetDataReaderQoSTest, nominal_conversion) {
...
+
+  EXPECT_TRUE(get_datareader_qos(qos_profile_, subscriber_attributes_));
+
+  EXPECT_EQ(
+    eprosima::fastrtps::KEEP_LAST_HISTORY_QOS,
+    subscriber_attributes_.topic.historyQos.kind);
+  EXPECT_GE(depth, static_cast<size_t>(subscriber_attributes_.topic.historyQos.depth));
+
+  using depth_type = decltype(subscriber_attributes_.topic.historyQos.depth);
+  size_t max_depth = static_cast<size_t>(std::numeric_limits<depth_type>::max());
```
brawner, line 114: I think `max_depth` can be declared constexpr, which means the comparison in the if statement would be evaluated at compile time and removed if not needed. Alternatively, you could use `std::is_same<size_t, depth_type>::value` in the if for the same effect (if that has the same behavior you want).

Would it make sense to add an EXPECT_TRUE check for `qos_profile_.depth = max_depth`?

hidmic: Good call ! it also made me realize I had those `EXPECT_*` backwards. See https://github.com/ros2/rmw_fastrtps/pull/457/commits/9f24d76d5370e2f83a67cbec9aaed8da9351c2fb.

hidmic:
CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12573/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7541/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10284/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12498/)

brawner (approved):

PR #458, opened 2020-10-01
Ensure rmw_destroy_node() completes despite run-time errors.

hidmic:
Precisely what the title says. Destruction should never return early i.e. partially destruct an object. This patch ensures this is the case _and_ that no error gets lost. 

CI up to `rcl` and `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12585/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7552/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10297/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12510/)

Hopefully this will fix the flakes we've seen in https://github.com/ros2/rmw_implementation/pull/144 and others.

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_node.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_node.cpp
@@ -85,12 +85,34 @@ rmw_destroy_node(rmw_node_t * node)
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
 
   rmw_context_t * context = node->context;
-  rmw_ret_t ret = rmw_fastrtps_shared_cpp::__rmw_destroy_node(
+
+  rmw_ret_t ret = RMW_RET_OK;
+  rmw_error_state_t error_state;
+  rmw_ret_t inner_ret = rmw_fastrtps_shared_cpp::__rmw_destroy_node(
     eprosima_fastrtps_identifier, node);
   if (RMW_RET_OK != ret) {
-    return ret;
+    error_state = *rmw_get_error_state();
+    ret = inner_ret;
+    rmw_reset_error();
   }
-  return rmw_fastrtps_shared_cpp::decrement_context_impl_ref_count(context);
+
+  inner_ret = rmw_fastrtps_shared_cpp::decrement_context_impl_ref_count(context);
+  if (RMW_RET_OK != inner_ret) {
+    if (RMW_RET_OK != ret) {
+      RMW_SAFE_FWRITE_TO_STDERR(rmw_get_error_string().str);
+      RMW_SAFE_FWRITE_TO_STDERR(" during '" RCUTILS_STRINGIFY(__function__) "'\n");
+    } else {
+      error_state = *rmw_get_error_state();
+      ret = inner_ret;
+    }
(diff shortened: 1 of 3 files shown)


rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -85,12 +85,34 @@ rmw_destroy_node(rmw_node_t * node)
...
+    rmw_reset_error();
+  }
+
+  inner_ret = rmw_fastrtps_shared_cpp::decrement_context_impl_ref_count(context);
+  if (RMW_RET_OK != inner_ret) {
+    if (RMW_RET_OK != ret) {
+      RMW_SAFE_FWRITE_TO_STDERR(rmw_get_error_string().str);
+      RMW_SAFE_FWRITE_TO_STDERR(" during '" RCUTILS_STRINGIFY(__function__) "'\n");
+    } else {
+      error_state = *rmw_get_error_state();
```
brawner, line 105: Is there an easy way to combine the two error states should there be multiple failures?

rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -85,12 +85,34 @@ rmw_destroy_node(rmw_node_t * node)
...
   rmw_context_t * context = node->context;
-  rmw_ret_t ret = rmw_fastrtps_shared_cpp::__rmw_destroy_node(
+
+  rmw_ret_t ret = RMW_RET_OK;
+  rmw_error_state_t error_state;
+  rmw_ret_t inner_ret = rmw_fastrtps_shared_cpp::__rmw_destroy_node(
     eprosima_fastrtps_identifier, node);
   if (RMW_RET_OK != ret) {
-    return ret;
+    error_state = *rmw_get_error_state();
```
brawner, line 94: You don't want to log the error here as well?

brawner (commented):
Your PR title and description are not particularly explanatory. What is the issue, what does this fix do?

rmw_fastrtps_cpp/src/rmw_node.cpp
hidmic: There're no errors before this one, so we can put in the local error state w/o printing.

rmw_fastrtps_cpp/src/rmw_node.cpp
hidmic: As of today, there isn't. In general, we don't handle errors during error handling and/or finalization very well or at all. I opened an issue about this: https://github.com/ros2/rcutils/issues/269, to have an standard idiom for this.

For the moment, this makes sure the first error goes into the error state and any other errors that may follow do not get lost but do not obscure what is assumed to be the root cause of the error.

hidmic:
> Your PR title and description are not particularly explanatory. What is the issue, what does this fix do?

Good point. Updated.

brawner (approved):

hidmic:
Alright, thanks for the review @brawner ! Going in!

PR #467, opened 2020-10-21
Discriminate when the Client has gone from when the Client has not completely matched

JLBuenoLopez-eProsima:
This pull request deals with ros2/ros2#1024.

`RMW_RET_TIMEOUT` is only returned when the client reader has not matched after a wait of 100 ms (already introduced in #390).
With this workaround the case when the Client has gone before the Service has sent the response is dealt with.
In this case, `RMW_RET_OK` is returned to prevent the unhandled exception RCLError reported in the mentioned issue.

Bear in mind that `RMW_RET_TIMEOUT` is not handled upstream and will cause the RCLError exception to happen.
`rcl` should deal with this return type knowing that it means that the Client has not completely matched so a response could still be sent calling `rmw_send_response` again (if the matching is by then completed).

This workaround will be superseded by the implementation of #418.


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
```diff
@@ -62,6 +71,92 @@ typedef struct CustomServiceRequest
...
+    if (clients_endpoints_.find(guid) != clients_endpoints_.end()) {
+      // Wait for subscription
+      if (!wait_for_subscription(guid, std::chrono::milliseconds(100))) {
+        return client_present_t::MAYBE;
+      }
+    } else {
+      // Client has gone
+      return client_present_t::GONE;
+    }
+    return client_present_t::YES;
```
hidmic, line 142: @JLBuenoLopez-eProsima nit: consider
```suggestion
    if (clients_endpoints_.find(guid) == clients_endpoints_.end()) {
      // Client is gone
      return client_present_t::GONE;
    }  
    // Wait for subscription
    if (!wait_for_subscription(guid, std::chrono::milliseconds(100))) {
      return client_present_t::MAYBE;
    }
    return client_present_t::YES;
```
for readability.

hidmic (approved):
LGTM pending green CI. 

Thanks for taking care of this @JLBuenoLopez-eProsima. I completely forgot about https://github.com/ros2/ros2/issues/1017.

hidmic:
> Bear in mind that RMW_RET_TIMEOUT is not handled upstream and will cause the RCLError exception to happen.
rcl should deal with this return type knowing that it means that the Client has not completely matched so a response could still be sent calling rmw_send_response again (if the matching is by then completed).

That's right. `rmw_cyclonedds_cpp` has the same issue. It's just way less likely to occur than the client going away completely. At any rate, all `rmw` APIs are a potential points of failure. No upstream code should assume otherwise (and allow exceptions to terminate the program).

hidmic:
Also, @JLBuenoLopez-eProsima it'd be great to have a regression test for this in [`test_rmw_implementation`](https://github.com/ros2/rmw_implementation/tree/master/test_rmw_implementation). At least for the "client's gone before server replies" scenario.

JLBuenoLopez-eProsima:
> Also, @JLBuenoLopez-eProsima it'd be great to have a regression test for this in [`test_rmw_implementation`](https://github.com/ros2/rmw_implementation/tree/master/test_rmw_implementation). At least for the "client's gone before server replies" scenario.

@hidmic,

ros2/rmw_implementation#162 takes care of your suggestion

hidmic:
CI up to `test_rmw_implementation`, `test_rclcpp`, and `test_communication`, including https://github.com/ros2/rmw_implementation/pull/162:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7790/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10547/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12778/)

hidmic:
Alright, tests failing on Windows are known to be flaky. I'll merge. Thanks for the contribution @JLBuenoLopez-eProsima !

PR #469, opened 2020-10-23
<Foxy> Discriminate when the Client has gone from when the Client has not completely matched

JLBuenoLopez-eProsima:
This backports #467 to Foxy.

Note, this breaks ABI compatibility. The first workaround (#412) also had to deal with this problem. The breakage may be acceptable considering the improvement to service discovery (ros2/ros2#1024).

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -82,11 +82,7 @@ class ServicePubListener : public eprosima::fastrtps::PublisherListener
       rmw_fastrtps_shared_cpp::hash_fastrtps_guid>;
 
 public:
-  explicit ServicePubListener(CustomServiceInfo * info)
-  : info_(info)
-  {
-    (void)info_;
-  }
+  ServicePubListener() = default;
 
   void
   onPublicationMatched(
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -20,6 +20,7 @@
 #include <list>
 #include <mutex>
 #include <unordered_set>
+#include <unordered_map>
 
 #include "fastcdr/FastBuffer.h"
 
@@ -37,6 +38,14 @@
 
 class ServiceListener;
 class ServicePubListener;
+
+enum class client_present_t
+{
+  FAILURE,  // an error occurred when checking
+  MAYBE,    // reader not matched, writer still present
+  YES,      // reader matched
+  GONE      // neither reader nor writer
+};
 
 typedef struct CustomServiceInfo
 {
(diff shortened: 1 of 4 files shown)


hidmic:
@ros-pull-request-builder please retest this.

hidmic:
@ros-pull-request-builder retest this please.

hidmic (commented):
This looks OK to me, but I'll defer to @jacobperron the decision as to whether we can accept an ABI breakage into Foxy or not.

MiguelCompany:
@hidmic @jacobperron Please bear in mind that the ABI break is on package `rmw_fastrtps_shared_cpp`, which is never used by a ROS user directly.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
```diff
@@ -62,6 +71,91 @@ typedef struct CustomServiceRequest
...
+    std::unordered_set<eprosima::fastrtps::rtps::GUID_t,
+      rmw_fastrtps_shared_cpp::hash_fastrtps_guid>;
+  using clients_endpoints_map_t =
+    std::unordered_map<eprosima::fastrtps::rtps::GUID_t,
+      eprosima::fastrtps::rtps::GUID_t,
+      rmw_fastrtps_shared_cpp::hash_fastrtps_guid>;
+
+public:
+  explicit ServicePubListener(CustomServiceInfo * info)
+  : info_(info)
```
jacobperron, line 86: Maybe I'm missing something, but it doesn't look like `info_` is used anywhere. Do we need to make this API change or can we revert it?

jacobperron (commented):
An ABI change is one thing, but this is changing API in a public header. Although it is unlikely that external packages are using the affected type, I'm rather not risk releasing a source-incompatible change into Foxy, especially in a core package that many depend on, albeit transitively.

Perhaps we could modify the change to retain ABI (or at least API) compatibility?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
JLBuenoLopez-eProsima: Thanks @jacobperron!!

It is true! This change was introduced by a previous solution I tried and I forgot to revert it. It is no longer needed and I have reverted it in the last commit. I will create another pull request to revert it in master too.

JLBuenoLopez-eProsima:
#475 reverts the API break in master.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
```diff
@@ -62,6 +71,87 @@ typedef struct CustomServiceRequest
...
+
+  // Accesors
+  clients_endpoints_map_t & clients_endpoints()
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return clients_endpoints_;
+  }
+
+private:
+  CustomServiceInfo * info_;
```
jacobperron, line 148: We can remove this member, right?

JLBuenoLopez-eProsima: Right! Thanks!

jacobperron:
I've discussed this change offline when the ROS 2 team and we would rather not break ABI. I've proposed an alternative PR that attempts to backport this change in an ABI compatible way (see https://github.com/ros2/rmw_fastrtps/pull/479).

jacobperron:
Backported in #479, maintaining ABI compatibility.

PR #466, opened 2020-10-13
Update README to clarify the use of environment variable RMW_FASTRTPS_USE_QOS_FROM_XML

JLBuenoLopez-eProsima:
Explanation about how to use environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` can be clearer preventing misunderstandings and issues that some users have reported (#465)

PR #470, opened 2020-10-26
New environment variable to change easily the publication mode

JLBuenoLopez-eProsima:
Several issues are related to the default publication mode set in `rmw_fastrtps` which is `ASYNCHRONOUS` publish mode.
This configuration can be changed only if the user defines a XML file and also sets the environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` to 1. This could be cumbersome to the user and also, as the issues show, prone to errors.

This pull request uses a new environment variable `RMW_FASTRTPS_PUBLICATION_MODE` so the user can change easily the publishing mode without needing to define a XML file.
This new environment variable can be used also in other RMW implementations and the values that have been considered are the following (this is open to discussion and more values can be given to this new environment variable):
* `ASYNCHRONOUS`: asynchronous publishing mode (default option in `rmw_fastrtps`).
* `SYNCHRONOUS`: synchronous publishing mode.
* `AUTO`: leave the rmw implementation to configure this parameter. In the case of `rmw_fastrtps` this value will take the value defined in the XML file or, failing this, the default value given in Fast DDS (`SYNCHRONOUS`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,13 +49,15 @@ You have two ways of telling you ROS 2 application which XML to use:
 Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
 The admissible values are:
 * `ASYNCHRONOUS`: asynchronous publication mode.
-This implies that when the publisher calls the write operation, the data is copied in a queue and control is returned to the user thread before the data is actually sent.
-The asynchronous thread is in charge of consuming the queue and send the data to any matched reader.
+Setting this mode implies that when the publisher invoques the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
+A background thread (asynchronous thread) is in turn in charge of consuming the queue and sending the data to every matched reader.
 * `SYNCHRONOUS`: synchronous publication mode.
-This implies that the data is sent directly in the context of the user thread.
+Setting this mode implies that the data is sent directly within the context of the user thread.
+This entails that any blocking call occurring during the write operation would block the user thread, thus preventing the application with continuing its operation.
+It is important to note that this mode typically yields higher throughput rates at lower latencies, since the notification and context switching between threads is not present.
 * `AUTO`: let Fast DDS select the publication mode. This implies using the publication mode set in the XML file or, failing that, the default value set in Fast DDS (which currently is set to `SYNCHRONOUS`).
 
-By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS` and it is the publication mode in use if this environment variable is not set.
+By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`, which is the publication mode in use if this environment variable is not set.
 
 ## Example
 
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ You have two ways of telling you ROS 2 application which XML to use:
 1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
 2. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to your XML file.
 
+### Change publication mode
+
+Another way to change easily the publication mode is to use the environment variable `ROS_PUBLICATION_MODE`.
+The admissible values are:
+* `ASYNCHRONOUS`: asynchronous publication mode
+* `SYNCHRONOUS`: synchronous publication mode
+* `AUTO`: let the rmw implementation select the publication mode.
+
+By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`.
+Also, if you set the variable to `AUTO`, `rmw_fastrtps` will use the publication mode set in the XML file or, failing that, the default value set in Fast DDS (currently set to `SYNCHRONOUS`).
+
 ## Example
 
 The following example configures Fast DDS to publish synchronously, and to have a pre-allocated history that can be expanded whenever it gets filled.
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ You have two ways of telling you ROS 2 application which XML to use:
 
 ### Change publication mode
 
-Another way to change easily the publication mode is to use the environment variable `ROS_PUBLICATION_MODE`.
+Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
 The admissible values are:
 * `ASYNCHRONOUS`: asynchronous publication mode
 * `SYNCHRONOUS`: synchronous publication mode
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ You can however set it to `rmw_fastrtps_dynamic_cpp` using the environment varia
 * Publication mode: `ASYNCHRONOUS_PUBLISH_MODE`
 
 However, it is possible to fully configure Fast DDS (including the history memory policy and the publication mode) using an XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
-If you want to modify the history memory policy or the publication mode you must set environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` to 1 (it is set to 0 by default).
+Bear in mind that if you want to modify the history memory policy or the publication mode you must set environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` to 1 (it is set to 0 by default) besides defining the XML file.
 This tells `rmw_fastrtps` that it should override both the history memory policy and the publication mode using the XML.
 Bear in mind that if you set this environment variable but do not give a value to either of these policies, defaults will be used.
 Current Fast-DDS defaults are:
@@ -46,7 +46,9 @@ You have two ways of telling you ROS 2 application which XML to use:
 
 ### Change publication mode
 
-Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
+Another way to change easily the publication mode without the need of defining a XML file is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
+This variable has lower precedence than `RMW_FASTRTPS_USE_QOS_FROM_XML`.
+Therefore, it is only taken into account when `RMW_FASTRTPS_USE_QOS_FROM_XML` is not set (or given a value different than `1`).
 The admissible values are:
 * `ASYNCHRONOUS`: asynchronous publication mode.
 Setting this mode implies that when the publisher invokes the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ This entails that any blocking call occurring during the write operation would b
 It is important to note that this mode typically yields higher throughput rates at lower latencies, since the notification and context switching between threads is not present.
 * `AUTO`: let Fast DDS select the publication mode. This implies using the publication mode set in the XML file or, failing that, the default value set in Fast DDS (which currently is set to `SYNCHRONOUS`).
 
-By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`, which is the publication mode in use if this environment variable is not set.
+If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` behave as if it were set to `ASYNCHRONOUS`.
 
 ## Example
 
--- a/README.md
+++ b/README.md
@@ -48,12 +48,14 @@ You have two ways of telling you ROS 2 application which XML to use:
 
 Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
 The admissible values are:
-* `ASYNCHRONOUS`: asynchronous publication mode
-* `SYNCHRONOUS`: synchronous publication mode
-* `AUTO`: let the rmw implementation select the publication mode.
+* `ASYNCHRONOUS`: asynchronous publication mode.
+This implies that when the publisher calls the write operation, the data is copied in a queue and control is returned to the user thread before the data is actually sent.
+The asynchronous thread is in charge of consuming the queue and send the data to any matched reader.
+* `SYNCHRONOUS`: synchronous publication mode.
+This implies that the data is sent directly in the context of the user thread.
+* `AUTO`: let Fast DDS select the publication mode. This implies using the publication mode set in the XML file or, failing that, the default value set in Fast DDS (which currently is set to `SYNCHRONOUS`).
 
-By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`.
-Also, if you set the variable to `AUTO`, `rmw_fastrtps` will use the publication mode set in the XML file or, failing that, the default value set in Fast DDS (currently set to `SYNCHRONOUS`).
+By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS` and it is the publication mode in use if this environment variable is not set.
 
 ## Example
 
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ You have two ways of telling you ROS 2 application which XML to use:
 Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
 The admissible values are:
 * `ASYNCHRONOUS`: asynchronous publication mode.
-Setting this mode implies that when the publisher invoques the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
+Setting this mode implies that when the publisher invokes the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
 A background thread (asynchronous thread) is in turn in charge of consuming the queue and sending the data to every matched reader.
 * `SYNCHRONOUS`: synchronous publication mode.
 Setting this mode implies that the data is sent directly within the context of the user thread.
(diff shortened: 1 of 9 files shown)


JLBuenoLopez-eProsima:
@hidmic,

If you agree that it is interesting to use this environment variable in other rmw implementations, I assume that I should document this new variable in some other more general repository. I would be grateful if you could tell me where.

README.md
```diff
@@ -44,6 +44,17 @@ You have two ways of telling you ROS 2 application which XML to use:
 1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
 2. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to your XML file.
 
+### Change publication mode
+
+Another way to change easily the publication mode is to use the environment variable `ROS_PUBLICATION_MODE`.
```
hidmic, line 49: @JLBuenoLopez-eProsima I'm somewhat inclined to stick to `RMW` for a prefix: 
```suggestion
Another way to change easily the publication mode is to use the environment variable `RMW_PUBLICATION_MODE`.
```

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -194,6 +197,20 @@ rmw_fastrtps_shared_cpp::create_participant(
   }
   if (env_value != nullptr) {
     leave_middleware_default_qos = strcmp(env_value, "1") == 0;
+  } else if (env_value == nullptr || !leave_middleware_default_qos) {
+    error_str = rcutils_get_env("ROS_PUBLICATION_MODE", &env_value);
+    if (error_str != NULL) {
+      RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_shared_cpp", "Error getting env var: %s\n", error_str);
```
hidmic, line 203: @JLBuenoLopez-eProsima consider using `RMW_SET_ERROR` instead of logging.

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -194,6 +197,20 @@ rmw_fastrtps_shared_cpp::create_participant(
...
+      return nullptr;
+    }
+    if (env_value != nullptr) {
+      // Synchronous publishing
+      if (strcmp(env_value, "SYNCHRONOUS") == 0) {
+        publishing_mode = publishing_mode_t::SYNCHRONOUS;
+      } else if (strcmp(env_value, "AUTO") == 0) {
+        publishing_mode = publishing_mode_t::AUTO;
+      }
```
hidmic, line 212: @JLBuenoLopez-eProsima if it's neither, a warning log stating that the given value is unknown and the default will take over would be nice.

hidmic (commented):
@JLBuenoLopez-eProsima to configure modes like this would be an interesting feature, but to put it in the `rmw` API spec we have to unambiguously define what each mode entails, decide on what the best granularity for configuration would be, and how should implementations that lack support for some (or all) of these modes behave.

Either a design document or a draft update to [`rmw`](https://github.com/ros2/rmw) documentation could kickstart the discussion.

hidmic:
CC @wjwwood @ivanpauno for thoughts on making (or not making) this an `rmw` feature.

ivanpauno:
> ROS_PUBLICATION_MODE

Considering this option is only supported in `rmw_fastrtps`, I would name the variable `RMW_FASTRTPS_PUBLICATION_MODE`.
Otherwise, this sounds good to me.

wjwwood:
> Considering this option is only supported in rmw_fastrtps, I would name the variable RMW_FASTRTPS_PUBLICATION_MODE.
Otherwise, this sounds good to me.

Agreed. I do not think we want `rmw` or ROS specific features flagged on and off with env variables... This variable will affect all publishers, and will change the behavior significantly of all publish calls. It should be used very carefully imo. I'm only in favor of it because it affects on `rmw_fastrtps_cpp` and because it is already possible by setting two other environment variables.

EduPonz:
Our line of thought was following the one discussed in ros2/ros2#255, #343, and ros2/rmw_cyclonedds#89. From those discussions, it can be concluded that asynchronous publishing is indeed a requirement for the middlewares, and that having a synchronous mode is optional. In fact, #343 was not merged because of this, even though synchronous publication has better latency performance that can compete with that of other vendors.

IMHO there are cases when applications benefit more from one than the other. We have in fact dealt with issues regarding this over and over, where we recommend changing the publication mode. However, this is currently cumbersome since it involves XML configuration files + environment variables. In that sense, I think it's better to make it as easy as possible for the user to make the change (because why not). In fact, I think this could be part of the upper-level API, so it can be configured programatically from the  beginning. In that sense, maybe the publish mode configuration could be publisher based, publication based, context based, etc.

Sure right now it's only Fast DDS who can make the switch, but that is just because no other vendors (apart from connext) meet the requirement regarding this. However, that might change in the future, so having an mechanism already in place would benefit everyone. Furthermore, user have easier access to upper-layer documentations, so it'll simply get more visibility upstream than just being a Fast DDS specific thing (when it should not). This will also entail letting the user know when the variable does not apply when using other RMWs, which is probably a good thing, since it let's the user know about the middleware behaviour, instead of being different from one RMW to the other without the user knowing or even wanting.

Also, there are other cases were environment variables change ROS behaviour. An example of it would be `ROS_LOCALHOST_ONLY`, so it won't be a first. I agree with @wjwwood that it would require careful usage, but that does not mean we should limit users' options. Instead, a good documentation will help them make the best decision for their use-case, instead of us making a decision for them that maybe they don't want, and that is different from RMW to RMW.

rmw_fastrtps_shared_cpp/src/participant.cpp
JLBuenoLopez-eProsima: Done

rmw_fastrtps_shared_cpp/src/participant.cpp
JLBuenoLopez-eProsima: Done

README.md
JLBuenoLopez-eProsima: For the moment, I have changed the name to the one proposed by @ivanpauno. If later other rmw implementations start to use a similar variable to configure the publication mode we can consider updating this code.

wjwwood:
> Sure right now it's only Fast DDS who can make the switch, but that is just because no other vendors (apart from connext) meet the requirement regarding this. However, that might change in the future, so having an mechanism already in place would benefit everyone. 

Yes, but in my opinion that should be a QoS setting that you set when you create a publisher, not an environment variable.

> An example of it would be ROS_LOCALHOST_ONLY, so it won't be a first.

For me the important difference (which makes `ROS_LOCALHOST_ONLY` ok and something like `RMW_ASYNCHRONOUS_PUBLISHING` not ok imo), is that the `ROS_LOCALHOST_ONLY` doesn't (significantly or concretely) affect the behavior of the API calls, whereas this change can significantly affect the blocking behavior of `DataWriter::write()`.

EduPonz:
> Yes, but in my opinion that should be a QoS setting that you set when you create a publisher, not an environment variable.

I completely agree with this vision. It'd be great to be able to create a publisher with either publication mode. The thing is that currently this is not possible, that's why we wanted some way to make it easier for the user to do so. I think this is related to some discussions held here and there, about how to design an API that is agnostic of the specific middleware, while at the same time lets the user really tune the middleware (https://discourse.ros.org/t/restricting-communication-between-robots/2931/15?u=eduponz).

I'm very much ok with setting the name to something like `RMW_FASTRTPS_PUBLICATION_MODE` if such APIs will exist in the future in a more general and ordered manner (maybe in Galactic). In fact, I think it'd be really great for users and vendors to have some set of requirements in terms of features, performance, configuration paramentes, etc. I think the way the Quality Levels were worked out is a good approach to these kinds of problems, specially because the whenever anyone would make a RMW switch, everything that that entails would be well documented and quite visible.

Such set of requirements may be worked out in the middleware working group, and could also help the decision of the default middleware. They'd most importantly be very beneficial for the community as a whole, since they would push development towards the real end-users needs. The requirement would also help designing what QoS are configured programatically; if a feature is required, it should be configurable via QoS. IMO this would be a very interesting discussion to have if you guys are up to moving it somewhere else. Doing so, we would avoid the somewhat unorthodox way of configuring with XML we have right now.

> For me the important difference (which makes ROS_LOCALHOST_ONLY ok and something like RMW_ASYNCHRONOUS_PUBLISHING not ok imo), is that the ROS_LOCALHOST_ONLY doesn't (significantly or concretely) affect the behavior of the API calls, whereas this change can significantly affect the blocking behavior of DataWriter::write().

You are completely right, the change will significantly affect the behaviour of the call. However, I do not think that's a bad thing in itself, as long as the implications are well documented. @JLBuenoLopez-eProsima , what do you think about explaining a bit more the difference between the modes in the README file?

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -195,6 +198,26 @@ rmw_fastrtps_shared_cpp::create_participant(
...
+      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("Error getting env var: %s\n", error_str);
+      return nullptr;
+    }
+    if (env_value != nullptr) {
+      // Synchronous publishing
+      if (strcmp(env_value, "SYNCHRONOUS") == 0) {
+        publishing_mode = publishing_mode_t::SYNCHRONOUS;
+      } else if (strcmp(env_value, "AUTO") == 0) {
+        publishing_mode = publishing_mode_t::AUTO;
+      } else if (strcmp(env_value, "ASYNCHRONOUS") != 0 && strcmp(env_value, "") != 0) {
```
fujitatomoya, line 213:
```suggestion
      } else if (strcmp(env_value, "ASYNCHRONOUS") != 0) {
```

I think empty case `""` is not admissible based on the design.

JLBuenoLopez-eProsima: The environment variable is read using `rcutils_get_env()`:

```
const char *
rcutils_get_env(const char * env_name, const char ** env_value)
{
  RCUTILS_CAN_RETURN_WITH_ERROR_OF("some string error");

  if (NULL == env_name) {
    return "argument env_name is null";
  }
  if (NULL == env_value) {
    return "argument env_value is null";
  }

  // TODO(Suyash458): getenv is deprecated on Windows; consider using getenv_s instead
  *env_value = getenv(env_name);

  if (NULL == *env_value) {
    *env_value = "";
  }
  return NULL;
}
```
It uses function `getenv` that returns `NULL` if the variable does not exist. Below, if `NULL` has been returned, `rcutils_get_env()` returns empty string `""`. Therefore, to avoid the warning log appearance when the environment variable is not defined, this condition should also be checked.

JLBuenoLopez-eProsima:
@EduPonz,

> @JLBuenoLopez-eProsima , what do you think about explaining a bit more the difference between the modes in the README file?

Please read last commit.

rmw_fastrtps_shared_cpp/src/participant.cpp
fujitatomoya: that makes sense, appreciate for the explanation.

README.md
```diff
@@ -44,6 +44,19 @@ You have two ways of telling you ROS 2 application which XML to use:
 1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
 2. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to your XML file.
 
+### Change publication mode
+
+Another way to change easily the publication mode is to use the environment variable `RMW_FASTRTPS_PUBLICATION_MODE`.
+The admissible values are:
+* `ASYNCHRONOUS`: asynchronous publication mode.
+This implies that when the publisher calls the write operation, the data is copied in a queue and control is returned to the user thread before the data is actually sent.
+The asynchronous thread is in charge of consuming the queue and send the data to any matched reader.
```
EduPonz, line 53:
```suggestion
Setting this mode implies that when the publisher invoques the write operation, the data is copied into a queue, a notification about the addition to the queue is performed, and control of the thread is returned to the user before the data is actually sent.
A background thread (asynchronous thread) is in turn in charge of consuming the queue and sending the data to every matched reader.
```

README.md
```diff
@@ -44,6 +44,19 @@ You have two ways of telling you ROS 2 application which XML to use:
...
+* `SYNCHRONOUS`: synchronous publication mode.
+This implies that the data is sent directly in the context of the user thread.
```
EduPonz, line 55:
```suggestion
Setting this mode implies that the data is sent directly within the context of the user thread.
This entails that any blocking call occurring during the write operation would block the user thread, thus preventing the application with continuing its operation.
It is important to note that this mode typically yields higher throughput rates at lower latencies, since the notification and context switching between threads is not present.
```

README.md
```diff
@@ -44,6 +44,19 @@ You have two ways of telling you ROS 2 application which XML to use:
...
+* `AUTO`: let Fast DDS select the publication mode. This implies using the publication mode set in the XML file or, failing that, the default value set in Fast DDS (which currently is set to `SYNCHRONOUS`).
+
+By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS` and it is the publication mode in use if this environment variable is not set.
```
EduPonz, line 58:
```suggestion
By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`, which is the publication mode in use if this environment variable is not set.
```

EduPonz (changes requested):
Some suggestions in the docs

hidmic (commented):
@JLBuenoLopez-eProsima how do `RMW_FASTRTPS_USE_QOS_FROM_XML` and  `RMW_FASTRTPS_PUBLICATION_MODE` interact? 

I'll kickstart CI in the meantime:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12928/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7880/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10647/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12878/)

JLBuenoLopez-eProsima:
@hidmic,

I have been studying the test that is failing in MacOS (`TestNodeConstructionDestruction__rmw_[implementation].create_with_internal_errors`). I have launched these tests locally in MacOS using directly the Rolling Ridley versions and I have seen that the test is flaky. Also, not only is failing in `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` implementations, but also in `rmw_cyclonedds_cpp`. The test fails with an uncaught exception of type `std::__1::system_error: mutex lock failed: Invalid  argument`.

Therefore, the MacOS failures seem unrelated to this pull request and a more in-depth analysis of this test should be done.

hidmic:
> RMW_FASTRTPS_USE_QOS_FROM_XML has priority over RMW_FASTRTPS_PUBLICATION_MODE.

Great. Thanks for the clarification. Would you mind adding that sentence to the documentation?

hidmic:
> Therefore, the MacOS failures seem unrelated to this pull request and a more in-depth analysis of this test should be done.

Indeed. I can see flakes in MacOS nightlies.

hidmic (approved):
LGTM once https://github.com/ros2/rmw_fastrtps/pull/470#issuecomment-721945928 is addressed.

JLBuenoLopez-eProsima:
> LGTM once [#470 (comment)](https://github.com/ros2/rmw_fastrtps/pull/470#issuecomment-721945928) is addressed.

Last commit explains the precedence between both environment variables.

It also clarifies the relationship between the XML file and `RMW_FASTRTPS_USE_QOS_FROM_XML` taking into account [#466 (comment)](https://github.com/ros2/rmw_fastrtps/pull/466/files#r517322626).

EduPonz:
Hi @hidmic ,

Do you think this is good to go then? IMHO, it provides the users of `rmw_fastrtps` with an easy a quick solution for a recurring problem that otherwise is more tedious to solve, while at the same time documenting the capabilities regarding the publication mode better. I think it can only do good to have it.

hidmic (approved):
Thanks @JLBuenoLopez-eProsima. @EduPonz yes, I'm onboard with this change. @wjwwood has a good point though. Hopefully, at some point, the `rmw` API is flexible enough to support such configurations and environment variables are no longer needed. Proposals are most welcome.

hidmic:
Another approval would be nice @wjwwood @ivanpauno @mjeronimo.

README.md
```diff
@@ -44,6 +44,23 @@ You have two ways of telling you ROS 2 application which XML to use:
...
+A background thread (asynchronous thread) is in turn in charge of consuming the queue and sending the data to every matched reader.
+* `SYNCHRONOUS`: synchronous publication mode.
+Setting this mode implies that the data is sent directly within the context of the user thread.
+This entails that any blocking call occurring during the write operation would block the user thread, thus preventing the application with continuing its operation.
+It is important to note that this mode typically yields higher throughput rates at lower latencies, since the notification and context switching between threads is not present.
+* `AUTO`: let Fast DDS select the publication mode. This implies using the publication mode set in the XML file or, failing that, the default value set in Fast DDS (which currently is set to `SYNCHRONOUS`).
+
+By default, `rmw_fastrtps` assumes this variable to be `ASYNCHRONOUS`, which is the publication mode in use if this environment variable is not set.
```
mjeronimo, line 62: Do you really mean "assumes" here? That is, is there an implicit assumption that the value is ASYNCHRONOUS? Or, does this mean that rmw_fastrtps sets the value, by default, to ASYNCHRONOUS? I suppose the code will clarify, but the description should be clear.

mjeronimo (approved):

README.md
wjwwood: I think it must be "if `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then `rmw_fastrtps_cpp` behaves as if it were set to `ASYNCHRONOUS`.

wjwwood (approved):

README.md
JLBuenoLopez-eProsima: Done in 2bb6f5fbb1e7f971ab941c944998d0ca20a23c88

EduPonz:
Hi everyone,

The suggestion made by @wjwwood was accounted for in 2bb6f5f. Is there anything else to be done before merging?

hidmic:
Nothing else to be added.

EduPonz:
Thanks so much to everybody to push this further!

PR #477, opened 2020-11-03
Fix rmw_time to Duration_t conversion

mjeronimo:
The rmw_time_t uses 64-bit fields for seconds and nanoseconds. These
values are truncated to the eprosima::fastrtps::Duration_t's fields
which are int32 and uint32. This truncation causes problems when
the seconds value is greater than INT_MAX. For example, when attempting
to create a subscriber with a QoS policy that has a Duration field
(such as lifespan, deadline, or liveliness_leas_duration) where
seconds = INT_MAX and nanoseconds = UINT_MAX.

As a work-around (until Fast-DDS supports 64-bit seconds and nanoseconds)
this change ensures that the seconds value can be represented with a
32-bit integer.

Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -31,7 +31,7 @@ rmw_time_to_fastrtps(const rmw_time_t & time)
 
   // TODO(mjeronimo): Since we're going to truncate the 64-bit values
   // down to 32-bit, make sure at least that the seconds value does
-  // not overflow. If the coming value is greater than INT_MAX, move
+  // not overflow. If the incoming value is greater than INT_MAX, move
   // the excess seconds over to the nanoseconds field.
   if (t.sec > INT_MAX) {
     uint64_t diff = t.sec - INT_MAX;
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <limits>
+#include <limits.h>
 
 #include "rmw_fastrtps_shared_cpp/qos.hpp"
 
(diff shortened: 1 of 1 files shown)


mjeronimo:
To reproduce, insert the following line:

`custom_qos_profile.liveliness_lease_duration = Duration(seconds=2147483647, nanoseconds=4294967295);`

into the talker_qos.py demo at line 69 and run it:

https://github.com/ros2/demos/blob/master/demo_nodes_py/demo_nodes_py/topics/talker_qos.py#L69

It fails because the lease duration value has been truncated and is now less than the announcement period:

```
[INFO] [1604369510.056638524] [talker_qos]: Best effort talker
2020-11-02 18:11:50.057 [RTPS_QOS_CHECK Error] WRITERQOS: LeaseDuration <= announcement period. -> Function checkQos

```

MiguelCompany:
> until Fast-DDS supports 64-bit seconds and nanoseconds

This will probably never happen. The DDS standard specifies the `Duration_t` type with the following IDL:

```
struct Duration_t {
 long sec;
 unsigned long nanosec;
 };
```

And the [IDL to C++11 mapping](https://www.omg.org/spec/CPP11/1.4/PDF) states that the C++ types for `long` and `unsigned long` are mapped to `int32_t` and `uint32_t` respectively.

The DDS standard also states that nanoseconds should be normalized (i.e. `the
nanosec field must verify 0 <= nanosec < 1000000000`), and defines the following constants to indicate an infinite duration:

```
const long DURATION_INFINITE_SEC = 0x7fffffff;
const unsigned long DURATION_INFINITE_NSEC = 0x7fffffff;
```

So I wonder if these restrictions should be taken into account on the upper layers, and be common to all RMW implementations.

clalancette:
> This will probably never happen. The DDS standard specifies the `Duration_t` type with the following IDL:

But it has to happen, one way or another.  In 2038, the Unix epoch is going to roll over for 32-bit second representations.  At which point, everything having to do with DDS is going to break badly (actually it will probably break long before that, but that's another discussion).

So I'd much rather make the upper layers deal with 64-bit seconds as much as possible, and then have OMG consider how to handle the transition from 32-bit representation of seconds to 64-bit in the DDS standard.

MiguelCompany:
@clalancette But we are talking about durations here, not timestamps. I agree with you for the definition of the `Time_t` structure. On `Duration_t` it may not be necessary, but I would do it nevertheless for simplicity (allows dowing `using Duration_t = Time_t`)

One more thing. It has already been reported [here](https://issues.omg.org/issues/spec/DDS/1.4#issue-43220) that the standard does not define the calendar time of `Time_t` at `{ sec = 0, nanosec = 0 }`, even though I would assume all implementations are using Unix epoch for it.

clalancette:
> @clalancette But we are talking about durations here, not timestamps. I agree with you for the definition of the `Time_t` structure.

Good point.

> On `Duration_t` it may not be necessary, but I would do it nevertheless for simplicity (allows dowing `using Duration_t = Time_t`)

But yeah, I agree with this; we may as well make both Duration and Time 64-bit quantities.

> One more thing. It has already been reported [here](https://issues.omg.org/issues/spec/DDS/1.4#issue-43220) that the standard does not define the calendar time of `Time_t` at `{ sec = 0, nanosec = 0 }`, even though I would assume all implementations are using Unix epoch for it.

Right.  If most of the implementations are using Unix epoch, then we have a problem.  But that can only be solved by the OMG.

Coming back to this patch, I agree that this is a generic problem with any DDS implementation.  @mjeronimo I wonder if it makes sense to make a function somewhere in `rmw_dds_common` to deal with this, and then call it from the various RMW implementations.

mjeronimo:
Thanks, @MiguelCompany and @clalancette. I'll close this one and open a PR over in rmw_dds_common.

Issue #478, opened 2020-11-04
Segmentation faults and uninitialised wait sets


gbiggs:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Binaries, from source
- Version or commit hash:
  - 
  - Binaries: 1.2.1-1focal.20201007.210239
  - Source: f54c74b80981fd03857c896b980924f132e8483f
- DDS implementation:
  - rmw_fastrtps_cpp
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

1. Check out the rmf_core repository into a workspace: https://github.com/osrf/rmf_core
1. Switch to the [`fastdds_segfaults` branch](https://github.com/osrf/rmf_core/tree/fastdds_segfaults)
1. Compile the `rmf_fleet_adapter` package:
   ```
   rosdep install --from-paths src --ignore-src -yr
   colcon build --packages-up-to rmf_fleet_adapter
   ```
1. Execute the small program that reliably triggers the segmentation fault
   ```
   source install/setup.bash
   ./build/rmf_fleet_adapter/segfaulter
   ```

#### Expected behavior

The sample program completes successfully without any errors.

#### Actual behavior

The sample program, in most iterations after the first couple, either fails to delete a wait set or causes segmentation faults in `rmw_fastrtps_cpp` code.

Example output:

```
$ ./build/rmf_fleet_adapter/segfaulter   
0
[INFO] [1604462249.727237249] [test_node_0]: Added a robot named [test_robot] with participant ID [0]
1
[INFO] [1604462251.764308128] [test_node_1]: Added a robot named [test_robot] with participant ID [0]
2
[INFO] [1604462253.797161121] [test_node_2]: Added a robot named [test_robot] with participant ID [0]
"/home/geoff/src/workspaces/ros2_foxy_debug/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/listener_thread.cpp":__function__:150"failed to destroy wait set": ros discovery info listener thread will shutdown ...
"/home/geoff/src/workspaces/ros2_foxy_debug/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/listener_thread.cpp":__function__:150"failed to destroy wait set": ros discovery info listener thread will shutdown ...
3
[INFO] [1604462255.833644587] [test_node_3]: Added a robot named [test_robot] with participant ID [0]
[...]

PR #486, opened 2020-11-17
Check for correct context shutdown

imontesino:
Added checks for the creation of nodes in context which has been shutdown, and for shutdown of a context with active nodes. 

Thess throw a `RMW_RET_ERROR` instead of causing segfaults as seen in ros2/rmw_fastrtps#478

Signed-off-by: Ignacio Montesino <ignaciomontesino@eprosima.com>

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -134,6 +134,10 @@ rmw_shutdown(rmw_context_t * context)
     context->implementation_identifier,
     eprosima_fastrtps_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  if (context->impl->count > 0) {
+    RMW_SET_ERROR_MSG("Shutting down context with active nodes");
+    return RMW_RET_ERROR;
+  }
   context->impl->is_shutdown = true;
   return RMW_RET_OK;
 }
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -134,10 +134,6 @@ rmw_shutdown(rmw_context_t * context)
     context->implementation_identifier,
     eprosima_fastrtps_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
-  if (context->impl->count > 0) {
-    RMW_SET_ERROR_MSG("Shutting down context with active nodes");
-    return RMW_RET_ERROR;
-  }
   context->impl->is_shutdown = true;
   return RMW_RET_OK;
 }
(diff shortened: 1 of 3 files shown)


EduPonz:
@ivanpauno Can you take a look at this in relation with https://github.com/ros2/rmw_fastrtps/issues/478#issuecomment-724788134?

ivanpauno (changes requested):
The issue isn't `rmw_shutdown` but ``rmw_context_fini`.
Shutdown is a flag that doesn't clean any resource, and it's safe to call it will a node created from that context is alive.

The issue is calling `rmw_context_fini()` when not all of its node have been destroyed.

EduPonz:
What we understood from https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp/src/rmw_init.cpp#L154-L157 is that `rmw_context_fini()` checks whether the context has been shutdown, and so I thought that the nodes' destruction was a pre-condition for `rmw_shutdown()`. If that is not the case, then we'll update this PR accordingly

ivanpauno:
> What we understood from https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp/src/rmw_init.cpp#L154-L157 is that rmw_context_fini() checks whether the context has been shutdown, and so I thought that the nodes' destruction was a pre-condition for rmw_shutdown(). If that is not the case, then we'll update this PR accordingly

Context shutdown is a precondition for context fini, but not for node fini.
All node finis are a precondition of the parent context fini.

ivanpauno:
To be more clear, context shutdown at rmw level is actually "doing nothing", so it cannot cause an use after free as reported in #478.

EduPonz:
@ivanpauno , could you take a look at this?

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -155,6 +155,10 @@ rmw_context_fini(rmw_context_t * context)
     RCUTILS_SET_ERROR_MSG("context has not been shutdown");
     return RMW_RET_INVALID_ARGUMENT;
   }
+  if (context->impl->count > 0) {
+    RMW_SET_ERROR_MSG("Deleting a context with active nodes");
```
ivanpauno, line 159: nitpick^N:
```suggestion
    RMW_SET_ERROR_MSG("Finalizing a context with active nodes");
```

rmw_fastrtps_dynamic_cpp/src/rmw_init.cpp
```diff
@@ -155,6 +155,10 @@ rmw_context_fini(rmw_context_t * context)
     RCUTILS_SET_ERROR_MSG("context has not been shutdown");
     return RMW_RET_INVALID_ARGUMENT;
   }
+  if (context->impl->count > 0) {
+    RMW_SET_ERROR_MSG("Deleting a context with active nodes");
```
ivanpauno, line 159:
```suggestion
    RMW_SET_ERROR_MSG("Finalizing a context with active nodes");
```

ivanpauno (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13177/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8108/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10893/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13191/)

ivanpauno:
@imontesino can you check the macos failures?
The windows warnings don't seem to be related.

jacobperron:
The macOS errors have been occurring a little while now. See https://github.com/ros2/rmw_fastrtps/issues/491.

jacobperron (approved):

PR #488, opened 2020-11-20
Make use of error return value in decrement_context_impl_ref_count

brawner:
This was found during scan-build, but the return value `ret` is not actually returned here. I suspect it was meant to be returned at the end of the if statement? 

Once this is resolved, scan-build passes cleanly for `rmw_fastrtps_cpp` and `rmw_fastrtps_shared_cpp`

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/init_rmw_context_impl.cpp
+++ b/rmw_fastrtps_shared_cpp/src/init_rmw_context_impl.cpp
@@ -39,70 +39,71 @@ rmw_fastrtps_shared_cpp::decrement_context_impl_ref_count(rmw_context_t * contex
   assert(context->impl);
   assert(0u < context->impl->count);
 
-  if (0u == --context->impl->count) {
-    rmw_ret_t err = RMW_RET_OK;
-    rmw_ret_t ret = RMW_RET_OK;
-    rmw_error_string_t error_string;
+  if (--context->impl->count > 0) {
+    return RMW_RET_OK;
+  }
 
-    ret = rmw_fastrtps_shared_cpp::join_listener_thread(context);
-    if (RMW_RET_OK != ret) {
-      return ret;
-    }
+  rmw_ret_t err = RMW_RET_OK;
+  rmw_ret_t ret = RMW_RET_OK;
+  rmw_error_string_t error_string;
 
-    auto common_context = static_cast<rmw_dds_common::Context *>(context->impl->common);
-    auto participant_info = static_cast<CustomParticipantInfo *>(context->impl->participant_info);
-
-    if (!common_context->graph_cache.remove_participant(common_context->gid)) {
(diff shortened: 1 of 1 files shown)


brawner:
Testing `--packages-above rmw_fastrtps_shared_cpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13131/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8062/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13125/)

rmw_fastrtps_shared_cpp/src/init_rmw_context_impl.cpp
```diff
@@ -102,6 +102,7 @@ rmw_fastrtps_shared_cpp::decrement_context_impl_ref_count(rmw_context_t * contex
     delete common_context;
     context->impl->common = nullptr;
     context->impl->participant_info = nullptr;
+    return ret;
   }
   return RMW_RET_OK;
```
clalancette, line 107: One way I think we could make this a bit clearer is to do a slight refactor of this method to move the `context->impl->count != 0` check to the top of the method.  That is, the start of the method would look something like:

```
std::lock_guard<std::mutex> guard(context->impl->mutex);
context->impl->count--;
if (context->impl->count > 0) {
  return RMW_RET_OK;
}

// implicitly, the count is now 0, and we can do cleanup things.
```

(we would still need your fix to `return ret` here, but now I think it would be a bit more clear that you have to do so, and the lack of it would have been picked up by normal compiler flags)

brawner: I love using early returns whenever possible. Fixed.

clalancette (approved):
I like it.  Looks good with green CI.

brawner:
The above CI was run with `--packages-above` which showed some test failures on OSX and Windows, but they were either for a different rmw, or were pre-existing on windows nightlies (https://ci.ros2.org/view/nightly/job/nightly_win_rel/1757/#showFailuresLink).

This only runs `--packages-select rmw_fastrtps_shared_cpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13144/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8075/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13140/)

brawner:
Rpr job failed due to docker pull rate limiting: 

> Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading

PR #489, opened 2020-11-23
Revert "Discriminate when the Client has gone from when the Client has not completely matched (#479)"

sloretz:
This reverts commit c78c31a825932d9cc6682120d06e6fcea733a3bb from #479 @jacobperron FYI

It appears to be the cause of [Foxy CI with `rmw_fastrtps_dynamic_cpp` hanging](http://build.ros2.org/view/Fci/job/Fci__nightly-fastrtps-dynamic_ubuntu_focal_amd64/205/). It causes the rclcpp composable node container to crash, and that causes a test in `test_launch_ros` to hang forever.

A more minimal example is:

```
$ export RMW_IMPLEMENTATION=rmw_fastrtps_dynamic_cpp
$ ./install/rclcpp_components/lib/rclcpp_components/component_container
```

```
$ export RMW_IMPLEMENTATION=rmw_fastrtps_dynamic_cpp
$ ros2 component list
```

When `ros2 component list` is called, the container segfaults.

<details><summary>Crash happens entirely in FastRTPS</summary>

```
#0  0x00007ffff6e9e4c4 in std::_Hashtable<eprosima::fastrtps::rtps::GUID_t, std::pair<eprosima::fastrtps::rtps::GUID_t const, eprosima::fastrtps::rtps::GUID_t>, std::allocator<std::pair<eprosima::fastrtps::rtps::GUID_t const, eprosima::fastrtps::rtps::GUID_t> >, std::__detail::_Select1st, std::equal_to<eprosima::fastrtps::rtps::GUID_t>, rmw_fastrtps_shared_cpp::hash_fastrtps_guid, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true> >::_M_find_before_node(unsigned long, eprosima::fastrtps::rtps::GUID_t const&, unsigned long) const () from /home/sloretz/install/rmw_fastrtps_dynamic_cpp/lib/librmw_fastrtps_dynamic_cpp.so
[...]

PR #490, opened 2020-12-01
[foxy] Redo "Discriminate when the Client has gone from when the Client has not completely matched"

jacobperron:
Revert #489 (9094971) and add a patch to fix the reported bug (34a23d5).

I think the issue was that the shared function for sending the service response [assumes we are using the patched version of the service listener](https://github.com/ros2/rmw_fastrtps/blob/34a23d51d4611cb5398c11e827bcc67fc02ccc13/rmw_fastrtps_shared_cpp/src/rmw_response.cpp#L121), and I had not updated the `rmw_fastrtps_dynamic_cpp` implementation to use this type. This PR fixes that issue.

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_service.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_service.cpp
@@ -241,7 +241,7 @@ rmw_create_service(
         delete info->pub_listener_;
       }
     });
-  info->pub_listener_ = new (std::nothrow) ServicePubListener();
+  info->pub_listener_ = new (std::nothrow) PatchedServicePubListener();
   if (!info->pub_listener_) {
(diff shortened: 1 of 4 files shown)


jacobperron:
CI with Fast-RTPS dynamic:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13194/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8125/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10910/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13208/)

wjwwood (approved):
lgtm

PR #492, opened 2020-12-07
Make sure to lock the mutex protecting client_endpoints_.

clalancette:
There were actually 2 problems here:
1.  We were missing the lock in check_for_subscription()
completely (this is the original issue that caused me to
examine this code).
2.  We had an accessor for clients_endpoints_ so that external
users could get a reference.  But with that being the case,
any access to the returned reference would be without a lock.
Since we were only accessing client_endpoints_ in two places,
make some methods in ServicePubListener to do the requested
operations while holding the lock, getting rid of the problem.

I was initially investigating an issue pointed out by https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/627/gcc/new/ , but upon review of the code I found both problems above.  I did some local tests, and this seems to work.  I'm going to run a full CI on all platforms to make sure there are no lingering problems with this.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -125,10 +125,13 @@ public:
   check_for_subscription(
     const eprosima::fastrtps::rtps::GUID_t & guid)
   {
-    // Check if the guid is still in the map
-    if (clients_endpoints_.find(guid) == clients_endpoints_.end()) {
-      // Client is gone
-      return client_present_t::GONE;
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      // Check if the guid is still in the map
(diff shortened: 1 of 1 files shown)


clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8161/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10946/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13263/)

fujitatomoya (approved):
lgtm

EduPonz:
Once this is merged, it'd be really good to have it backported to Foxy. It'd actually be great if the backport could be included in Friday's Foxy release. Can we do something to help you guys with that?

hidmic (approved):
Looks reasonable to me, and Windows CI test failures seem unrelated flakes to me.

clalancette:
Thanks for the reviews, I'm going to go ahead and merge this.

@EduPonz I'll open a backport PR to Foxy as well, and we'll see how @jacobperron wants to handle it.

PR #493, opened 2020-12-09
[Foxy backport]: Make sure to lock the mutex protecting client_endpoints_. (#492)

clalancette:
There were actually 2 problems here:
1.  We were missing the lock in check_for_subscription()
completely (this is the original issue that caused me to
examine this code).
2.  We had an accessor for clients_endpoints_ so that external
users could get a reference.  But with that being the case,
any access to the returned reference would be without a lock.
Since we were only accessing client_endpoints_ in two places,
make some methods in ServicePubListener to do the requested
operations while holding the lock, getting rid of the problem.

Note that the patch here is slightly different than the one on master, since I had to take into account the ABI-compatible differences.  Like I did with master, I'm going to run a full CI on this.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp
@@ -175,10 +175,13 @@ public:
   check_for_subscription(
     const eprosima::fastrtps::rtps::GUID_t & guid)
   {
-    // Check if the guid is still in the map
-    if (clients_endpoints_.find(guid) == clients_endpoints_.end()) {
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13245/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8176/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10960/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13278/)

jacobperron (approved):

jacobperron:
I'm pretty sure the macOS and Windows test failures are pre-existing. I'm going ahead and merging this before making new release.

PR #495, opened 2020-12-15
Handle typesupport errors on fetch.

hidmic:
Follow-up after https://github.com/ros2/rosidl_typesupport/pull/99. Missing error handling code caused test failures in downstream packages (e.g. https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_release/1415/). This patch addresses them.

Full CI (just in case):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13273/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8205/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10993/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13304/)

### DIFF
--- a/rmw_fastrtps_dynamic_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/rmw_client.cpp
@@ -97,9 +97,11 @@ rmw_create_client(
   const rosidl_service_type_support_t * type_support = get_service_typesupport_handle(
     type_supports, rosidl_typesupport_introspection_c__identifier);
   if (!type_support) {
+    rcutils_reset_error();
     type_support = get_service_typesupport_handle(
       type_supports, rosidl_typesupport_introspection_cpp::typesupport_identifier);
     if (!type_support) {
+      rcutils_reset_error();
       RMW_SET_ERROR_MSG("type support not from this implementation");
       return nullptr;
     }
--- a/rmw_fastrtps_dynamic_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/rmw_client.cpp
@@ -97,12 +97,19 @@ rmw_create_client(
   const rosidl_service_type_support_t * type_support = get_service_typesupport_handle(
     type_supports, rosidl_typesupport_introspection_c__identifier);
   if (!type_support) {
+    rcutils_error_string_t prev_error_string = rcutils_get_error_string();
     rcutils_reset_error();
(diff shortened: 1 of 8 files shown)


clalancette (approved):
My understanding here is that we first try to find the C type support, and then fall back to the CPP type support.  In the process of finding the C type support, an error gets set in the rcutils layer.  Prior to https://github.com/ros2/rosidl_typesupport/pull/99 , we were just ignoring that error, but that PR added additional error checking.  This fix changes it so that we are sure to reset the error after not finding the C type support, so if we find the CPP type support, no error at all is set.

So with all of that said, this looks like the correct fix to me, assuming CI comes back green.

hidmic:
> My understanding here is that we first try to find the C type support, and then fall back to the CPP type support. In the process of finding the C type support, an error gets set in the rcutils layer. Prior to ros2/rosidl_typesupport#99 , we were just ignoring that error, but that PR added additional error checking. This fix changes it so that we are sure to reset the error after not finding the C type support, so if we find the CPP type support, no error at all is set.

Your assessment is correct, but failures actually happen up the stack in tests that assert an error is not set (and potentially in code that throws if an error is set).

rmw_fastrtps_dynamic_cpp/src/rmw_client.cpp
```diff
@@ -97,9 +97,11 @@ rmw_create_client(
   const rosidl_service_type_support_t * type_support = get_service_typesupport_handle(
     type_supports, rosidl_typesupport_introspection_c__identifier);
   if (!type_support) {
+    rcutils_reset_error();
     type_support = get_service_typesupport_handle(
       type_supports, rosidl_typesupport_introspection_cpp::typesupport_identifier);
     if (!type_support) {
+      rcutils_reset_error();
```
MiguelCompany, line 104: Why are we not getting old messages here, as it is done for `rmw_fastrtps_cpp` ?

hidmic: Oops, I missed that one. Thanks @MiguelCompany !

hidmic: See https://github.com/ros2/rmw_fastrtps/pull/495/commits/b45a8ec945452d1a78a41dfac9fdc40b239b191b.

hidmic:
MacOs and Windows test failures were there before last night crash.

PR #496, opened 2020-12-15
Set rmw_dds_common::GraphCache callback after init succeeds.

### LINKED ISSUE #491: TestNodeConstructionDestruction__rmw_fastrtps_cpp.create_with_internal_errors fails due to invalid mutex on macOS
## Bug report

**Required Info:**

- Operating System:
  OS X

#### Steps to reproduce issue
See https://ci.ros2.org/view/nightly/job/nightly_osx_release/1870/#showFailuresLink

#### Additional information

----

hidmic:
Otherwise, during stack unwinding, the graph cache may end up with a callback that captures and uses an already destroyed guard condition.

This patch (hopefully) fixes #491. 

Ci up `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13277/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11008/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13311/)

### DIFF
--- a/rmw_fastrtps_cpp/src/init_rmw_context_impl.cpp
+++ b/rmw_fastrtps_cpp/src/init_rmw_context_impl.cpp
@@ -151,13 +151,6 @@ init_context_impl(rmw_context_t * context)
     return RMW_RET_BAD_ALLOC;
   }
 
-  common_context->graph_cache.set_on_change_callback(
-    [guard_condition = graph_guard_condition.get()]() {
-      rmw_fastrtps_shared_cpp::__rmw_trigger_guard_condition(
-        eprosima_fastrtps_identifier,
-        guard_condition);
-    });
-
   common_context->gid = rmw_fastrtps_shared_cpp::create_rmw_gid(
     eprosima_fastrtps_identifier, participant_info->participant->getGuid());
   common_context->pub = publisher.get();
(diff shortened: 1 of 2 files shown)


clalancette (commented):
Just so I understand the problem:

What was happening previously was that if `run_listener_thread` failed, then we would return out of this function.  Since `graph_guard_condition` is a `unique_ptr`, it would be freed.

However, the fact that we go to this point also meant that a graph change would be triggered.  During the callback here, we would try to access the `graph_guard_condition`, which has already been freed.

The fix is to move the `set_on_change_callback` until after we can't fail anymore, so we can never get into this situation.

Does that describe the problem correctly?

hidmic:
> What was happening previously was that if run_listener_thread failed, then we would return out of this function. Since graph_guard_condition is a unique_ptr, it would be freed.
> However, the fact that we go to this point also meant that a graph change would be triggered. During the callback here, we would try to access the graph_guard_condition, which has already been freed.
> The fix is to move the set_on_change_callback until after we can't fail anymore, so we can never get into this situation.

Exactly.

clalancette (approved):

hidmic:
Thanks for the review!

Issue #500, opened 2021-01-12
[ros2] QoS parameters can't be configured by parsing DEFAULT_FASTRTPS_PROFILES.xml


yangli-flexiv:
[QoS_github.zip](https://github.com/ros2/rmw_fastrtps/files/5805402/QoS_github.zip)
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - install from binaries
- Version or commit hash:
  - eloquent of rmw_fastrtps
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

I have attached all the relevant files (well organized as a workspace) as a zip.

First, I created an `xml` file, `DEFAULT_FASTRTPS_PROFILES.xml`, to configure the QoS parameters for a publisher and subscriber node
```
<?xml version="1.0" encoding="UTF-8"?>
<dds xmlns="http://www.eprosima.com/XMLSchemas/fastRTPS_Profiles">
  <profiles>

    <publisher profile_name="publisher_profile" is_default_profile="true">
      <topic>
        <historyQos>
          <kind>KEEP_LAST</kind>
          <depth>1</depth>
        </historyQos>
      </topic>

      <qos> <!-- dataWriterQosPoliciesType -->
        <durability>
          <kind>VOLATILE</kind>
        </durability>
        <reliability>
          <kind>BEST_EFFORT</kind>
          <max_blocking_time>
            <sec>0</sec>
            <nanosec>1000000</nanosec>
          </max_blocking_time>
        </reliability>
      </qos>
      <historyMemoryPolicy>DYNAMIC</historyMemoryPolicy>
    </publisher>


    <subscriber profile_name="subscriber_profile" is_default_profile="true">
      <topic>
        <historyQos>
          <kind>KEEP_LAST</kind>
          <depth>1</depth>
        </historyQos>
      </topic>

      <qos> <!-- dataReaderQosPoliciesType -->
        <durability>
          <kind>VOLATILE</kind>
        </durability>
        <reliability>
          <kind>BEST_EFFORT</kind>
          <max_blocking_time>
            <sec>0</sec>
            <nanosec>1000000</nanosec>
          </max_blocking_time>
        </reliability>
      </qos>
      <historyMemoryPolicy>DYNAMIC</historyMemoryPolicy>
    </subscriber>

  </profiles>
</dds>
```

[...]

PR #503, opened 2021-01-15
Update README.md

briansoe66:
Clarify what QoS settings can be configured via the XML. 
This addresses #500

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,24 +62,24 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and 
 
 Fast DDS QoS policies can be fully configured through a combination of the [`rclcpp`/`rclpy` QoS] API, and the [Fast DDS XML] file's QoS elements. Configuraton depends on the environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML`.
 
-| RMW_FASTRTPS_USE_QOS_FROM_XML | [`rclcpp`/`rclpy` QoS]    | [Fast DDS XML] QoS                                               |
-| ----------------------------- | ------------------------- | ---------------------------------------------------------------- |
-| 0 (default)                   | Use default values        | Ignored - overridden by rclcpp/rclpy Qos                         |
-| 0 (default)                   | Set to non system default | Ignored - overridden by rclcpp/rclpy Qos                         |
-| 0 (default)                   | Set to system default     | Used                                                             |
-| 1                             | Use default values        | [History memory policy], [Publication mode] used, others ignored |
-| 1                             | Set to non system default | [History memory policy], [Publication mode] used, others ignored |
-| 1                             | Set to system default     | Used                                                             |
+| RMW_FASTRTPS_USE_QOS_FROM_XML | [`rclcpp`/`rclpy` QoS]    | [Fast DDS XML] QoS                   | [Fast DDS XML] [history memory policy] and [publication mode] |
+| ----------------------------- | ------------------------- | ------------------------------------ | ------------------------------------- |
+| 0 (default)                   | Use default values        | Ignored - overridden by rclcpp/rclpy | Ignored - overrided by `rmw_fastrtps` |
+| 0 (default)                   | Set to non system default | Ignored - overridden by rclcpp/rclpy | Ignored - overrided by `rmw_fastrtps` |
+| 0 (default)                   | Set to system default     | Used                                 | Ignored - overrided by `rmw_fastrtps` |
+| 1                             | Use default values        | Ignored - overridden by rclcpp/rclpy | Used                                  | 
+| 1                             | Set to non system default | Ignored - overridden by rclcpp/rclpy | Used                                  |
+| 1                             | Set to system default     | Used                                 | Used                                  |
 
 [`rclcpp`/`rclpy` QoS]: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/qos.hpp
 [Fast DDS XML]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html
-[History memory policy]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos
-[Publication mode]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy
+[history memory policy]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos
+[publication mode]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy
 
-Note: If `RMW_FASTRTPS_USE_QOS_FROM_XML` is set to 1, and [History memory policy] or [Publication mode] are not specified in the XML, then the Fast DDS' defaults configurations will be used: 
+Note: If `RMW_FASTRTPS_USE_QOS_FROM_XML` is set to 1, and [history memory policy] or [publication mode] are not specified in the XML, then the Fast DDS' defaults configurations will be used: 
 
-* [History memory policy] : `PREALLOCATED_MEMORY_MODE`.
-* [Publication mode] : `SYNCHRONOUS_PUBLISH_MODE`.
+* [history memory policy] : `PREALLOCATED_MEMORY_MODE`.
+* [publication mode] : `SYNCHRONOUS_PUBLISH_MODE`.
 
 There are two ways of telling a ROS 2 application which XML to use:
 
--- a/README.md
+++ b/README.md
@@ -60,13 +60,11 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and 
 
 ### Full QoS configuration
 
-With `rmw_fastrtps`, it is possible to fully configure Fast DDS using an XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
-When configuring the middleware using XML files, there are certain points that have to be taken into account:
+It is possible to full configure Fast DDS through a combination of 'rclcpp'/'rclpy' QoS, and the XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
 
-1. QoS set by `rclcpp`/`rclpy` are always honored.
-This means that setting any of them in the XML files has no effect, since they do not override what was used to create the publisher, subscription, service, or client.
-1. In order to modify the history memory policy or publication mode using XML files, environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` must be set to 1 (it is set to 0 by default).
-This tells `rmw_fastrtps` that it should override both the history memory policy and the publication mode using the XML.
+1. QoS configuration contained in `rclcpp`/`rclpy` QoS class, can only be set using `rclcpp`/`rclpy`.
+Setting any of them in the XML files has no effect, since they do not override what was used to create the publisher, subscription, service, or client.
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
LGTM with DCO

JLBuenoLopez-eProsima (commented):
I would suggest clarifying that the environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` only applies to override two specific QoS settings that are, by default, configured in the RMW layer. Those two QoS are the History Memory Policy and the Publication Mode. If you want to change the default behaviour of `rmw_fastrtps` you should set the environment variable.
However, there are other QoS not set in the `rclcpp`/`rclpy` layer that can be configured using the XML file without setting the environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML`.

README.md
```diff
@@ -60,13 +60,11 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
...
-When configuring the middleware using XML files, there are certain points that have to be taken into account:
+It is possible to fully configure Fast DDS through a combination of 'rclcpp'/'rclpy' QoS, and the XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
 
-1. QoS set by `rclcpp`/`rclpy` are always honored.
-This means that setting any of them in the XML files has no effect, since they do not override what was used to create the publisher, subscription, service, or client.
-1. In order to modify the history memory policy or publication mode using XML files, environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` must be set to 1 (it is set to 0 by default).
-This tells `rmw_fastrtps` that it should override both the history memory policy and the publication mode using the XML.
+1. QoS configuration contained in `rclcpp`/`rclpy` QoS class, can only be set using `rclcpp`/`rclpy`.
```
EduPonz, line 65: I'd add a link for reference here

README.md
```diff
@@ -60,13 +60,11 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
...
+Setting any of them in the XML files has no effect, since they do not override what was used to create the publisher, subscription, service, or client.
+1. QoS configuration not contained `rclcpp`/`rclpy`, such as the history memory policy or publication mode, can be overridden using XML files. The environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML` must be set to 1 (it is set to 0 by default).
```
EduPonz, line 67: Maybe a link to Fast DDS XML config documentation page would be good here

EduPonz (changes requested):

README.md
```diff
@@ -60,13 +60,11 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
 
 ### Full QoS configuration
 
-With `rmw_fastrtps`, it is possible to fully configure Fast DDS using an XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
-When configuring the middleware using XML files, there are certain points that have to be taken into account:
+It is possible to fully configure Fast DDS through a combination of 'rclcpp'/'rclpy' QoS, and the XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
```
hidmic, line 63: @briansoe66 here and below, I'd rather not mention an specific client library. There are others besides the core ones.

README.md
hidmic, line 65: @briansoe66 is this true? What if QoS policies are all set to `SYSTEM_DEFAULT` values?

EduPonz: You're actually correct @hidmic . [Here](https://github.com/ros2/rmw_fastrtps/blob/47be8ba4ed14303659e29094e72033b56449c15a/rmw_fastrtps_shared_cpp/src/qos.cpp#L55), you can see that in the case of `SYSTEM_DEFAULT`, the QoS are not overriden

hidmic: Awesome. @briansoe66 would you mind updating this patch to reflect that?

README.md
briansoe66: There is a link to the Fast DDS XML doc on line 63

hidmic:
Besides the failing DCO check, this patch still needs update @briansoe66.

briansoe66:
@hidmic , I am waiting for answers from @EduPonz to my three questions.

EduPonz:
Hi @briansoe66 , I'm afraid I must have missed them. Please, can you clarify which questions are you referring to?

briansoe66:
@EduPonz , you stated:
> You're actually correct @hidmic . Here, you can see that in the case of SYSTEM_DEFAULT, the QoS are not overriden

Could you clarify:
1. How can I set the QoS policy to use the SYSTEM_DEFAULT values?
2. In the case of SYSTEM_DEFAULT, does it use the default settings for Fast-RTPS? Like the default is up to the client implementation, and is not standardized by rclcpp / rclpy?
3. If so, doesn't that contradict the original readme that "QoS set by rclcpp/rclpy are always honored" (line 66)?

EduPonz:
Hi @briansoe66 ,

1. You can set the QoS to SYSTEM_DEFAULT using the `rclcpp`/`rclpy` API when creating the publisher/subscription. You can read more about that in:
    1. [ROS 2 QoS docs](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/)
    2. [RMW API reference](http://docs.ros2.org/latest/api/rmw/structrmw__qos__profile__t.html)
1. This is a default specified by ROS 2 (it's specified in the first link). Not explicitly modifying a field of `rmw_qos_profile_t` will result in using the ROS 2 default value. However, the user can let the implementation decide, which is what SYSTEM_DEFAULT means in this case.
1. You are right, the README was not correct. I corrected it yesterday in #504 . Now, it states:
```
ROS 2 QoS contained in rmw_qos_profile_t are always honored, unless set to *_SYSTEM_DEFAULT. In that case, XML values, or Fast DDS default values in the absences of XML ones, are applied. Setting any QoS in rmw_qos_profile_t to something other than *_SYSTEM_DEFAULT entails that specifying it via XML files has no effect, since they do not override what was used to create the publisher, subscription, service, or client.
```

I hope that soles your questions. Just contact me if I can provide further help. Thanks for the iterations and the patience!

README.md
```diff
@@ -60,27 +60,41 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
...
 1. Placing your XML file in the running directory under the name `DEFAULT_FASTRTPS_PROFILES.xml`.
 1. Setting environment variable `FASTRTPS_DEFAULT_PROFILES_FILE` to contain the path to your XML file (relative to the working directory).
 
+To verify the actual QoS settings:
+
+```cpp
+auto rcl_publisher = my_rclcpp_publisher->get_publisher_handle();
+const auto rmw_publisher = rcl_publisher_get_rmw_handle(rcl_publisher);
+rmw_qos_profile_t qos;
+rmw_publisher_get_actual_qos(rmw_publisher, &qos);
```
hidmic, line 95: @briansoe66 consider restricting the example to the `rmw` layer.

README.md
```diff
@@ -60,27 +60,41 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
 
 ### Full QoS configuration
 
-With `rmw_fastrtps`, it is possible to fully configure Fast DDS using an XML file as described in [Fast DDS documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html).
-When configuring the middleware using XML files, there are certain points that have to be taken into account:
+Fast DDS QoS policies can be fully configured through a combination of the [`rclcpp`/`rclpy` QoS] API, and the [Fast DDS XML] file's QoS elements. Configuraton depends on the environment variable `RMW_FASTRTPS_USE_QOS_FROM_XML`.
```
hidmic, line 63: @briansoe66 as I mentioned before, I'd rather not target specific client libraries. You can always link against the [`rmw`](https://github.com/ros2/rmw) interface layer for API documentation.

README.md
```diff
@@ -60,27 +60,41 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
...
-    * [Publication mode](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy): `SYNCHRONOUS_PUBLISH_MODE`.
+[`rclcpp`/`rclpy` QoS]: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/qos.hpp
+[Fast DDS XML]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html
+[history memory policy]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos
+[publication mode]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy
 
-1. Setting `RMW_FASTRTPS_USE_QOS_FROM_XML` effectively overrides whatever configuration was set with `RMW_FASTRTPS_PUBLICATION_MODE`, setting the publication mode to Fast DDS' default publication mode unless specified otherwise using an XML file.
+Note: If `RMW_FASTRTPS_USE_QOS_FROM_XML` is set to 1, and [history memory policy] or [publication mode] are not specified in the XML, then the Fast DDS' defaults configurations will be used:
```
hidmic, line 79: @briansoe66 nit: 
```suggestion
Note: If `RMW_FASTRTPS_USE_QOS_FROM_XML` is set to 1, and [history memory policy] or [publication mode] are not specified in the XML, then the Fast DDS' default configurations will be used: 
```

README.md
```diff
@@ -60,27 +60,41 @@ If `RMW_FASTRTPS_PUBLICATION_MODE` is not set, then both `rmw_fastrtps_cpp` and
...
 
-    * [History memory policy](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos): `PREALLOCATED_MEMORY_MODE`.
-    * [Publication mode](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy): `SYNCHRONOUS_PUBLISH_MODE`.
+[`rclcpp`/`rclpy` QoS]: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/qos.hpp
+[Fast DDS XML]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/xml_configuration.html
+[history memory policy]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#rtpsendpointqos
+[publication mode]: https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/core/policy/eprosimaExtensions.html#publishmodeqospolicy
 
-1. Setting `RMW_FASTRTPS_USE_QOS_FROM_XML` effectively overrides whatever configuration was set with `RMW_FASTRTPS_PUBLICATION_MODE`, setting the publication mode to Fast DDS' default publication mode unless specified otherwise using an XML file.
```
hidmic, line 77: @briansoe66 why remove this sentence?

hidmic (commented):
@briansoe66 I'd advise against referring to `rclcpp` directly (it's not the only client library, there are many more). Consider targeting [`rmw`'s interface](https://github.com/ros2/rmw) for API demonstration.

briansoe66:
@hidmic , I addressed your comments. Something is weird with the diffs though, not sure how to fix that.

I created this PR to clarify the nuances of setting QoS, by explicitly stating all different methods. But I cannot dedicate more time to this. You are welcome to take over the changes, or we can drop it.

README.md
hidmic: This still applies !

README.md
hidmic: This still applies !

EduPonz:
Hi @briansoe66 @hidmic ,

I can take over this if that's ok with you, although probably not today.

briansoe66:
@EduPonz , that would be great, thank you!

EduPonz:
I do not have privileges to push to @briansoe66 fork, so I've taken the commits onto a branch `eProsima:feature/briansoe66/readme` in eProsima's fork. I hope that's OK with everyone. This PR can be closed when merging #506

clalancette:
Since #506 has been merged, I'm going to close this out.

PR #513, opened 2021-02-25
Add content filter topic feature

iuhilnehc-ynos:
Related to https://github.com/ros2/design/pull/282 to add content-filtered-topic interfaces.

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -131,8 +131,8 @@ ament_export_dependencies(rosidl_typesupport_fastrtps_c)
 ament_export_dependencies(rosidl_typesupport_fastrtps_cpp)
 
 register_rmw_implementation(
-  "c:rosidl_typesupport_c:rosidl_typesupport_fastrtps_c"
-  "cpp:rosidl_typesupport_cpp:rosidl_typesupport_fastrtps_cpp")
+  "c:rosidl_typesupport_c:rosidl_typesupport_fastrtps_c:rosidl_typesupport_introspection_c"
+  "cpp:rosidl_typesupport_cpp:rosidl_typesupport_fastrtps_cpp:rosidl_typesupport_introspection_cpp")
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
--- a/rmw_fastrtps_cpp/src/publisher.cpp
+++ b/rmw_fastrtps_cpp/src/publisher.cpp
@@ -199,8 +199,7 @@ rmw_fastrtps_cpp::create_publisher(
   }
   info->type_support_ = fastdds_type;
 
-  if (!rmw_fastrtps_shared_cpp::register_type_object(type_supports, type_name))
-  {
+  if (!rmw_fastrtps_shared_cpp::register_type_object(type_supports, type_name)) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "failed to register type object with incompatible type %s",
       type_name.c_str());
--- a/rmw_fastrtps_cpp/src/publisher.cpp
+++ b/rmw_fastrtps_cpp/src/publisher.cpp
@@ -199,6 +199,14 @@ rmw_fastrtps_cpp::create_publisher(
   }
   info->type_support_ = fastdds_type;
 
+  if (!rmw_fastrtps_shared_cpp::register_type_object(type_supports, type_name))
+  {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "failed to register type object with incompatible type %s",
+      type_name.c_str());
+    return nullptr;
+  }
+
   /////
   // Create Listener
   if (create_publisher_listener) {
--- a/rmw_fastrtps_cpp/src/rmw_subscription.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_subscription.cpp
@@ -149,9 +149,9 @@ rmw_subscription_get_actual_qos(
 }
 
 rmw_ret_t
-rmw_subscription_set_cft_expression_parameters(
+rmw_subscription_set_content_filter(
   rmw_subscription_t * subscription,
-  const rmw_subscription_content_filtered_topic_options_t * options)
+  const rmw_subscription_content_filter_options_t * options)
 {
   // Unused in current implementation.
   (void) subscription;
@@ -161,10 +161,10 @@ rmw_subscription_set_cft_expression_parameters(
 }
 
 rmw_ret_t
-rmw_subscription_get_cft_expression_parameters(
+rmw_subscription_get_content_filter(
   const rmw_subscription_t * subscription,
   rcutils_allocator_t * allocator,
-  rmw_subscription_content_filtered_topic_options_t * options)
+  rmw_subscription_content_filter_options_t * options)
 {
   // Unused in current implementation.
   (void) subscription;
--- a/rmw_fastrtps_cpp/src/rmw_subscription.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_subscription.cpp
@@ -151,13 +151,11 @@ rmw_subscription_get_actual_qos(
 rmw_ret_t
 rmw_subscription_set_cft_expression_parameters(
   rmw_subscription_t * subscription,
-  const char * filter_expression,
-  const rcutils_string_array_t * expression_parameters)
+  const rmw_subscription_content_filtered_topic_options_t * options)
 {
   // Unused in current implementation.
   (void) subscription;
-  (void) filter_expression;
-  (void) expression_parameters;
+  (void) options;
   RMW_SET_ERROR_MSG("unimplemented");
   return RMW_RET_UNSUPPORTED;
 }
@@ -165,13 +163,13 @@ rmw_subscription_set_cft_expression_parameters(
 rmw_ret_t
 rmw_subscription_get_cft_expression_parameters(
   const rmw_subscription_t * subscription,
-  char ** filter_expression,
-  rcutils_string_array_t * expression_parameters)
+  rcutils_allocator_t * allocator,
+  rmw_subscription_content_filtered_topic_options_t * options)
 {
   // Unused in current implementation.
   (void) subscription;
-  (void) filter_expression;
-  (void) expression_parameters;
+  (void) allocator;
+  (void) options;
   RMW_SET_ERROR_MSG("unimplemented");
   return RMW_RET_UNSUPPORTED;
 }
--- a/rmw_fastrtps_cpp/src/rmw_subscription.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_subscription.cpp
@@ -150,7 +150,7 @@ rmw_subscription_get_actual_qos(
 
 rmw_ret_t
 rmw_subscription_set_cft_expression_parameters(
-  const rmw_subscription_t * subscription,
+  rmw_subscription_t * subscription,
   const char * filter_expression,
   const rcutils_string_array_t * expression_parameters)
 {
--- a/rmw_fastrtps_cpp/src/rmw_subscription.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_subscription.cpp
@@ -149,6 +149,34 @@ rmw_subscription_get_actual_qos(
 }
 
 rmw_ret_t
+rmw_subscription_set_cft_expression_parameters(
+  const rmw_subscription_t * subscription,
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters)
+{
+  // Unused in current implementation.
+  (void) subscription;
+  (void) filter_expression;
+  (void) expression_parameters;
+  RMW_SET_ERROR_MSG("unimplemented");
+  return RMW_RET_UNSUPPORTED;
+}
+
+rmw_ret_t
+rmw_subscription_get_cft_expression_parameters(
+  const rmw_subscription_t * subscription,
+  char ** filter_expression,
+  rcutils_string_array_t * expression_parameters)
+{
+  // Unused in current implementation.
+  (void) subscription;
+  (void) filter_expression;
+  (void) expression_parameters;
+  RMW_SET_ERROR_MSG("unimplemented");
+  return RMW_RET_UNSUPPORTED;
+}
+
+rmw_ret_t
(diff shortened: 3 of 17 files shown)


fujitatomoya (approved):

MiguelCompany (approved):
LGTM

rmw_fastrtps_cpp/src/rmw_subscription.cpp
```diff
@@ -148,6 +148,32 @@ rmw_subscription_get_actual_qos(
...
 
+rmw_ret_t
+rmw_subscription_set_content_filter(
+  rmw_subscription_t * subscription,
+  const rmw_subscription_content_filter_options_t * options)
+{
+  // Unused in current implementation.
+  (void) subscription;
+  (void) options;
+  RMW_SET_ERROR_MSG("unimplemented");
```
ivanpauno, line 159: what will happen in rclcpp when the rmw implementation doesn't support content filtering?

fujitatomoya: `NOT SUPPORT` will be returned to application for now. @iuhilnehc-ynos if i am mistaken, let us know your thought.

iuhilnehc-ynos: As `rclcpp` always did, it'll throw an exception (RCLError, I will add a new exception named RCLUnsupport) if users call `set_content_filter` without checking `is_cft_enabled`.

https://github.com/ros2/rclcpp/blob/f509a3469b790fd80311a89e2931b420e7ca9a16/rclcpp/src/rclcpp/subscription_base.cpp#L369-L375

iuhilnehc-ynos:
@MiguelCompany 

> @iuhilnehc-ynos I think we should change [this](https://github.com/ros2/rmw_fastrtps/blob/bde606898a97f70a3fd454310424275d81972074/rmw_fastrtps_cpp/CMakeLists.txt#L133) to look similar to [this](https://github.com/ros2/rmw_connextdds/blob/35a37aa0a227849467c8302d7c52a4fd2cab976a/rmw_connextdds/CMakeLists.txt#L52)

The `rosidl_typesupport_introspection_c/cpp` is only used in the `rmw_fastrtps_shared_cpp`, so I just added the necessary information in https://github.com/ros2/rmw_fastrtps/blob/83f90452bc89864595fd83608bc8cf11392cf221/rmw_fastrtps_shared_cpp/CMakeLists.txt.

MiguelCompany:
@iuhilnehc-ynos 
> The `rosidl_typesupport_introspection_c/cpp` is only used in the `rmw_fastrtps_shared_cpp`, so I just added the necessary information in https://github.com/ros2/rmw_fastrtps/blob/83f90452bc89864595fd83608bc8cf11392cf221/rmw_fastrtps_shared_cpp/CMakeLists.txt.

That is correct, but `register_rmw_implementation` should be called with the corresponding type supports. This is the way to inform the building infrastructure what type supports should be generated for each message definition. Failing to do so, we may end in a situation where the introspection type support is not generated.

rmw_fastrtps_cpp/CMakeLists.txt
```diff
@@ -130,8 +130,8 @@ ament_export_dependencies(rosidl_typesupport_fastrtps_c)
 ament_export_dependencies(rosidl_typesupport_fastrtps_cpp)
 
 register_rmw_implementation(
-  "c:rosidl_typesupport_c:rosidl_typesupport_fastrtps_c"
-  "cpp:rosidl_typesupport_cpp:rosidl_typesupport_fastrtps_cpp")
+  "c:rosidl_typesupport_c:rosidl_typesupport_fastrtps_c:rosidl_typesupport_introspection_c"
+  "cpp:rosidl_typesupport_cpp:rosidl_typesupport_fastrtps_cpp:rosidl_typesupport_introspection_cpp")
```
ivanpauno, line 134: is this related to the PR?

It would be good to create another one if not

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -199,6 +199,13 @@ rmw_fastrtps_cpp::create_publisher(
   }
   info->type_support_ = fastdds_type;
 
+  if (!rmw_fastrtps_shared_cpp::register_type_object(type_supports, type_name)) {
```
ivanpauno, line 202: is this related to the PR?
could you explain how if it is?

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -225,8 +232,30 @@ create_subscription(
...
   des_topic = topic.desc;
 
+  // Create ContentFilteredTopic
+  if (subscription_options->content_filter_options) {
+    rmw_subscription_content_filter_options_t * options =
+      subscription_options->content_filter_options;
+    if (nullptr != options->filter_expression) {
+      eprosima::fastdds::dds::ContentFilteredTopic * filtered_topic = nullptr;
+      if (!rmw_fastrtps_shared_cpp::create_content_filtered_topic(
+          dds_participant, des_topic,
```
ivanpauno, line 248: is in all DDS implementations possible to create a "content filtered topic" from a topic?

The proposed API forces all implementations that want to support content filtered topics to support them in any subscription.
Maybe a more limited API, with a different type for content filtered subscriptions would make things easier to other DDS implementations or nonDDS rmw implementations.

rmw_fastrtps_cpp/CMakeLists.txt
MiguelCompany: See my comment below

rmw_fastrtps_cpp/src/subscription.cpp
MiguelCompany: This is the interface required by the DDS standard. The user calls `create_contentfilteredtopic` on the participant, which is specified as:

```
ContentFilteredTopic create_contentfilteredtopic(
 in string name,
 in Topic related_topic,
 in string filter_expression,
 in StringSeq expression_parameters);
```

```
2.2.2.2.1.7 create_contentfilteredtopic
This operation creates a ContentFilteredTopic. As described in 2.2.2.3, Topic-Definition Module, a ContentFilteredTopic can 
be used to do content-based subscriptions. 
The related Topic being subscribed to is specified by means of the related_topic parameter. The ContentFilteredTopic only 
relates to samples published under that Topic, filtered according to their content. The filtering is done by means of evaluating a 
logical expression that involves the values of some of the data-fields in the sample. The logical expression is derived from the 
filter_expression and expression_parameters arguments. 
The syntax of the filter expression and parameters is described in Annex B.
In case of failure, the operation will return a ‘nil’ value (as specified by the platform)
```

rmw_fastrtps_cpp/src/publisher.cpp
ivanpauno: Ok, that sounds fine, I will have to take a deeper look at the code.

MiguelCompany: Just FYI, @iuhilnehc-ynos took inspiration from the auto-generated code of the Fast DDS unit tests [here](https://github.com/eProsima/Fast-DDS/tree/master/test/unittest/dds/topic/DDSSQLFilter/data_types)

rmw_fastrtps_cpp/src/subscription.cpp
ivanpauno: > This is the interface required by the DDS standard. The user calls create_contentfilteredtopic on the participant, which is specified as

Nice!
If upgrading a `DataReader` to a `ContentFilteredDataReader` is required by the DDS standard this is less of a concern, but it might still limit future non-DDS based rmw implementations.

rmw_fastrtps_cpp/src/publisher.cpp
ivanpauno: > Just FYI, @iuhilnehc-ynos took inspiration from the auto-generated code of the Fast DDS unit tests [here](https://github.com/eProsima/Fast-DDS/tree/master/test/unittest/dds/topic/DDSSQLFilter/data_types)

Could you point out what part of the generated code?

rmw_fastrtps_dynamic_cpp/src/rmw_subscription.cpp
```diff
@@ -151,6 +154,32 @@ rmw_subscription_get_actual_qos(
   return rmw_fastrtps_shared_cpp::__rmw_subscription_get_actual_qos(subscription, qos);
 }
 
+rmw_ret_t
+rmw_subscription_set_content_filter(
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Yes, it should and will be supported in the `rmw_fastrtps_dynamic_cpp`.

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
```diff
@@ -106,4 +110,155 @@ __rmw_subscription_get_actual_qos(
...
+)
+{
+  auto info = static_cast<CustomSubscriberInfo *>(subscription->data);
+  eprosima::fastdds::dds::ContentFilteredTopic * filtered_topic = info->filtered_topic_;
+  const bool filter_expression_empty = (*options->filter_expression == '\0');
+
+  if (!filtered_topic && filter_expression_empty) {
+    // can't reset current subscriber
+    RMW_SET_ERROR_MSG(
+      "current subscriber has no content filter topic");
```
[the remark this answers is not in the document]
iuhilnehc-ynos: > so instead of getting a runtime error you get a compile time error.

What do you mean?

rmw_fastrtps_cpp/src/publisher.cpp
iuhilnehc-ynos: https://github.com/eProsima/Fast-DDS/blob/caa3fe0ff9dc6a0ec31b08b0e44ab45d4892d7c5/test/unittest/dds/topic/DDSSQLFilter/data_types/ContentFilterTestTypeTypeObject.cxx

rmw_fastrtps_shared_cpp/src/rmw_subscription.cpp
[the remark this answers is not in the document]
ivanpauno: If the prototype would be:

```c
rmw_subscription_set_content_filter(
  rmw_cft_subscription_t * subscription,
  const rmw_subscription_content_filter_options_t * options
)
```

(note `rmw_cft_subscription_t` instead of `rmw_subscription_t`)
you wouldn't have to check this at runtime, it would be a compile time error.

i.e. related to the discussion [here](https://github.com/ros2/rmw/pull/302#discussion_r823791348)

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
```diff
@@ -123,4 +141,450 @@ std::function<uint32_t()> TypeSupport::getSerializedSizeProvider(void * data)
...
+using MinimalStructType = eprosima::fastrtps::types::MinimalStructType;
+using MinimalStructMember = eprosima::fastrtps::types::MinimalStructMember;
+using SerializedPayload_t = eprosima::fastrtps::rtps::SerializedPayload_t;
+using TypeNamesGenerator = eprosima::fastrtps::types::TypeNamesGenerator;
+using TypeIdentifier = eprosima::fastrtps::types::TypeIdentifier;
+using TypeObject = eprosima::fastrtps::types::TypeObject;
+using TypeObjectFactory = eprosima::fastrtps::types::TypeObjectFactory;
+
+const rosidl_message_type_support_t *
+get_type_support_intro(
```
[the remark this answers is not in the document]
iuhilnehc-ynos: updated in 3510ebfe2e320fc597977d37ecb7c553b8c22134

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
```diff
@@ -123,4 +141,450 @@ std::function<uint32_t()> TypeSupport::getSerializedSizeProvider(void * data)
...
+  return ss.str();
+}
+
+typedef std::pair<const TypeIdentifier *, std::string> MemberIdentifierName;
+
+template<typename MembersType>
+MemberIdentifierName GetTypeIdentifier(const MembersType * member, uint32_t index, bool complete);
+
+template<typename MembersType>
+const TypeObject * GetCompleteObject(
```
ivanpauno, line 211: Why is all this magic needed?
If xtypes would be enabled, would be need this?

ivanpauno (approved):
LGTM!

The implementation is complex and I didn't fully undestand why we need to make `rosidl_typesupport_introspection` mandatory also for `rmw_fastrtps_cpp`.
How does DDS solve this? Do they use xtypes?

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
MiguelCompany: We need `rosidl_typesupport_introspection` to generate the xtypes information (a.k.a TypeObject/TypeIdentifier) so the content filter logic inside Fast DDS can calculate the offsets in the serialized payload to the fields of the type used in the filter expression.

So, in a way, what we are doing is *enabling* xtypes.

MiguelCompany: Side note: we didn't do this on `rosidl_typesupport_fastrtps` because it is also used on `rmw_connextdds`, and it only depends on Fast CDR.

Also because supporting content filter on `rmw_fastrtps_dynamic_cpp` would also require the TypeObject generation.

That's why we put the code on the common package for both (i.e. `rmw_fastrtps_shared_cpp`)

ivanpauno: To see if I understand correctly, the generated code of `rosidl_typesupport_fastrtps` doesn't provide a typeobject, and we're dynamically creating one from the instronspection typesupport.

That seems okay, though it sounds like something that would better be done in the generated code.
If there would be a common API for generating type objects for both fastrtps and connext maybe the typeobject could also be provided by the generated code.

MiguelCompany: Correct. But there's no common API, and `rmw_fastrtps_dynamic_cpp` depends only on `rosidl_typesupport_introspection`.

fujitatomoya:
CI is green, i will merge this into mainline.

- https://github.com/ros2/rmw/pull/302#issuecomment-1072005358
- https://github.com/ros2/rmw/pull/302#issuecomment-1072572088

PR #536, opened 2021-05-31
Support for SubscriptionOptions::ignore_local_publications

jamoralp:
Signed-off-by: Jose Antonio Moral <joseantoniomoralparras@gmail.com>

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -83,6 +83,15 @@ _take(
   data.data = ros_message;
   data.impl = info->type_support_impl_;
   if (info->data_reader_->take_next_sample(&data, &sinfo) == ReturnCode_t::RETCODE_OK) {
+    if (subscription->options.ignore_local_publications) {
+      auto sample_writer_guid =
+        eprosima::fastrtps::rtps::iHandle2GUID(sinfo.publication_handle);
+
+      if (sample_writer_guid.guidPrefix == info->data_reader_->guid().guidPrefix) {
+        // This is a local publication. Ignore it
+        return RMW_RET_OK;
+      }
+    }
     // Update hasData from listener
     info->listener_->update_has_data(info->data_reader_);
 
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -82,24 +82,29 @@ _take(
   data.is_cdr_buffer = false;
   data.data = ros_message;
   data.impl = info->type_support_impl_;
-  if (info->data_reader_->take_next_sample(&data, &sinfo) == ReturnCode_t::RETCODE_OK) {
-    if (subscription->options.ignore_local_publications) {
-      auto sample_writer_guid =
-        eprosima::fastrtps::rtps::iHandle2GUID(sinfo.publication_handle);
 
-      if (sample_writer_guid.guidPrefix == info->data_reader_->guid().guidPrefix) {
-        // This is a local publication. Ignore it
-        return RMW_RET_OK;
-      }
-    }
-    // Update hasData from listener
-    info->listener_->update_has_data(info->data_reader_);
+  while (0 < info->data_reader_->get_unread_count()) {
+    if (info->data_reader_->take_next_sample(&data, &sinfo) == ReturnCode_t::RETCODE_OK) {
+      // Update hasData from listener
+      info->listener_->update_has_data(info->data_reader_);
 
-    if (sinfo.valid_data) {
-      if (message_info) {
-        _assign_message_info(identifier, message_info, &sinfo);
+      if (subscription->options.ignore_local_publications) {
+        auto sample_writer_guid =
+          eprosima::fastrtps::rtps::iHandle2GUID(sinfo.publication_handle);
+
+        if (sample_writer_guid.guidPrefix == info->data_reader_->guid().guidPrefix) {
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -76,6 +76,15 @@ _take(
   data.data = ros_message;
   data.impl = info->type_support_impl_;
   if (info->data_reader_->take_next_sample(&data, &sinfo) == ReturnCode_t::RETCODE_OK) {
+    if (subscription->options.ignore_local_publications) {
```
hidmic, line 79: @jamoralp meta: won't this result in waitsets waking up when there's no need to? Can we avoid that entirely?

hidmic (commented):
@jamoralp A test for this in [`test_rmw_implementation`](https://github.com/ros2/rmw_implementation/blob/master/test_rmw_implementation/test/test_subscription.cpp) would be ideal. Would you mind adding it?

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
fujitatomoya: i am with @hidmic , i think this would affect performance. if there are many local publications, that means there are unnecessary executor loop to take & just drop the data? if i am not mistake.

MiguelCompany: As the reader may have data queued from both local and remote publications, samples from local ones need to be consumed (i.e. taken). Fast DDS doesn't (yet) have a mechanism to filter samples before adding them to the history, which would be the ideal way of doing this. The same thing is done on [rmw_connextdds](https://github.com/ros2/rmw_connextdds/blob/721976a825360b260c2b8adc0e6a905a6ee508a9/rmw_connextdds_common/src/common/rmw_impl.cpp#L1643-L1654)

MiguelCompany: I would make additional changes here, though.

First, `info->listener_->update_has_data(info->data_reader_)` should always be called after successfully taking a sample.
Second, I would wrap the code inside a `while (0 < info->data_reader_->get_unread_count())`, with a `break;` when we do `*taken = true;` and a `continue;` when a local publication sample is filtered.

jamoralp: Changes addressed :+1:

jamoralp:
> @jamoralp A test for this in [`test_rmw_implementation`](https://github.com/ros2/rmw_implementation/blob/master/test_rmw_implementation/test/test_subscription.cpp) would be ideal. Would you mind adding it?

Sure, I will take a look at it as soon as I can

jamoralp:
@MiguelCompany @fujitatomoya  @hidmic Please re-review. After this gets approved, I will update #537 and #535 accordingly. Thanks!

hidmic (approved):
LGTM pending green CI once we have that test.

fujitatomoya (approved):
lgtm

jamoralp:
Hello @hidmic @fujitatomoya I opened the pull request for testing this options in the file mentioned above, but no CI seems to be triggered. Also I am concerned that the CI will fail until this is merged, since I see no way of specifying to which branch/PR the `rmw_fastrtps` package should be checked out before running the test suite. Could you please give me more hints about this? Thanks.

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
hidmic: From [ros2/rmw_implementation (comment)](https://github.com/ros2/rmw_implementation/pull/192/files/254b69002e077fb9d9857f5c2bf7ee2a7a044274..f41148341892715da35bd7e0a33476539b1d4d8b#r646642889) I get that `rmw_cyclonedds_cpp` blocks message reception entirely when `ignore_local_publications` is set to true.

I'm not super thrilled about this variability. At the very least, I think it should be reflected in the `rmw` API contract for [ignore_local_publications](https://github.com/ros2/rmw/blob/a50f3bb1e24ac5338677990d74ab096a9d1776d7/rmw/include/rmw/types.h#L157-L170).

MiguelCompany:
@hidmic Anything pending here? Could we have the CI launched?

hidmic:
CI up to `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14752/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9502/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12416/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14899/)

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
hidmic: It'd be good to address this ^ at some point.

MiguelCompany:
@hidmic This PR was not rebased on top of master (thus the build failures). Please run CI again.

hidmic:
CI up to  `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14779/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9528/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12445/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14926/)

hidmic:
Alright, going in with this one. @jamoralp consider backporting the very same squashed commit.

PR #547, opened 2021-07-09
[Galactic] Loan messages implementation

MiguelCompany:
This is a backport of #523 into galactic without the need of ros2/rosidl_typesupport_fastrtps#67

### DIFF
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -47,6 +47,7 @@ void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 
   // Total size is encapsulation size + data size
   m_typeSize = 4 + data_size;
+  // Account for RTPS submessage alignment
   m_typeSize = (m_typeSize + 3) & ~3;
 }
 
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -27,22 +27,15 @@ TypeSupport::TypeSupport()
 {
   m_isGetKeyDefined = false;
   max_size_bound_ = false;
-  is_plain_ = false;
 }
 
 void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 {
   members_ = members;
 
-#ifdef ROSIDL_TYPESUPPORT_FASTRTPS_HAS_PLAIN_TYPES
-  char bounds_info;
-  auto data_size = static_cast<uint32_t>(members->max_serialized_size(bounds_info));
-  max_size_bound_ = 0 != (bounds_info & ROSIDL_TYPESUPPORT_FASTRTPS_BOUNDED_TYPE);
-  is_plain_ = bounds_info == ROSIDL_TYPESUPPORT_FASTRTPS_PLAIN_TYPE;
-#else
+  // Fully bound by default
   max_size_bound_ = true;
   auto data_size = static_cast<uint32_t>(members->max_serialized_size(max_size_bound_));
-#endif
 
   // A fully bound message of size 0 is an empty message
   if (max_size_bound_ && (data_size == 0) ) {
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -47,6 +47,7 @@ void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 
   // Total size is encapsulation size + data size
   m_typeSize = 4 + data_size;
+  m_typeSize = (m_typeSize + 3) & ~3;
 }
 
 size_t TypeSupport::getEstimatedSerializedSize(const void * ros_message, const void * impl) const
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -27,15 +27,22 @@ TypeSupport::TypeSupport()
 {
   m_isGetKeyDefined = false;
   max_size_bound_ = false;
+  is_plain_ = false;
 }
 
 void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 {
   members_ = members;
 
-  // Fully bound by default
+#ifdef ROSIDL_TYPESUPPORT_FASTRTPS_HAS_PLAIN_TYPES
+  char bounds_info;
+  auto data_size = static_cast<uint32_t>(members->max_serialized_size(bounds_info));
+  max_size_bound_ = 0 != (bounds_info & ROSIDL_TYPESUPPORT_FASTRTPS_BOUNDED_TYPE);
+  is_plain_ = bounds_info == ROSIDL_TYPESUPPORT_FASTRTPS_PLAIN_TYPE;
+#else
   max_size_bound_ = true;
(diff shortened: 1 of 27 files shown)


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -50,6 +56,7 @@ struct CustomSubscriberInfo : public CustomEventInfo
   const void * type_support_impl_{nullptr};
   rmw_gid_t subscription_gid_{};
   const char * typesupport_identifier_{nullptr};
+  std::shared_ptr<rmw_fastrtps_shared_cpp::LoanManager> loan_manager_;
```
hidmic, line 59: @MiguelCompany I know these bits are in a grey area when it comes to code visibility but this isn't ABI compatible, strictly speaking.

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -353,4 +360,133 @@ __rmw_take_serialized_message_with_info(
...
+  RMW_CHECK_ARGUMENT_FOR_NULL(loaned_message, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(taken, RMW_RET_INVALID_ARGUMENT);
+
+  auto info = static_cast<CustomSubscriberInfo *>(subscription->data);
+  auto loan_mgr = info->loan_manager_;
+  std::unique_lock<std::mutex> guard(loan_mgr->mtx);
+  auto item = loan_mgr->items.emplace_back();
+  if (nullptr == item) {
+    RMW_SET_ERROR_MSG("Out of resources for loaned message info");
+    return RMW_RET_ERROR;
```
hidmic, line 434: @MiguelCompany nit: `RMW_RET_BAD_ALLOC`?

hidmic (commented):
Overall LGTM. I'd like to hear @cottsay thoughts on backporting this though.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
MiguelCompany: It is, strictly speaking. But `rmw_fastrtps_shared_cpp` is designed to be only used by `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp`, which this PR is also modifying. This means ABI is kept to the users of the RMW.

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
MiguelCompany: Done in 1332adb6d2704e56e476d2182a5c458599785911

MiguelCompany:
@hidmic @cottsay This is ready for review

rmw_fastrtps_cpp/src/type_support_common.cpp
```diff
@@ -47,6 +47,7 @@ void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 
   // Total size is encapsulation size + data size
   m_typeSize = 4 + data_size;
+  m_typeSize = (m_typeSize + 3) & ~3;
```
hidmic, line 50: @MiguelCompany mind to explain this change?

MiguelCompany: @hidmic 
**TL/DR:** Account for RTPS sub-message alignment

The DDSI-RTPS spec requires each submessage to be aligned on a 4-byte boundary. This means that, when using `PREALLOCATED` memory policy, we could receive a serialized payload bigger than `m_typeSize`, which would thus be discarded. When using the rmw's default `PREALLOCATED_WITH_REALLOC` memory policy, an unnecessary reallocation could happen.

MiguelCompany: I added some comments on f40ff2d

hidmic: I see. How is this not a problem on `master`? Upstream Fast-DDS differences?

hidmic:
Galactic CI up to `test_rmw_implementation` and `rmw_fastrtps_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14934/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9676/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12610/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15098/)

rmw_fastrtps_cpp/src/type_support_common.cpp
MiguelCompany: @hidmic You are right! It is a problem on `master`. I've created #550 with the proper fix.

hidmic:
Alright, this is good to go. @cottsay ?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
cottsay: We need a clearer definition of "public" here. Please correct me if I'm mistaken but as I understand it, this will _technically_ break ABI compatibility even if you don't intend for this struct to be part of the formal `rmw_fastrtps*` public API. By that, I mean that if we publish this, and I `apt update ros-rolling-rmw-fastrtps-shared-cpp` without updating the consumers of that package, the consumers won't allocate adequate space for this struct but code in `rmw_fastrtps_shared_cpp` will try to access it anyway.

One can blur the lines of what constitutes "public API" by claiming that not everything exposed in the headers are formally part of the API, but the linker doesn't care what we say. ABI is ABI, and when you change the size of a struct, it is no longer compatible.

I'm going to bring this up at the next ROS 2 meeting so that we can get some solid documentation on the matter and clearly define our goals regarding ABI stability within a rosdistro so that we can come up with some criteria for when we can break it.

cottsay:
Beyond the one ABI break discussed above, this seems like a straightforward feature addition, so :+1: from me. I'd like to hold off on merging until we discuss it at the ROS 2 meeting on Tuesday, however.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
MiguelCompany: This new field is added at the end, so offsets of the old fields are not modified. The new field is only being accessed from new functions of `rmw_fastrtps_shared_cpp`. This means that consumers which are not updated will not be calling code accessing the new field.

cottsay: We discussed this at the ROS 2 meeting, and it sounds like stability among arbitrary combinations of package updates isn't part of our ABI stability goal, so my concerns are no longer valid. Thanks for your patience.

cottsay (approved):

hidmic:
Alright. Going in! Thanks for double-checking @cottsay.

PR #550, opened 2021-07-27
Fix type size alignment

MiguelCompany:
Account for possible RTPS submessage alignment.

The DDSI-RTPS spec requires each submessage to be aligned on a 4-byte boundary. This means that, when using `PREALLOCATED` memory policy, we could receive a serialized payload bigger than `m_typeSize`, which would thus be discarded. When using the rmw's default `PREALLOCATED_WITH_REALLOC` memory policy, an unnecessary reallocation could happen.

### DIFF
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -54,6 +54,7 @@ void TypeSupport::set_members(const message_type_support_callbacks_t * members)
 
   // Total size is encapsulation size + data size
   m_typeSize = 4 + data_size;
+  m_typeSize = (m_typeSize + 3) & ~3;
 }
 
 size_t TypeSupport::getEstimatedSerializedSize(const void * ros_message, const void * impl) const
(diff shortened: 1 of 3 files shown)


hidmic:
CI up to `test_rmw_implementation` and `test_communication`, against both Fast-DDS based RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9679/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12613/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15102/)

hidmic (approved):
LGTM! I'd love a regression test, but it's perhaps a difficult one.

MiguelCompany:
> I'd love a regression test, but it's perhaps a difficult one.

I agree it would be nice, but I think it would be too difficult to do. I think it will always have false positives.

hidmic:
Alright, going in!

PR #552, opened 2021-08-13
Update the LoanManager to do internal locking.

clalancette:
The original motivation to do this is to please the clang
thread safety analysis, which can't quite figure out the
locking regime with an external mutex.  Instead, move major
functionality into the LoanManager class, and then mark
the item list as private.

One thing to note is that the logic of
__rmw_take_loaned_message_internal is slightly changed here.
Before it would emplace an Item onto the LoanManager's list,
then look through the data_reader to see if it could take an
item.  If it could, it would just fill that Item in and return.
If it couldn't, it would pop that item out of the list.  This
commit changes it so that we instantiate a temporary Item.
If we can take an item, we then push that Item into the LoanManager
list.  We never pop it.  This should be equivalent (and
slightly more performant), but is a change from what we had
before.

This should fix the warnings we see when building with clang, like in https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/878/clang/new/ .

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <memory>
+#include <utility>
 
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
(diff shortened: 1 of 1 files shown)


clalancette:
Since we are touching a fairly low-level component, full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15050/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9784/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12737/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15218/)

fujitatomoya (approved):
this looks nice to me. it does not acquire lock until it needs to do so.

fujitatomoya:
https://ci.ros2.org/job/ci_linux/15050/testReport/junit/ros2param.ros2param.test/test_verb_dump/test_verb_dump/ can be reproducible local environment because of https://github.com/ros2/ros2cli/commit/0075a7174f4cb9a9a57efaa9811c314dd3bbb166

clalancette:
> https://ci.ros2.org/job/ci_linux/15050/testReport/junit/ros2param.ros2param.test/test_verb_dump/test_verb_dump/ can be reproducible local environment because of [ros2/ros2cli@0075a71](https://github.com/ros2/ros2cli/commit/0075a7174f4cb9a9a57efaa9811c314dd3bbb166)

Yeah, I agree; we see those failures in the nightlies: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/1688/testReport/junit/ros2param.ros2param.test/test_verb_dump/test_verb_dump/

I'm going to go ahead and merge this one.  Thanks for the review.

PR #558, opened 2021-09-08
Update rmw_context_impl_t definition

hidmic:
Connected to https://github.com/ros2/rmw/pull/313.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
@@ -17,6 +17,7 @@
 
 #include <mutex>
 
+// Definition of struct rmw_context_impl_s as declared in rmw/init.h
 struct rmw_context_impl_s
 {
   /// Pointer to `rmw_dds_common::Context`.
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
@@ -17,7 +17,7 @@
 
 #include <mutex>
 
-struct rmw_context_impl_t
+struct rmw_context_impl_s
 {
   /// Pointer to `rmw_dds_common::Context`.
   void * common;


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
```diff
@@ -17,7 +17,7 @@
 
 #include <mutex>
 
-struct rmw_context_impl_t
+struct rmw_context_impl_s
```
clalancette, line 20: Ah, I see what is happening here.

Over in https://github.com/ros2/rmw/pull/313/files#diff-0460a36c52d1df8664f7ab0ee834d07ac10cfaf17fda75f7477421e4aadf30d8R34, we are doing `typedef struct rmw_context_impl_s rmw_context_impl_t`, which we are then using later on.  So here in the RMW implementation, we need to define this as `struct rmw_context_impl_s`, but *not* make it `struct rmw_context_impl_t` (since that is done at the `rmw` layer, not the implementation layer).

I guess we do this in all of the RMW implementations, so this makes sense.  It also means that, in theory, we could write a RMW layer purely in C (though I doubt we would ever do that).

This change is OK with me, but this is definitely going to have repercussions for RMW implementations that are not in the default build.  We'll want to put a release note in place for this.  I also think we should probably put a comment here explaining that this is the implementation for the stuff over in `rmw/init.h`, since understanding this is pretty tricky.

clalancette (changes requested):

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/rmw_context_impl.hpp
hidmic: See https://github.com/ros2/rmw_fastrtps/pull/558/commits/e2b22c7130a4d67188399a2166f185eefb37d688.

clalancette (approved):

hidmic:
Full CI (from https://github.com/ros2/rmw/pull/313):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9938/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12898/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15400/)

PR #574, opened 2022-01-07
Fix cpplint error

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
@@ -21,7 +21,7 @@
 #include <fastdds/dds/subscriber/qos/DataReaderQos.hpp>
 #include <fastdds/dds/topic/qos/TopicQos.hpp>
 
-#include <fastrtps/qos/QosPolicies.h>
+#include "fastrtps/qos/QosPolicies.h"
 
 #include "rmw/rmw.h"
 


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
```diff
@@ -21,7 +21,7 @@
 #include <fastdds/dds/subscriber/qos/DataReaderQos.hpp>
 #include <fastdds/dds/topic/qos/TopicQos.hpp>
 
-#include <fastrtps/qos/QosPolicies.h>
+#include "fastrtps/qos/QosPolicies.h"
```
fujitatomoya, line 24: Just a question and study. according to https://ci.ros2.org/job/ci_linux/15247/testReport/junit/rmw_fastrtps_shared_cpp/cpplint/build_include_order__4____home_jenkins_agent_workspace_ci_linux_ws_src_ros2_rmw_fastrtps_rmw_fastrtps_shared_cpp_include_rmw_fastrtps_shared_cpp_qos_hpp_24_/, we can just move this line above `#include ...hpp`?

clalancette (approved):
Looks good with green CI.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/qos.hpp
jacobperron: Yes, we can move this line above the .hpp files and I think cpplint will be happy, because with angle brackets cpplint assumes this is a C system header. However, I think technically this is classified as "other" header, not a system header ([reference](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)).

So, I think leaving this is here with double-quotes is slightly more correct than moving above.

PR #580, opened 2022-02-09
Add in fake support for the MESSAGE_LOST event.

clalancette:
Fast-DDS (and rmw_fastrtps_cpp) don't really support this right
now, but having it in the list here allows RViz2 to start up.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

@MiguelCompany @EduPonz FYI

PR #581, opened 2022-02-10
Add support for the MESSAGE_LOST event

MiguelCompany:
This replaces #580 and adds proper support for MESSAGE_LOST event.

The interfaces are already present in Fast-DDS, but the listener callback is currently never called.
This means the RMW will behave as if messages are never lost, until we implement on_sample_lost on Fast DDS v2.6.0.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
@@ -112,6 +112,12 @@ public:
     eprosima::fastdds::dds::DataReader *,
     const eprosima::fastrtps::LivelinessChangedStatus &) final;
 
+  RMW_FASTRTPS_SHARED_CPP_PUBLIC
+  void
+  on_sample_lost(
+    eprosima::fastdds::dds::DataReader *,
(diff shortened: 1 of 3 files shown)


clalancette (approved):
Looks good to me with green CI.  Thanks @MiguelCompany !

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16128/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10757/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16459/)

MiguelCompany:
@clalancette [this test](https://ci.ros2.org/job/ci_linux/16128/testReport/junit/rcl/TestEventFixture/test_sub_message_lost_event/) will fail unless we have a correct implementation for `on_sample_lost` on Fast DDS.

The only way to push this foward I can think of is to disable MESSAGE_LOST support on the CI, by putting [this](https://github.com/ros2/rmw_fastrtps/blob/492cb4635f08ffa302a1d1499188c882a27059b0/rmw_fastrtps_shared_cpp/src/rmw_event.cpp#L27) inside a `#if !defined(BUILD_TESTING)` with a `// TODO: remove when on_sample_lost is correctly implemented by Fast DDS`

Let me know what you think.

PR #583, opened 2022-02-14
Complete events support

MiguelCompany:
This replaces #581 and adds support for the three missing events:

* **RMW_EVENT_OFFERED_QOS_INCOMPATIBLE**: The interfaces are present and working on Fast DDS.
* **RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE**: The interfaces are present and working on Fast DDS.
* **RMW_EVENT_MESSAGE_LOST**: The interfaces are present in Fast DDS, but the listener is currently never called.
  This means the RMW will behave as if messages are never lost, until we implement on_sample_lost on Fast DDS v2.6.0.

Connects to ros2/rcl#960, ros2/rclcpp#1887, and ros2/rclpy#895

Closes #356

PR #584, opened 2022-02-16
Add in support for the MESSAGE_LOST event.

clalancette:
Fast-DDS (and rmw_fastrtps_cpp) don't fully support this right
now, but having it in the list here allows RViz2 to start up.

Implementation extracted from
https://github.com/ros2/rmw_fastrtps/pull/583, which is the more
complete solution.

This also requires a small fix in rcl, which is upcoming.

@MiguelCompany FYI.  We really need to fix RViz2, and the only part we need in order to do that is the MESSAGE_LOST fix.  So this just extracts the necessary bits from https://github.com/ros2/rmw_fastrtps/pull/583 so we can get it in faster.  We can rebase 583 on top of this later on.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
@@ -64,6 +64,24 @@ void SubListener::on_liveliness_changed(
   liveliness_changes_.store(true, std::memory_order_relaxed);
 }
 
+void SubListener::on_sample_lost(
+  eprosima::fastdds::dds::DataReader * /* reader */,
+  const eprosima::fastdds::dds::SampleLostStatus & status)
+{
+  std::lock_guard<std::mutex> lock(internalMutex_);
+
+  // the change to sample_lost_status_ needs to be mutually exclusive with
+  // rmw_wait() which checks hasEvent() and decides if wait() needs to be called
+  ConditionalScopedLock clock(conditionMutex_, conditionVariable_);
+
+  // Assign absolute values
+  sample_lost_status_.total_count = status.total_count;
+  // Accumulate deltas
+  sample_lost_status_.total_count_change += status.total_count_change;
+
+  sample_lost_changes_.store(true, std::memory_order_relaxed);
+}
+
 bool SubListener::hasEvent(rmw_event_type_t event_type) const
 {
   assert(rmw_fastrtps_shared_cpp::internal::is_event_supported(event_type));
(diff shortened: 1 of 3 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16173/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10800/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16514/)

hidmic (commented):
@clalancette This LGTM pending green CI, but I wonder, why not simply review and merge #584?

clalancette:
> @clalancette This LGTM pending green CI, but I wonder, why not simply review and merge #584?

I assume you mean #583.  The main reason is that #583 is adding in support for some other events as well, and that is causing lots of spam on the terminal, as well as failed tests.  I want to get just this piece of it in ASAP, because until this is in RViz2 is completely broken on Rolling.

rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
```diff
@@ -104,6 +124,15 @@ bool SubListener::takeNextEvent(rmw_event_type_t event_type, void * event_info)
         deadline_changes_.store(false, std::memory_order_relaxed);
       }
       break;
+    case RMW_EVENT_MESSAGE_LOST:
+      {
+        auto rmw_data = static_cast<rmw_message_lost_status_t *>(event_info);
+        rmw_data->total_count = sample_lost_status_.total_count;
+        rmw_data->total_count_change = sample_lost_status_.total_count_change;
+        sample_lost_status_.total_count_change = 0;
+        sample_lost_changes_.store(false, std::memory_order_relaxed);
```
wjwwood, line 133: So, can `total_count_change` not be used as a proxy for `sample_lost_changes_`? Like if `total_count_change` != 0, then `sample_lost_changes_` will always be `true`, and the opposite as well, if it is == 0, then `sample_lost_changes_` will always be `false`, right?

In which case, it might make sense to wrap the entire thing as `std::atomic<eprosima::fastdds::dds::SampleLostStatus>`? Depending on the size of the fields in that struct, and the processor it may still be truly atomic, otherwise it will wrap it in a mutex.

wjwwood (commented):
One question about the approach, but looks good for the most part.

wjwwood (approved):

fujitatomoya (approved):

MiguelCompany:
@clalancette @fujitatomoya Please see [my comment](https://github.com/ros2/rmw_fastrtps/pull/583#issuecomment-1044348388) on #583

clalancette:
Closing since we merged #583 in instead.

PR #587, opened 2022-03-03
Add sequence numbers to message info structure

ivanpauno:
Implements https://github.com/ros2/rmw/pull/318.

`reception_sequence_number` doesn't seem to be available in fastrtps.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -48,8 +48,8 @@ _assign_message_info(
   message_info->source_timestamp = sinfo->source_timestamp.to_ns();
   message_info->received_timestamp = sinfo->reception_timestamp.to_ns();
   auto fastdds_sn = sinfo->sample_identity.sequence_number();
-  message_info->publication_sequence_number = (static_cast<int64_t>(fastdds_sn.high)) <<
-    32 | static_cast<int64_t>(fastdds_sn.low);
+  message_info->publication_sequence_number = (static_cast<uint64_t>(fastdds_sn.high)) <<
+    32 | static_cast<uint64_t>(fastdds_sn.low);
   message_info->reception_sequence_number = RMW_MESSAGE_INFO_SEQUENCE_NUMBER_UNSUPPORTED;
   rmw_gid_t * sender_gid = &message_info->publisher_gid;
   sender_gid->implementation_identifier = identifier;
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -47,6 +47,10 @@ _assign_message_info(
 {
   message_info->source_timestamp = sinfo->source_timestamp.to_ns();
   message_info->received_timestamp = sinfo->reception_timestamp.to_ns();
+  auto fastdds_sn = sinfo->sample_identity.sequence_number();
+  message_info->publication_sequence_number = (static_cast<int64_t>(fastdds_sn.high)) <<
+    32 | static_cast<int64_t>(fastdds_sn.low);
+  message_info->reception_sequence_number = sinfo->reception_timestamp.to_ns();
   rmw_gid_t * sender_gid = &message_info->publisher_gid;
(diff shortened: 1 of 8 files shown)


rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -47,6 +47,10 @@ _assign_message_info(
 {
   message_info->source_timestamp = sinfo->source_timestamp.to_ns();
   message_info->received_timestamp = sinfo->reception_timestamp.to_ns();
+  auto fastdds_sn = sinfo->sample_identity.sequence_number();
+  message_info->publication_sequence_number = (static_cast<uint64_t>(fastdds_sn.high)) <<
+    32 | static_cast<uint64_t>(fastdds_sn.low);
```
wjwwood, line 52: Did you mean to do this?

```suggestion
  message_info->publication_sequence_number =
    (static_cast<uint64_t>(fastdds_sn.high) << 32) |
    static_cast<uint64_t>(fastdds_sn.low);
```

If not, then maybe there is an extra set of parenthesis?

```suggestion
  message_info->publication_sequence_number = static_cast<uint64_t>(fastdds_sn.high) <<
    32 | static_cast<uint64_t>(fastdds_sn.low);
```

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -47,6 +47,10 @@ _assign_message_info(
...
+  message_info->reception_sequence_number = RMW_MESSAGE_INFO_SEQUENCE_NUMBER_UNSUPPORTED;
```
wjwwood, line 53: Should we just implement this by taking a timestamp here?

wjwwood: Ideally we'd take the timestamp when it is received, not as it is being taken. So maybe leaving it as-is would be best. I'm not sure. Can we ask if it's possible to take a timestamp when that occurs?

I was actually just using this timestamp to debug an issue, because it's useful to know the time between the middleware receiving the message (ready to deliver) and the user code actually taking it, as it tells you something about the latency introduced by waiting or the executor or listener-style callbacks, which ever you are using.

ivanpauno: > Should we just implement this by taking a timestamp here?

The reception sequence number?
If we want to manually implement it, I would use a counter, not a timestamp.
I would expect the reception sequence number to not have gaps, that's why it would be weird to use a timestamp IMO.

> I was actually just using this timestamp to debug an issue, because it's useful to know the time between the middleware receiving the message (ready to deliver) and the user code actually taking it, as it tells you something about the latency introduced by waiting or the executor or listener-style callbacks, which ever you are using.

Maybe tracing is the correct way to measure that.
We probably need just to add a couple more tracepoints.

wjwwood: Doh, I'm just confused. I read reception and just auto-completed in my brain to timestamp, lol.

We already have the reception timestamp. You're right it should just be a counter.

So, original question, should we ask about how to implement that? We'd need to increment and store the counter when receiving the message, not taking it, so it'd have to be in a different part of the code.

ivanpauno: > So, original question, should we ask about how to implement that? We'd need to increment and store the counter when receiving the message, not taking it, so it'd have to be in a different part of the code.

I guess it's maybe possible to do that in a listener, but I guess you could miss a message that way (because many were got together and the queue was overridden), so maybe it's better to leave it unimplemented until there's actual support for it (maybe there's support already and I haven't found it).

@MiguelCompany @richiprosima is the reception sequence number available in FastDDS?

jacobperron (approved):
LGTM pending green CI.

It's easy to come back and add the reception sequence number if it becomes available.

wjwwood (approved):
lgtm

It would be nice to have the reception sequence number, so you could detect when a message arrived but you didn't take it (though you can infer that too from some of the QoS Events). But we could add that later if anyone on the FastDDS team has an idea how to implement it.

PR #591, opened 2022-03-17
Add pub/sub init, publish and take instrumentation using tracetools

christophebedard:
This adds tracing instrumentation using `tracetools` for publisher/subscription initialization at the `rmw` level. By recording the pub/sub GIDs, we can match ROS 2 pubs/subs with the corresponding DDS writers/readers.

This also adds a tracepoint for `rmw_publish` and `rmw_take` so that we can track published/taken messages at the rmw level.

For simplicity, this only adds instrumentation for `rmw_fastrtps_cpp` and not `rmw_fastrtps_dynamic_cpp`, since the former is the default RMW implementation while the latter is a [tier 2 implementation](https://www.ros.org/reps/rep-2000.html#id41).

Equivalent to this PR for `rmw_cyclonedds_cpp`: https://github.com/ros2/rmw_cyclonedds/pull/329. Since the default RMW implementation was changed from `rmw_cyclonedds_cpp` to `rmw_fastrtps_cpp`, it would be great to support `rmw_fastrtps_cpp`.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>

### DIFF
--- a/rmw_fastrtps_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_cpp/CMakeLists.txt
@@ -35,6 +35,7 @@ find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rmw_dds_common REQUIRED)
 find_package(rmw_fastrtps_shared_cpp REQUIRED)
+find_package(tracetools REQUIRED)
 
 find_package(fastrtps_cmake_module REQUIRED)
 find_package(fastcdr REQUIRED CONFIG)
@@ -105,6 +106,7 @@ ament_target_dependencies(rmw_fastrtps_cpp
   "rmw_fastrtps_shared_cpp"
   "rmw"
   "rosidl_runtime_c"
+  "tracetools"
 )
 
 target_link_libraries(rmw_fastrtps_cpp
@@ -130,6 +132,7 @@ ament_export_dependencies(rmw_fastrtps_shared_cpp)
 ament_export_dependencies(rosidl_runtime_c)
 ament_export_dependencies(rosidl_typesupport_fastrtps_c)
 ament_export_dependencies(rosidl_typesupport_fastrtps_cpp)
+ament_export_dependencies(tracetools)
 
 register_rmw_implementation(
   "c:rosidl_typesupport_c:rosidl_typesupport_fastrtps_c:rosidl_typesupport_introspection_c"
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -29,6 +29,7 @@
   <build_depend>rosidl_runtime_cpp</build_depend>
   <build_depend>rosidl_typesupport_fastrtps_c</build_depend>
   <build_depend>rosidl_typesupport_fastrtps_cpp</build_depend>
+  <build_depend>tracetools</build_depend>
 
   <build_export_depend>fastcdr</build_export_depend>
   <build_export_depend>fastrtps</build_export_depend>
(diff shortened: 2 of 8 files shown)


christophebedard:
CI testing `rmw_fastrtps_cpp rmw_fastrtps_shared_cpp rmw_fastrtps_dynamic_cpp rmw_implementation test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10953/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16705/)

fujitatomoya (approved):

WideAwakeTN:
Is the TRACEPOINT macro dangerous for hard realtime (i.e. using a mutex or the system heap)? The use of TRACEPOINT in rmw_publish.cpp and rmw_take.cpp could then be problematic.

christophebedard:
> Is the TRACEPOINT macro dangerous for hard realtime (i.e. using a mutex or the system heap)? The use of TRACEPOINT in rmw_publish.cpp and rmw_take.cpp could then be problematic.

I think this is a pretty simplistic question; there is a bit more to it:

There are a bunch of `TRACEPOINT`s here and there in ROS 2, not just here. This macro itself is just a function call, and by default, unless you install the LTTng tracer and re-build ROS 2, the function will be empty.

I do think it would be worth including the LTTng tracepoints by default (see https://github.com/ros2/ros2/issues/1177). In this case or if you install LTTng and re-build ROS 2, if you do not setup LTTng + enable that tracepoint event, it only does a simple check that has virtually no impact at all. If you do enable the event, it can be compatible with real-time if you set up LTTng correctly, see: https://gitlab.com/ros-tracing/ros2_tracing#real-time.

If you really have a real-time application (_hard_ real-time or not), chances are you don't use ROS 2 as-is; you probably have a slightly-modified version that you build yourself, so you can just enable the right build option and remove all `TRACEPOINT()` calls if you want: https://gitlab.com/ros-tracing/ros2_tracing#disabling-tracing

fujitatomoya:
@WideAwakeTN 

> Is the TRACEPOINT macro dangerous for hard realtime

what is your definition about `hard realtime`?

> (i.e. using a mutex or the system heap)?

btw, these are used everywhere in ROS 2 system.

i think hardware realtime system is, (probably my definition, maybe not correct...)
- it must absolutely hit every deadline.
- it must be able to reproduce the same problem(deadline miss).

if this is what your are expecting for ROS 2 system as-is, probably ROS 2 cannot cover your requirements?

WideAwakeTN:
> There are a bunch of TRACEPOINTs here and there in ROS 2, not just here. This macro itself is just a function call, and by default, unless you install the LTTng tracer and re-build ROS 2, the function will be empty.

That sound's fair enough. 

> what is your definition about hard realtime

I think it's similar to yours. I experienced in the DDS context that people often use the term realtime in a soft or loose manner, i.e. expressing that code is "fast". I use the term hard realtime to distinguish between "fast" and code that is not interrupted unexpectedly and has timiing characteristics is similar to a microcontroller (deadlines, repeatability).

> probably ROS 2 cannot cover your requirements

My impression was the goal of ROS 2 is to be hard realtime. Having a single system heap access definitely destroys this goal, which is what I have witnessed personally in a project in which I was involved. Heap access in a realtime critical thread creates an interaction with all other heap accessing threads in a system, which leads to total unpredictability and hard to dignose errors at the customer. In general, lock free code is not so easy to achieve.

Is there some documentation available on what is holding Ros2 back from achieving hard realtime at the moment? Depending on how my job life evolves I may be getting more involved with Ros2 development in the future. I also don't want to reinvent the wheel, I also saw that happening over and over.

Thanks for your feedback.

fujitatomoya:
> Is there some documentation available on what is holding Ros2 back from achieving hard realtime at the moment? Depending on how my job life evolves I may be getting more involved with Ros2 development in the future. I also don't want to reinvent the wheel, I saw that happening over and over too.

one thing that i can introduce is we have https://github.com/ros-realtime/community. that could be the place we can discuss details and plans, i think.

fujitatomoya:
@christophebedard can you fix the conflict? this needs to be merged in today.

christophebedard:
@fujitatomoya I rebased to fix the conflicts. New CI run just to be safe:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10992/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16755/)

fujitatomoya:
@christophebedard right, this warning is unrelated to this PR.

PR #600, opened 2022-05-03
Add EventsExecutor API for dynamic_fastrtps

j-rivero:
Fixes https://github.com/ros2/rmw_fastrtps/issues/599

Two commits:
  * Copy the implementation done in #468 for dynamic version
  * Add some check to parameters passed to fastrtps_cpp to keep both implementations in sync

Testing:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11226/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17042/)

I think that cpplint is not related to this PR. I can open a new PR to fix it.

PR #601, opened 2022-05-04
Fix cpplint error on master

j-rivero:
Minor problem detected while processing #600


clalancette (approved):
This looks good to me.  Since this is only a whitespace fix, I think we can merge it once the Rpr job comes back clean.

PR #602, opened 2022-05-05
Allow null arguments in the EventsExecutor parameters

j-rivero:
After merging #600, failing tests have appeared in rclcpp that points to problems with assertions introduced on parameter there.

I did that following the [API](https://docs.ros.org/en/ros2_packages/rolling/api/rmw/generated/function_rmw_8h_1a2195fea162b6f4b30f2d7bdd00a56401.html ) that explicit mention when a parameter can be null (i.e. `user_data`). Seems to me like the null check and error handling is being done at a different place of the stack.

Testing:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16652/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17049/)

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_client.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_client.cpp
@@ -537,9 +537,6 @@ rmw_client_set_on_new_response_callback(
   rmw_event_callback_t callback,
   const void * user_data)
 {
-  RMW_CHECK_ARGUMENT_FOR_NULL(rmw_client, RMW_RET_INVALID_ARGUMENT);
-  RMW_CHECK_ARGUMENT_FOR_NULL(callback, RMW_RET_INVALID_ARGUMENT);
-
   return rmw_fastrtps_shared_cpp::__rmw_client_set_on_new_response_callback(
     rmw_client,
     callback,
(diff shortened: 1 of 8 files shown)


clalancette (approved):

j-rivero:
While reviewing the error logs in the past builds I noticed that probably the problem was only in checking callbacks that can be `nullptr`, the value is handled in different software layer (i. e. https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp#L203).

So I think that we can leave the checks on the other parameters:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16653/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11241/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17050/)

clalancette (approved):
Still looks OK, assuming CI comes back green.

Blast545:
This PR and https://github.com/ros2/rmw_fastrtps/pull/600 should be backported to Humble to address the build regressions on the `fastrtps-dynamic` Hci jobs.

PR #642, opened 2022-10-18
Fix for solving error when building

TheConstructAi:
```
/home/user/ros2_ws/src/rmw_fastrtps/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp:206:52: error: no matching function for call to 'eprosima::fastdds::dds::DataReader::get_unread_count(bool)'
  206 |     return info_->request_reader_->get_unread_count(true);
```
The only real change is to set 

`get_unread_count(true)`

To

`get_unread_count()`

Otherwise, colcon build generated the above mentioned error.

The other changes are linter changes.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
@@ -18,8 +18,8 @@
 #include <memory>
 #include <mutex>
 #include <set>
-#include <utility>
 #include <string>
+#include <utility>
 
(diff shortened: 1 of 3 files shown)


MiguelCompany:
@TheConstructAi Building from sources using branch 2.6.x  of Fast DDS (which is what [these lines](https://github.com/ros2/ros2/blob/00f3ba9f73916a5eab322710edaaf197f0f10e31/ros2.repos#L34-L37) suggest should work.

clalancette:
Yes, this should not be necessary on Humble; we are already shipping version 2.6.x of Fast-DDS there.  I'm going to close this out, but feel free to reopen if you think this is in error.

PR #645, opened 2022-11-28
Implement matched event

Barry-Xu-2018:
Implement the feature about ros2/rmw#330

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp
@@ -191,9 +191,6 @@ private:
   eprosima::fastdds::dds::PublicationMatchedStatus matched_status_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
-  size_t matched_unread_
-  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
-
   void trigger_event(rmw_event_type_t event_type);
 };
 
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp
@@ -152,6 +152,13 @@ public:
     eprosima::fastdds::dds::QosPolicyId_t last_policy_id,
     uint32_t total_count, uint32_t total_count_change);
 
+  RMW_FASTRTPS_SHARED_CPP_PUBLIC
+  void update_matched(
+    int32_t total_count,
+    int32_t total_count_change,
+    int32_t current_count,
+    int32_t current_count_change);
+
 private:
   CustomPublisherInfo * publisher_info_ = nullptr;
 
@@ -176,6 +183,15 @@ private:
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
   eprosima::fastdds::dds::OfferedIncompatibleQosStatus incompatible_qos_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  bool matched_changes_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  eprosima::fastdds::dds::PublicationMatchedStatus matched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  size_t matched_unread_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
   void trigger_event(rmw_event_type_t event_type);
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
@@ -196,6 +196,13 @@ public:
     eprosima::fastdds::dds::QosPolicyId_t last_policy_id,
     uint32_t total_count, uint32_t total_count_change);
 
+  RMW_FASTRTPS_SHARED_CPP_PUBLIC
+  void update_matched(
+    int32_t total_count,
+    int32_t total_count_change,
+    int32_t current_count,
+    int32_t current_count_change);
+
 private:
   CustomSubscriberInfo * subscriber_info_ = nullptr;
 
@@ -223,6 +230,15 @@ private:
   eprosima::fastdds::dds::RequestedIncompatibleQosStatus incompatible_qos_status_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
+  bool matched_changes_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  eprosima::fastdds::dds::SubscriptionMatchedStatus matched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  size_t matched_unread_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
   std::set<eprosima::fastrtps::rtps::GUID_t> publishers_ RCPPUTILS_TSA_GUARDED_BY(
     publishers_mutex_);
 
@@ -233,6 +249,8 @@ private:
   std::mutex on_new_message_m_;
 
   mutable std::mutex publishers_mutex_;
+
+  void trigger_event(rmw_event_type_t event_type);
 };
 
 #endif  // RMW_FASTRTPS_SHARED_CPP__CUSTOM_SUBSCRIBER_INFO_HPP_
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
@@ -236,9 +236,6 @@ private:
   eprosima::fastdds::dds::SubscriptionMatchedStatus matched_status_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
-  size_t matched_unread_
-  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
-
   std::set<eprosima::fastrtps::rtps::GUID_t> publishers_ RCPPUTILS_TSA_GUARDED_BY(
     publishers_mutex_);
 
--- a/rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
@@ -91,8 +91,7 @@ RMWPublisherEvent::RMWPublisherEvent(CustomPublisherInfo * info)
   deadline_changed_(false),
   liveliness_changed_(false),
   incompatible_qos_changed_(false),
-  matched_changes_(false),
-  matched_unread_(0)
+  matched_changes_(false)
 {
 }
 
@@ -248,10 +247,9 @@ void RMWPublisherEvent::set_on_new_event_callback(
         break;
       case RMW_EVENT_PUBLICATION_MATCHED:
         {
-          publisher_info_->data_writer_->get_publication_matched_status(matched_status_);
-          if (matched_unread_ > 0) {
-            callback(user_data, matched_unread_);
-            matched_unread_ = 0;
+          if (matched_status_.total_count_change > 0) {
+            callback(user_data, matched_status_.total_count_change);
+            publisher_info_->data_writer_->get_publication_matched_status(matched_status_);
             matched_status_.total_count_change = 0;
             matched_status_.current_count_change = 0;
           }
@@ -374,16 +372,6 @@ void RMWPublisherEvent::trigger_event(rmw_event_type_t event_type)
 {
   if (on_new_event_cb_[event_type]) {
     on_new_event_cb_[event_type](user_data_[event_type], 1);
-  } else {
-    switch (event_type) {
-      case RMW_EVENT_PUBLICATION_MATCHED:
-        {
-          matched_unread_++;
-          break;
-        }
-      default:
-        break;
-    }
   }
 
   event_guard[event_type].set_trigger_value(true);
--- a/rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
@@ -44,7 +44,15 @@ void CustomDataWriterListener::on_publication_matched(
   } else if (status.current_count_change == -1) {
     publisher_event_->untrack_unique_subscription(
       eprosima::fastrtps::rtps::iHandle2GUID(status.last_subscription_handle));
+  } else {
+    return;
   }
+
+  publisher_event_->update_matched(
+    status.total_count,
+    status.total_count_change,
+    status.current_count,
+    status.current_count_change);
 }
 
 
@@ -82,7 +90,9 @@ RMWPublisherEvent::RMWPublisherEvent(CustomPublisherInfo * info)
 : publisher_info_(info),
   deadline_changed_(false),
   liveliness_changed_(false),
-  incompatible_qos_changed_(false)
+  incompatible_qos_changed_(false),
+  matched_changes_(false),
+  matched_unread_(0)
 {
 }
 
@@ -158,6 +168,30 @@ bool RMWPublisherEvent::take_event(
         inconsistent_topic_status_.total_count_change = 0;
       }
       break;
+    case RMW_EVENT_PUBLICATION_MATCHED:
+      {
+        auto rmw_data = static_cast<rmw_matched_status_t *>(event_info);
+
+        if (matched_changes_) {
+          rmw_data->total_count = static_cast<size_t>(matched_status_.total_count);
+          rmw_data->total_count_change = static_cast<size_t>(matched_status_.total_count_change);
+          rmw_data->current_count = static_cast<size_t>(matched_status_.current_count);
+          rmw_data->current_count_change = matched_status_.current_count_change;
+          matched_changes_ = false;
+        } else {
+          eprosima::fastdds::dds::PublicationMatchedStatus matched_status;
+          publisher_info_->data_writer_->get_publication_matched_status(matched_status);
+
+          rmw_data->total_count = static_cast<size_t>(matched_status.total_count);
+          rmw_data->total_count_change = static_cast<size_t>(matched_status.total_count_change);
+          rmw_data->current_count = static_cast<size_t>(matched_status.current_count);
+          rmw_data->current_count_change = matched_status.current_count_change;
+        }
+
+        matched_status_.total_count_change = 0;
+        matched_status_.current_count_change = 0;
+      }
+      break;
     default:
       return false;
   }
@@ -210,6 +244,17 @@ void RMWPublisherEvent::set_on_new_event_callback(
         if (inconsistent_topic_status_.total_count_change > 0) {
           callback(user_data, inconsistent_topic_status_.total_count_change);
           inconsistent_topic_status_.total_count_change = 0;
+        }
+        break;
+      case RMW_EVENT_PUBLICATION_MATCHED:
+        {
+          publisher_info_->data_writer_->get_publication_matched_status(matched_status_);
+          if (matched_unread_ > 0) {
+            callback(user_data, matched_unread_);
+            matched_unread_ = 0;
+            matched_status_.total_count_change = 0;
+            matched_status_.current_count_change = 0;
+          }
         }
         break;
       default:
@@ -307,10 +352,38 @@ void RMWPublisherEvent::update_inconsistent_topic(uint32_t total_count, uint32_t
   trigger_event(RMW_EVENT_PUBLISHER_INCOMPATIBLE_TYPE);
 }
 
+void RMWPublisherEvent::update_matched(
+  int32_t total_count,
+  int32_t total_count_change,
+  int32_t current_count,
+  int32_t current_count_change)
+{
+  std::lock_guard<std::mutex> lock(on_new_event_m_);
+
+  matched_status_.total_count = total_count;
+  matched_status_.total_count_change += total_count_change;
+  matched_status_.current_count = current_count;
+  matched_status_.current_count_change += current_count_change;
+
+  matched_changes_ = true;
+
+  trigger_event(RMW_EVENT_PUBLICATION_MATCHED);
+}
+
 void RMWPublisherEvent::trigger_event(rmw_event_type_t event_type)
 {
   if (on_new_event_cb_[event_type]) {
     on_new_event_cb_[event_type](user_data_[event_type], 1);
+  } else {
+    switch (event_type) {
+      case RMW_EVENT_PUBLICATION_MATCHED:
+        {
+          matched_unread_++;
+          break;
+        }
+      default:
+        break;
+    }
   }
 
   event_guard[event_type].set_trigger_value(true);
--- a/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
@@ -111,8 +111,7 @@ RMWSubscriptionEvent::RMWSubscriptionEvent(CustomSubscriberInfo * info)
   liveliness_changed_(false),
   sample_lost_changed_(false),
   incompatible_qos_changed_(false),
-  matched_changes_(false),
-  matched_unread_(0)
+  matched_changes_(false)
 {
 }
 
@@ -305,10 +304,9 @@ void RMWSubscriptionEvent::set_on_new_event_callback(
         break;
       case RMW_EVENT_SUBSCRIPTION_MATCHED:
         {
-          subscriber_info_->data_reader_->get_subscription_matched_status(matched_status_);
-          if (matched_unread_ > 0) {
-            callback(user_data, matched_unread_);
-            matched_unread_ = 0;
+          if (matched_status_.total_count_change > 0) {
+            callback(user_data, matched_status_.total_count_change);
+            subscriber_info_->data_reader_->get_subscription_matched_status(matched_status_);
             matched_status_.total_count_change = 0;
             matched_status_.current_count_change = 0;
           }
@@ -500,16 +498,6 @@ void RMWSubscriptionEvent::trigger_event(rmw_event_type_t event_type)
 {
   if (on_new_event_cb_[event_type]) {
     on_new_event_cb_[event_type](user_data_[event_type], 1);
-  } else {
-    switch (event_type) {
-      case RMW_EVENT_SUBSCRIPTION_MATCHED:
-        {
-          matched_unread_++;
-          break;
-        }
-      default:
-        break;
-    }
   }
 
   event_guard[event_type].set_trigger_value(true);
--- a/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
+++ b/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
@@ -44,7 +44,15 @@ CustomDataReaderListener::on_subscription_matched(
   } else if (info.current_count_change == -1) {
     subscription_event_->untrack_unique_publisher(
       eprosima::fastrtps::rtps::iHandle2GUID(info.last_publication_handle));
+  } else {
+    return;
   }
+
+  subscription_event_->update_matched(
+    info.total_count,
+    info.total_count_change,
+    info.current_count,
+    info.current_count_change);
 }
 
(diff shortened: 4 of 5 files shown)


Barry-Xu-2018:
@gbiggs @sloretz   
While I change code, I have a question about below codes   

https://github.com/ros2/rmw_fastrtps/blob/ccea7a356900a16a2872670b67952c99a72ab65d/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L191-L205  

Could you explain what cases need double-check for `active` ?  
I want to modify it (Please refer to my change in rmw_wait.cpp) since 2 RMW events is corresponding to one DDS event for implementation.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -96,6 +96,46 @@ bool PubListener::take_event(
...
+          matched_changes_ = false;
+        } else {
+          eprosima::fastdds::dds::PublicationMatchedStatus matched_status;
+          publisher_info_->data_writer_->get_publication_matched_status(matched_status);
+          rmw_data->total_count_change = matched_status.total_count_change;
+          rmw_data->current_count = matched_status.current_count;
+          rmw_data->current_count_change = matched_status.current_count_change;
+        }
+        matched_status_.current_count_change = 0;
+        matched_status_.total_count_change = 0;
```
ivanpauno, line 116: why is the if-else needed?
`get_publication_matched_status()` and the `on_publication_matched()` listener should reset the internal `current_count_change`/`total_count_change` counters.

As we are using the listener, we should use:

```suggestion
          std::lock_guard guard{on_new_event_m_};
          rmw_data->total_count_change = matched_status_.total_count_change;
          rmw_data->current_count = matched_status_.current_count;
          rmw_data->current_count_change = matched_status_.current_count_change;
          matched_status_.current_count = 0;
          matched_status_.current_count_change = 0;
```

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -96,6 +96,46 @@ bool PubListener::take_event(
...
+          unmatched_changes_ = false;
+        } else {
+          eprosima::fastdds::dds::PublicationMatchedStatus matched_status;
+          publisher_info_->data_writer_->get_publication_matched_status(matched_status);
+          rmw_data->total_count_change = matched_status.total_count_change;
+          rmw_data->current_count = matched_status.current_count;
+          rmw_data->current_count_change = matched_status.current_count_change;
+        }
+        unmatched_status_.current_count_change = 0;
+        unmatched_status_.total_count_change = 0;
```
ivanpauno, line 136: same

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -138,6 +178,32 @@ void PubListener::set_on_new_event_callback(
           incompatible_qos_status_.total_count_change = 0;
         }
         break;
+      case RMW_EVENT_PUBLICATION_MATCHED:
+        {
+          publisher_info_->data_writer_->get_publication_matched_status(
+            matched_status_);
```
ivanpauno, line 184: it shouldn't be triggered the event callback for previous events.
It makes sense for messages, but IMO not for new matches.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +226,41 @@ void PubListener::set_on_new_event_callback(
...
+void
+PubListener::on_publication_matched(
+  eprosima::fastdds::dds::DataWriter * /* writer */,
+  const eprosima::fastdds::dds::PublicationMatchedStatus & info)
+{
+  rmw_event_type_t event_type = RMW_EVENT_INVALID;
+  {
+    std::lock_guard<std::mutex> lock(discovery_m_);
+    if (info.current_count_change == 1) {
+      subscriptions_.insert(eprosima::fastrtps::rtps::iHandle2GUID(info.last_subscription_handle));
```
ivanpauno, line 238: I know this is old, but what's being `subscriptions_` used for?
It seems that we're filling it but never using it.
In that case, I would suggest removing the container.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -204,12 +196,26 @@ class SubListener : public EventListenerInterface, public eprosima::fastdds::dds
   bool incompatible_qos_changes_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
+  eprosima::fastdds::dds::SubscriptionMatchedStatus matched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  bool matched_changes_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  eprosima::fastdds::dds::SubscriptionMatchedStatus unmatched_status_
```
ivanpauno, line 205: we don't need `matched_status_` and `unmatched_status_`, we should only have one

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +226,41 @@ void PubListener::set_on_new_event_callback(
...
+    matched_status_.total_count_change += info.total_count_change;
+    matched_status_.current_count = info.current_count;
+    matched_status_.current_count_change += info.current_count_change;
+
+    unmatched_status_.total_count_change += info.total_count_change;
+    unmatched_status_.current_count = info.current_count;
+    unmatched_status_.current_count_change += info.current_count_change;
+
+    matched_changes_ = true;
+    unmatched_changes_ = true;
```
ivanpauno, line 258:
```suggestion
```

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +226,41 @@ void PubListener::set_on_new_event_callback(
...
+
+    trigger_event(event_type);
```
ivanpauno, line 260: the event callback should be called here

ivanpauno (changes requested):
first review

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: > why is the if-else needed?

I refer to the process of other event.  I don't sure which case may have no listener (Have to call get_publication_matched_status() to get status).   
If listener is always used, we need not get status by get_publication_matched_status(). I will update it.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: Will update

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: So you mean we should directly use matched_status_.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: `subscriptions_` is used for public API

https://github.com/ros2/rmw_fastrtps/blob/rolling/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp#L102-L106

And this API is called by rmw_publisher_count_matched_subscriptions()

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
Barry-Xu-2018: We have 2 events.  
While user takes status for matched event, we can clear matched status (But we should not clear unmatched status). 

```c++
typedef struct RMW_PUBLIC_TYPE rmw_matched_status_s
{
  ...

  /**
   * The change in current_count since the last time the status was read.
   */
  int32_t current_count_change;
} rmw_matched_status_t;
```

current_count _change means `The change in current_count since the last time the status was read`.  So I think the change should be from the last matched event status read. This read should not affect unmatched status.  

So we use 2 status separately for matched event and unmatched event.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
ivanpauno: I mean that the callback shouldn't be called when it's added.
i.e. do nothing in this listener.

Though we're already doing something for all the other events, which I'm not sure if it's a good idea or not, but we're already doing.
So in that case, yes using `matched_status_` would also be okay.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
ivanpauno: fair enough.
It would be better to keep a count in the future though, which we're already doing in `matched_status_`.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
ivanpauno: i.e. an event that's triggered when the "current_count_change" increased or decreased (or even when it's zero).

Barry-Xu-2018: If we only use matched event, user cannot be notified while unmatched event.  

User can set callback via rmw_event_set_callback().  
The callback is defined as the below 
```c
/// Common event callback type signature.
/**
 * Event callbacks of this type can be called in various scenarios, e.g.
 * data becomes available on a subscription, a QoS event has occurred, or
 * something similar.
 *
 * The user_data argument is given by the user when registering the callback,
 * and is given back to the callback each time so it can have associated,
 * user-defined state.
 *
 * The number_of_events argument indicates the number of events since the
 * callback was called.
 * This is most often 1, but can be > 1 when events occur before the callback
 * is registered.
 * It should never be 0.
 *
 * \sa rmw_subscription_set_on_new_message_callback()
 * \sa rmw_service_set_on_new_request_callback()
 * \sa rmw_client_set_on_new_response_callback()
 * \sa rmw_event_set_callback()
 */
typedef void (* rmw_event_callback_t)(const void * user_data, size_t number_of_events);
```
number_of_events should be bigger than 0.  We cannot use negative number for unmatched. Do you want to change this interface to allow negative number ?

ivanpauno: That seems to be a problem in the events executor, I'm not sure why the data that it's passed to the callback in the case of an event is only a number, instead of the full event status struct.
I wouldn't try to modify how the event should be defined to try to fit that.

MiguelCompany: I agree with @ivanpauno. May we always pass `1` as the `number_of_events`?

It will mean something like. The matching status has changed `1` times.

Barry-Xu-2018: According to current implementation, there are 2 ways

1. Register callback to DDS
`rmw_event_callback_t` is only used for rmw_event_callback_t() which register callback directly to DDS.   
That is, while the event occurs, the callback function is called at DDS layer.   

The event number always set to 1 while calling callback.
e.g.  
https://github.com/ros2/rmw_fastrtps/blob/ccea7a356900a16a2872670b67952c99a72ab65d/rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp#L213-L220

2. Wait event and take status
Full event status can be gotten via rmw_take_event(). It is used to wait event via wait_set(). And Once event occur, get event status via rmw_take_event() and call user's callback function with taken data.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: Yes.  
The registered callback function (by rmw_event_set_callback()) will be called. 

```c++
void PubListener::trigger_event(rmw_event_type_t event_type)
{
  if (on_new_event_cb_[event_type]) {
    on_new_event_cb_[event_type](user_data_[event_type], 1);
  }

  event_guard[event_type].set_trigger_value(true);
}
```

Barry-Xu-2018:
Updated code based on review comments.  
Now remain one comments https://github.com/ros2/rmw_fastrtps/pull/645#discussion_r1034772808

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
Barry-Xu-2018: @ivanpauno Friendly ping.  
Whether do we consider only one matched event ?   
If we only use one matched event, while `rmw_event_callback_t` is called, user cannot directly distinguish matched/unmatched event. User have to use rmw_take_event() to further check.

ivanpauno: > If we only use one matched event, while rmw_event_callback_t is called, user cannot directly distinguish matched/unmatched event. User have to use rmw_take_event() to further check.

Yes, that will be a problem with the event callback API.

Barry-Xu-2018: > Yes, that will be a problem with the event callback API.

So we keep current design which use matched/unmatched 2 events.    
Besides, I consider changing rmw_matched_status_t to make more clear definition for user.  
https://github.com/ros2/rmw/pull/331#issuecomment-1321135801  
What do you think ?

ivanpauno: > So we keep current design which use matched/unmatched 2 events.

As commented above, I would prefer to not have both.
It seems that we're forcing them to be different events, but in reality they're the same thing.
As @MiguelCompany commented, you can use 1 in number of events.
The user would have to get the event status to know the actual info when using callbacks.

ivanpauno: @wjwwood do you have any opinion on this?

Barry-Xu-2018: > As commented above, I would prefer to not have both.
> It seems that we're forcing them to be different events, but in reality they're the same thing.
> As @MiguelCompany commented, you can use 1 in number of events.
> The user would have to get the event status to know the actual info when using callbacks.

I got your thourght.   
If the user wants to get unmatched event, the case of only matched event will make the user inconvenient to use. He have to do further check if current event is match or unmatch.  Users who do not know the DDS specification will wonder why it was designed this way.

Barry-Xu-2018:
do rebase.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +197,40 @@ void PubListener::set_on_new_event_callback(
...
+    }
+  }
+
+  if (event_type != RMW_EVENT_INVALID) {
+    {
+      std::lock_guard<std::mutex> lock_mutex(on_new_event_m_);
+
+      matched_status_.total_count_change += info.total_count_change;
+      matched_status_.current_count = info.current_count;
+      matched_status_.current_count_change += info.current_count_change;
```
fujitatomoya, line 223: these are stored only if `event_type` is equal to `RMW_EVENT_PUBLICATION_MATCHED`?

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +197,40 @@ void PubListener::set_on_new_event_callback(
...
+
+      unmatched_status_.total_count_change += info.total_count_change;
+      unmatched_status_.current_count = info.current_count;
+      unmatched_status_.current_count_change += info.current_count_change;
```
fujitatomoya, line 227: these are stored only if event_type is equal to RMW_EVENT_PUBLICATION_UNMATCHED?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -90,24 +90,16 @@ class SubListener : public EventListenerInterface, public eprosima::fastdds::dds
     , liveliness_changes_(false)
     , sample_lost_changes_(false)
     , incompatible_qos_changes_(false)
+    , matched_changes_(false)
```
fujitatomoya, line 93: Do we need to keep this?

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -138,6 +161,20 @@ void PubListener::set_on_new_event_callback(
           incompatible_qos_status_.total_count_change = 0;
         }
         break;
+      case RMW_EVENT_PUBLICATION_MATCHED:
+        {
+          // Not call callback since recorded matched event happened before
```
fujitatomoya, line 166: I think this is correct and expected behavior that application does not want to get the matched or unmatched event happened before.

rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
```diff
@@ -317,10 +340,47 @@ void SubListener::on_requested_incompatible_qos(
...
   }
 
-  event_guard[RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE].set_trigger_value(true);
+  if (event_type != RMW_EVENT_INVALID) {
+    {
+      std::lock_guard<std::mutex> lock_mutex(on_new_event_m_);
+
+      matched_status_.total_count_change += info.total_count_change;
+      matched_status_.current_count = info.current_count;
+      matched_status_.current_count_change += info.current_count_change;
```
fujitatomoya, line 377: these are stored only if `event_type` is RMW_EVENT_SUBSCRIPTION_MATCHED?

rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp
```diff
@@ -317,10 +340,47 @@ void SubListener::on_requested_incompatible_qos(
...
+
+      unmatched_status_.total_count_change += info.total_count_change;
+      unmatched_status_.current_count = info.current_count;
+      unmatched_status_.current_count_change += info.current_count_change;
```
fujitatomoya, line 381: these are stored only if `event_type` is RMW_EVENT_SUBSCRIPTION_UNMATCHED?

fujitatomoya (changes requested):
some minor comments need to be resolved.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
Barry-Xu-2018: Not used. 
This is residual code.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -204,12 +196,26 @@ class SubListener : public EventListenerInterface, public eprosima::fastdds::dds
   bool incompatible_qos_changes_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
+  eprosima::fastdds::dds::SubscriptionMatchedStatus matched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  bool matched_changes_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
```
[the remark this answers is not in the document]
Barry-Xu-2018: Not used.   
Remove it.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -204,12 +196,26 @@ class SubListener : public EventListenerInterface, public eprosima::fastdds::dds
...
+
+  eprosima::fastdds::dds::SubscriptionMatchedStatus unmatched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  bool unmatched_changes_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
```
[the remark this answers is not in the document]
Barry-Xu-2018: Not used.   
Remove it.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
Barry-Xu-2018: Yes. `matched_status_` is used for RMW_EVENT_PUBLICATION_MATCHED. 
And the status saved in `matched_status_` is taken by rmw_take_event. 
So while event happy, related status should be recorded in `matched_status_`.  

After rmw_take_event with `RMW_EVENT_PUBLICATION_MATCHED` is called, matched_status_ will be cleared.  
This is why we use `matched_status_` and `unmatched_status_` to separately save status.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
iuhilnehc-ynos: Sorry, I don't think so.

I think we might lose a event that users care about.
The case as below,

```
a subscriber(A) is running,
then run a publisher:
   the DDS publisher created.      // the publisher might get the subscriber matched status immediately which happens in the DDS
   set_event_callback (callback)   // callback can't be triggered for the connected subscriber(A).
                                   // if there is no other subscribers comming any more.
   executor.spin()
```

Barry-Xu-2018: Yes. For this case, event may be lost.   

Now except matched/unmatched event, other events all deal with previous event while registering callback. Not sure whether original design may have taken this situation into consideration.   
@fujitatomoya What do you think ?

fujitatomoya: So after all, event that happened before also notifies the user application.

i was not sure about this but thinking about the use case for the application, probably this is useful especially https://github.com/ros2/rmw/issues/330#issue-1371797574 it makes sense to receive the matched/unmatched event that happened before.

fujitatomoya (approved):
lgtm

fujitatomoya:
@iuhilnehc-ynos can you have another look on this?

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -201,6 +201,17 @@ __rmw_wait(
         if (guard_condition.get_trigger_value()) {
           active = true;
           guard_condition.set_trigger_value(false);
+        } else {
+          switch (event->event_type) {
+            case RMW_EVENT_SUBSCRIPTION_MATCHED:
```
iuhilnehc-ynos, line 206: It's better to add a comment to explain why we add these codes.

iuhilnehc-ynos (commented):
I don't know if it's the correct way to use them, but it looks good to me overall.

I just wondered something as below,
What if there is a requirement about adding `matched status` event based on the DDS feature in the future? Do we need to update the whole related PRs?

I mean if we don't add the only `matched status` event, we should not use the name `matched status` in  `rmw`.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
Barry-Xu-2018: Yes. I should add explanation for the change.

Barry-Xu-2018:
> I just wondered something as below, What if there is a requirement about adding `matched status` event based on the DDS feature in the future? Do we need to update the whole related PRs?
>
> I mean if we didn't add the `matched status` event, we should not use the name `matched status` in `rmw`.

Current rmw `matched status` is redefined based on DDS matched status definition for the convenience of user understanding.     
If introduce original DDS matched status, maybe we can use dds_matched_status.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp
```diff
@@ -146,6 +138,12 @@ class PubListener : public EventListenerInterface, public eprosima::fastdds::dds
   bool incompatible_qos_changes_
   RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
 
+  eprosima::fastdds::dds::PublicationMatchedStatus matched_status_
+  RCPPUTILS_TSA_GUARDED_BY(on_new_event_m_);
+
+  eprosima::fastdds::dds::PublicationMatchedStatus unmatched_status_
```
iuhilnehc-ynos, line 144: It's weird for me to define a `unmatched_status` with a structure named `PublicationMatchedStatus`.
Can we use the structures defined in the `rmw` directly?

Currently, there are two separate calculations in both `on_*_matched` and `take_event` for `unmatched_status`.
Can we just update the final data in `on_*_matched`?

Barry-Xu-2018: > Can we use the structures defined in the rmw directly?

Yes. Need not use eprosima::fastdds::dds::PublicationMatchedStatus.

> Currently, there are two separate calculations in both on_*_matched and take_event for unmatched_status.
> Can we just update the final data in on_*_matched?

Yes. I will optimize code.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -160,6 +198,41 @@ void PubListener::set_on_new_event_callback(
...
 
+void
+PubListener::on_publication_matched(
+  eprosima::fastdds::dds::DataWriter * /* writer */,
+  const eprosima::fastdds::dds::PublicationMatchedStatus & info)
+{
+  rmw_event_type_t event_type = RMW_EVENT_INVALID;
+  {
+    std::lock_guard<std::mutex> lock(discovery_m_);
+    if (info.current_count_change == 1) {
```
fujitatomoya, line 209: related to https://github.com/ros2/rmw_connextdds/pull/101#discussion_r1129753385, same question.

`current_count_change` will be either 1 or -1 always? otherwise, we will miss the event. as far as i see the specification, this could be 2 or -2 for example.

Barry-Xu-2018: These codes aren't changed. I only move it.  
The original code is located at   
https://github.com/ros2/rmw_fastrtps/blob/rolling/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp#L70-L84

Barry-Xu-2018: I think on_publication_matched() was called while each DDS publication matched event happened. So the value of it will not exceed 1.

iuhilnehc-ynos: Yes, it should be 1 or -1 always. Because the [listener](https://github.com/ros2/rmw_fastrtps/blob/8510fc1582383ad66092cf8b2236713df0e236e6/rmw_fastrtps_cpp/src/publisher.cpp#L267-L271) for `publication_matched`  is registered at the beginning.

https://github.com/eProsima/Fast-DDS/blob/bf181c8146f75741417590d6170e3395c2c86355/src/cpp/fastdds/publisher/DataWriterImpl.cpp#L1343-L1345 will be called every time from ([1](https://github.com/eProsima/Fast-DDS/blob/1535aef6276670fe215a3cfb35510966409497db/src/cpp/rtps/builtin/discovery/endpoint/EDP.cpp#L629), [2](https://github.com/eProsima/Fast-DDS/blob/1535aef6276670fe215a3cfb35510966409497db/src/cpp/rtps/builtin/discovery/endpoint/EDP.cpp#L1195), ...).

Barry-Xu-2018:
Now interface is changed. We only use matched event in rmw layer (https://github.com/ros2/rmw/pull/331).  
So code changes are quite significant.  
Besides, merged implementation of inconsistent topic leads to many conflicts with old commits while doing rebase.
So I will merge all commits in this PR to one commit and then do rebase.

Barry-Xu-2018:
Check failure is because rmw package is old in CI environment.

mjcarroll (approved):

PR #653, opened 2022-12-14
[rmw_fastrtps] Improve handling of dynamic discovery

gbiggs:
This PR adds to `rmw_fastrtps` functionality necessary to support the [improved handling of dynamic discovery](https://github.com/ros2/ros2/issues/1359).

It adds handling of the new discovery parameters passed down from `rcl`.

This implementation allows a host specified in static peers to be a host name or an IPv4 or an IPv6 addresses. It will automatically manage translating between these to ensure that the static peer is correctly identified, no matter what it reports itself as.

Current status: The desired functionality is mostly complete. Most of the entries in the [goal matrix](https://discourse.ros.org/t/proposed-changes-to-how-ros-performs-discovery-of-nodes/27640) are working correctly. There is currently a bug where specifying a static peer appears to disable dynamic discovery. This needs to be resolved.


ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-02-16/29927/1

sloretz:
@EduPonz Any ETA on when the changes from this branch https://github.com/eProsima/Fast-DDS/compare/master...poc/ros2-iron-discovery-options will get merged into Fast-DDS? They're required for these discovery changes to work with Fast-DDS.

I'm hoping we can get this in before the upcoming ROS Iron RMW freeze deadline.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -46,16 +54,16 @@
...
 
-enum class publishing_mode_t
-{
-  ASYNCHRONOUS,  // Asynchronous publishing mode
-  SYNCHRONOUS,   // Synchronous publishing mode
-  AUTO           // Use publishing mode set in XML file or Fast DDS default
+enum class publishing_mode_t {
+  ASYNCHRONOUS, // Asynchronous publishing mode
+  SYNCHRONOUS,  // Synchronous publishing mode
+  AUTO          // Use publishing mode set in XML file or Fast DDS default
```
wjwwood, line 66: nitpick: This change seems unnecessary to me, maybe we should drop it?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -130,77 +138,107 @@ typedef struct CustomParticipantInfo
...
-    rmw_dds_common::Context * context)
-  : context(context),
-    identifier_(identifier)
-  {}
+  explicit ParticipantListener(const char *identifier,
+                               rmw_dds_common::Context *context,
+                               const std::string &hostname,
+                               const rmw_discovery_options_t *discovery_options)
+      : context(context), identifier_(identifier), my_hostname_(hostname),
+        discovery_options_(*discovery_options) {}
```
wjwwood, line 149: nitpick: it's weird that we changed the style as we did this. It makes it hard to understand what actually changed due to the whitespace change.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+                                                    proxyData.guid()),
+            is_reader);
+      }
+    }
+  }
+
+  std::unordered_set<std::string>
+  get_other_static_peers(
+    const std::map<std::string, std::vector<uint8_t>> &map) {
+    std::unordered_set<std::string> result;
```
wjwwood, line 286: nitpick:

```suggestion
  std::unordered_set<std::string>
  get_other_static_peers(const std::map<std::string, std::vector<uint8_t>> &map) {
    std::unordered_set<std::string> result;
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+
+    auto static_peers_entry = map.find("staticpeers");
+    if (static_peers_entry != map.end()) {
+      auto static_peers_string = std::string(static_peers_entry->second.begin(),
+                                             static_peers_entry->second.end());
+      std::string delimiter(",");
+      size_t start = 0;
+      size_t end = 0;
+      while ((end = static_peers_string.find(delimiter, start)) !=
+             std::string::npos) {
```
wjwwood, line 296: nitpick:

```suggestion
      while (
        (end = static_peers_string.find(delimiter, start)) != std::string::npos
      ) {
```

wjwwood: Also, you might consider just using https://github.com/ros2/rcpputils/blob/39b20134e571ba74baa7c77750eab586da90b7a5/include/rcpputils/split.hpp#L79-L95, because I think the crux of this block of code is splitting the string right?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+          RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT ==
+              discovery_options_.automatic_discovery_range) {
+
+        if (!is_static_peer(hostname, other_static_peers)) {
+          should_ignore = true;
+        }
+      }
+    }
+
+    return should_ignore;
```
wjwwood, line 343: maybe?

```suggestion
    if (RMW_AUTOMATIC_DISCOVERY_RANGE_OFF ==
          discovery_options_.automatic_discovery_range) {
      return true;
    }
    if (hostname != my_hostname_) {
      if (RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST ==
              discovery_options_.automatic_discovery_range ||
          RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT ==
              discovery_options_.automatic_discovery_range) {

        if (!is_static_peer(hostname, other_static_peers)) {
          return true;
        }
      }
    }

    return false;
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+  bool is_static_peer(const std::string &hostname,
+                      const std::unordered_set<std::string> &other_static_peers) {
+
+    using namespace rmw_fastrtps_shared_cpp;
+    // Check if the host is a static peer on our list
+    auto aliases = utils::get_peer_aliases(hostname);
+    for (size_t ii = 0; ii < discovery_options_.static_peers_count; ++ii) {
+      if (hostname == discovery_options_.static_peers[ii].peer_address) {
+        RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_shared_cpp",
+                                "Matching host in our static peer list");
```
wjwwood, line 356: I wonder if this log message is actually useful or not, because without any of the related information for context it doesn't tell you much. Maybe either remove it or add the hostname and peer address or something.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+        return true;
+      }
+
+      if (aliases.count(discovery_options_.static_peers[ii].peer_address) > 0)
+      {
+        RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_shared_cpp",
+                              "Matching host in our static peer list");
```
wjwwood, line 363: Same comment for this one.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+      }
+    }
+
+    auto peer_aliases = utils::get_peer_aliases(my_hostname_);
+
+    for (const auto &alias: peer_aliases)
+    {
+      if (std::find(other_static_peers.begin(), other_static_peers.end(), alias) != other_static_peers.end()) {
+        RCUTILS_LOG_DEBUG_NAMED("rmw_fastrtps_shared_cpp",
+                                "Matched us in their static peer list");
```
wjwwood, line 374: And here.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -235,16 +273,115 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
...
+        return true;
       }
     }
+    return false;
   }
 
-  rmw_dds_common::Context * context;
-  const char * const identifier_;
+  rmw_dds_common::Context *context;
+  const char *const identifier_;
```
wjwwood, line 382: Again this (odd) style changes make this much harder to review. It might actually be worth reworking this to not have any incidental style changes just so it's easier to understand later (as a separate commit).

rmw_fastrtps_shared_cpp/src/utils/net_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "rmw_fastrtps_shared_cpp/utils/net_utils.hpp"
+
+namespace rmw_fastrtps_shared_cpp
+{
+namespace utils
+{
+std::string get_fqdn_for_host(const std::string& hostname) {
+  struct addrinfo hints, *addresses = nullptr;
+
+  memset(&hints, 0, sizeof(hints));
```
wjwwood, line 22: memset is always a bit of a red flag for me, is there a better (more idiomatic C++ way) to initialize this? I've had to deal with `memset_s` and other issues in the past when considering safety standards, which is annoying, but for that reason I try to avoid it.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
arjo129: I considered that, but this is more efficient as it does not do any memory allocations so I kept it.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Removed whitespace changes in `custom_participant_info.hpp` in 65de1fb

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Removed whitespace changes in `custom_participant_info.hpp` in 65de1fb

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Removed whitespace changes in `custom_participant_info.hpp` in 65de1fb

rmw_fastrtps_shared_cpp/src/utils/net_utils.cpp
sloretz: One way is to use brace initialization. I'm not sure how portable it is though. Do any of these structs have `#ifdef ... int some_platform_specific_field;`?

```c++
struct addrinfo hints{0, 0, 0, 0, 0, 0, 0, 0};
```

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Better error message in 5b29937

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Better error message in 5b29937

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Better error message in 5b29937

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Eliminated variable in favor of returning in fe77f29

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: Whitespace change in dfd7dd2

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
sloretz: whitespace change in a2577c5

rmw_fastrtps_shared_cpp/src/utils/net_utils.cpp
sloretz: Speaking of which, does this even work on Windows?

sloretz:
Checking if this compiles on Windows:

build: `--packages-up-to rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_fastrtps_shared_cpp`
test: `--packages-select rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_fastrtps_shared_cpp`
[Build Status](https://ci.ros2.org/job/ci_windows/19017/)

rmw_fastrtps_shared_cpp/src/utils/net_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+      if (getaddrinfo(peer.c_str(), nullptr, nullptr, &addresses) != 0) {
+        // Failed to look up the hostname
+        RCUTILS_LOG_WARN_NAMED("rmw_fastrtps_shared_cpp",
+                               "Could not look up hostname %s", peer.c_str());
+      } else {
+        for (struct addrinfo *address_entry = addresses;
+             address_entry != nullptr; address_entry = address_entry->ai_next) {
+          void * address_data = nullptr;
+          char address_string[INET6_ADDRSTRLEN];
+          memset(address_string, 0, NI_MAXHOST);
```
wjwwood, line 63: This line produces this warning with clang:

```
/Users/william/ros2_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/utils/net_utils.cpp:63:11: warning: 'memset' will always overflow; destination buffer has size 46, but size argument is 1025 [-Wfortify-source]
          memset(address_string, 0, NI_MAXHOST);
          ^
1 warning generated.
```

wjwwood: Should probably be:

```suggestion
          memset(address_string, 0, INET6_ADDRSTRLEN);
```

wjwwood (changes requested):

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,71 @@ rmw_fastrtps_shared_cpp::create_participant(
...
+        eprosima::fastrtps::rtps::DiscoveryProtocol_t::NONE;
+      break;
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST: {
+        // Clear the list of multicast listening locators
+        domainParticipantQos.wire_protocol().builtin.metatrafficMulticastLocatorList.clear();
+        // Add a unicast locator to prevent creation of default multicast locator
+        eprosima::fastrtps::rtps::Locator_t default_unicast_locator;
+        domainParticipantQos.wire_protocol()
+        .builtin.metatrafficUnicastLocatorList.push_back(default_unicast_locator);
+        // Add a shared memory transport
```
MiguelCompany, line 183: Note: It is recommended to disable the built-in transports when you don't want them to be instantiated (like in this case)

```suggestion
        // Disabling built-in transports, since we are configuring our own.
        domainParticipantQos.transport().use_builtin_transports = false;
        // Add a shared memory transport
```

sloretz: Disabled built-in transports in d1e94da

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,78 @@ rmw_fastrtps_shared_cpp::create_participant(
...
 
+  // Configure discovery
+  switch (discovery_options->automatic_discovery_range) {
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET:
+      RMW_SET_ERROR_MSG("automatic discovery range must be set");
+      return nullptr;
+      break;
+    case RMW_AUTOMATIC_DISCOVERY_RANGE_OFF:
+      domainParticipantQos.wire_protocol().builtin.discovery_config.discoveryProtocol =
+        eprosima::fastrtps::rtps::DiscoveryProtocol_t::NONE;
```
MiguelCompany, line 174: This will prevent readers and writers on the local participant to be discovered, which I don't think is the desired behavior.

If what we want is that only the local participant is discovered, we could do the following:

```suggestion
      // Limit the number of participants to 1 (the local participant)
      domainParticipantQos.allocation().participants.initial = 1;
      domainParticipantQos.allocation().participants.maximum= 1;
      domainParticipantQos.allocation().participants.increment= 0;

      // Clear the list of multicast listening locators
      domainParticipantQos.wire_protocol().builtin.metatrafficMulticastLocatorList.clear();
      // Add a unicast locator to prevent creation of default multicast locator
      eprosima::fastrtps::rtps::Locator_t default_unicast_locator;
      domainParticipantQos.wire_protocol()
      .builtin.metatrafficUnicastLocatorList.push_back(default_unicast_locator);
```

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,78 @@ rmw_fastrtps_shared_cpp::create_participant(
...
-    domainParticipantQos.transport().use_builtin_transports = false;
-
-    // Add a UDPv4 transport with only localhost enabled
-    auto udp_transport = std::make_shared<eprosima::fastdds::rtps::UDPv4TransportDescriptor>();
-    udp_transport->interfaceWhiteList.emplace_back("127.0.0.1");
-    domainParticipantQos.transport().user_transports.push_back(udp_transport);
+  // Add initial peers if LOCALHOST or SUBNET are used
+  if (
+    RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST == discovery_options->automatic_discovery_range ||
+    RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET == discovery_options->automatic_discovery_range)
```
MiguelCompany, line 210: Do we really need to add the static peers for `RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET`?
If so, see my next comment.

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,78 @@ rmw_fastrtps_shared_cpp::create_participant(
...
+  }
 
-    // Add SHM transport if available
-    auto shm_transport = std::make_shared<eprosima::fastdds::rtps::SharedMemTransportDescriptor>();
-    domainParticipantQos.transport().user_transports.push_back(shm_transport);
+  if (RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST == discovery_options->automatic_discovery_range) {
+    // Add localhost as a static peer
+    eprosima::fastrtps::rtps::Locator_t peer;
+    eprosima::fastrtps::rtps::IPLocator::setIPv4(peer, "127.0.0.1");
+    domainParticipantQos.wire_protocol().builtin.initialPeersList.push_back(peer);
```
MiguelCompany, line 237: For the case of `RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET`, if `initialPeersList` is not empty, we should add the default multicast address to it.

MiguelCompany (changes requested):

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,78 @@ rmw_fastrtps_shared_cpp::create_participant(
   eprosima::fastdds::dds::DomainParticipantQos domainParticipantQos =
     eprosima::fastdds::dds::DomainParticipantFactory::get_instance()->get_default_participant_qos();
 
+  // Configure discovery
+  switch (discovery_options->automatic_discovery_range) {
```
MiguelCompany, line 167: Additional note: I think we should decide whether to override the value to `RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT` when `RMW_FASTRTPS_USE_QOS_FROM_XML=1`.

Whatever we decide, should be documented on the table [here](https://github.com/ros2/rmw_fastrtps#full-qos-configuration)

rmw_fastrtps_shared_cpp/src/participant.cpp
sloretz: I think so. It allows a peer with `SUBNET` range to also discover a peer that's only listening via unicast.

rmw_fastrtps_shared_cpp/src/participant.cpp
sloretz: To the initial peer list, or as a multicast locator? I assumed the latter in cf9bf83

rmw_fastrtps_shared_cpp/src/participant.cpp
sloretz: Assuming it's possible for an XML to not contain any participant discovery related settings, I would recommend `RMW_FASTRTPS_USE_QOS_FROM_XML` not change the automatic discovery range so that a user can combine both options.

The intent here was the user would set `RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT` to prevent this code from overriding any participant discovery settings that might be configured out-of-band, such as via an xml file.

Added documentation in a5dae76

rmw_fastrtps_cpp/test/test_get_native_entities.cpp
```diff
@@ -56,8 +57,11 @@ class TestNativeEntities : public ::testing::Test
 
   void TearDown() override
   {
-    rmw_ret_t ret = rmw_destroy_node(node);
-    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+    rmw_ret_t ret;
+    if (nullptr != node) {
+      ret = rmw_destroy_node(node);
+      EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+    }
```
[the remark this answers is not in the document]
sloretz: I removed this change in https://github.com/ros2/rmw_fastrtps/pull/653/commits/60ce2250a8079a8d8cf1a9bd54badc563148ee0a

rmw_fastrtps_shared_cpp/src/participant.cpp
sloretz: I applied your suggestion in https://github.com/ros2/rmw_fastrtps/pull/653/commits/581ab867b71f2d0cc4aa30055384cd0119e91f6f

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -161,19 +163,98 @@ rmw_fastrtps_shared_cpp::create_participant(
...
-    domainParticipantQos.transport().user_transports.push_back(shm_transport);
+  if (
+    RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET == discovery_options->automatic_discovery_range &&
+    domainParticipantQos.wire_protocol().builtin.initialPeersList.size())
+  {
+    // Need to add default the default multicast address
+    eprosima::fastrtps::rtps::Locator_t locator;
+    eprosima::fastrtps::rtps::IPLocator::setIPv4(locator, 239, 255, 0, 1);
+    domainParticipantQos.wire_protocol()
+    .builtin.metatrafficMulticastLocatorList.push_back(locator);
```
MiguelCompany, line 257: I meant the `initialPeersList`. Announcements are always sent to the initial peer list. When it is empty, the middleware sets its value to the `metatrafficMulticastLocatorList`.

So if we want to publish announcements to 192.168.1.10 and the multicast address, both should be in the initial peer list.

```suggestion
    .builtin.initialPeersList.push_back(locator);
```

MiguelCompany (changes requested):

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -142,8 +142,10 @@ class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantList
 
   void on_participant_discovery(
     eprosima::fastdds::dds::DomainParticipant *,
-    eprosima::fastrtps::rtps::ParticipantDiscoveryInfo && info) override
+    eprosima::fastrtps::rtps::ParticipantDiscoveryInfo && info,
+    bool & should_be_ignored) override
   {
+    should_be_ignored = false;
```
MiguelCompany, line 148: These changes will only build with Fast DDS 2.10.0 onwards, so I suggest changing the version here:

https://github.com/ros2/rmw_fastrtps/blob/cf9bf8323e34956f98e925cc8c1501e63ccf8eb5/rmw_fastrtps_shared_cpp/CMakeLists.txt#L49-L50

MiguelCompany (approved):
Nice job!

sloretz:
Need one small fix/enhancement to Fast-DDS https://github.com/eProsima/Fast-DDS/pull/3437

wjwwood (approved):

PR #654, opened 2022-12-20
Implement inconsistent topic event

clalancette:
This PR implements the inconsistent topic event as part of https://github.com/ros2/ros2/issues/1361 .

Note that as it stands, this is not hooked up inside of Fast-DDS, so this event will never fire.  That said, this should be enough to make it work once it is hooked up inside of Fast-DDS.

To make it so that this could be hooked up, a bunch of internal refactoring had to be done.  In particular, `rmw_fastrtps` conflated the events that happen from Fast-DDS (the "listeners" in DDS parlance), with the events that need to be handed to the rmw/rcl layer (the RMW events in ROS 2 parlance).  Since inconsistent topic uses a different listener, that had to be broken apart.  That's what the first part of this series does, and then the last patch actually does the inconsistent topic implementation.


fujitatomoya (approved):
lgtm

fujitatomoya:
@MiguelCompany @iuhilnehc-ynos can you take a look at this ?

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -225,7 +232,7 @@ rmw_fastrtps_cpp::create_publisher(
   rmw_fastrtps_shared_cpp::TopicHolder topic;
   if (!rmw_fastrtps_shared_cpp::cast_or_create_topic(
       dds_participant, des_topic,
-      topic_name_mangled, type_name, topic_qos, true, &topic))
+      topic_name_mangled, type_name, topic_qos, true, &topic, info->topic_listener_))
```
MiguelCompany, line 235: Since `cast_or_create_topic` would only create the topic the first time a specific topic name is used, this could lead to using a destroyed listener.

Consider the following situation:
- Publisher for topic `topic_a` is created => will set its `CustomTopicListener` as listener for `topic_a`
- Subscription for topic `topic_a`  is created. Its `CustomSubscriptionTopicListener` will not be used.
- Publisher is deleted => its `CustomTopicListener` is destroyed
- An inconsistent topic is detected. This will try to invoke the registered listener, which has already been destroyed

MiguelCompany (changes requested):

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -242,6 +242,7 @@ create_subscription(
   info->dds_participant_ = dds_participant;
   info->subscriber_ = subscriber;
   info->topic_name_mangled_ = topic_name_mangled;
+  info->topic_ = topic_holder.topic;
```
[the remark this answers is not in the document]
clalancette: That is there so that we can call `get_inconsistent_topic_status` during a `SubListener::take_event`:
https://github.com/ros2/rmw_fastrtps/pull/654/files#diff-f393290aa0cfff761c6a3114fc26319b9edd9fd3c915db07165399b79e69394bR204

rmw_fastrtps_cpp/src/publisher.cpp
clalancette: Oh, I didn't realize that this was shared for every single topic in a process.  I'll have to make sure it always gets registered, even if the topic is already there.  Thanks for the hint.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -216,7 +203,7 @@ class SubListener : public EventListenerInterface, public eprosima::fastdds::dds
 
   std::mutex on_new_message_m_;
 
-  std::mutex discovery_m_;
+  mutable std::mutex discovery_m_;
```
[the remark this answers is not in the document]
clalancette: Agreed.  Fixed in 697f75591d157e39f8d8c839b03930c2a2c71873

rmw_fastrtps_cpp/src/publisher.cpp
```diff
@@ -164,6 +164,7 @@ rmw_fastrtps_cpp::create_publisher(
   auto cleanup_info = rcpputils::make_scope_exit(
     [info, dds_participant]() {
       delete info->publisher_event_;
+      delete info->topic_listener_;
```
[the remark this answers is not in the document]
clalancette: Good call.  Fixed in 5c5490a9ec969be6141ffa6a0712c0064f7d261b

rmw_fastrtps_cpp/src/publisher.cpp
clalancette: @MiguelCompany Could you maybe give some pointers on what we are trying to accomplish with https://github.com/ros2/rmw_fastrtps/blob/e12ff1a1be8ded9623ff770a1f4f9e4b51ca8f53/rmw_fastrtps_shared_cpp/src/utils.cpp#L56-L91 ?

In particular, it seems to do some weird things for subscribers.  For instance, if you are the first and only subscription to this topic in the system, then `desc` will be nullptr, and `is_writer_topic` will be false, so `topic_holder->topic` will remain nullptr and the function will return true.  However, if there is already a publisher *or subscriber* on the topic in this process, then `desc` will be non-null, and thus we will call `participant->create_topic()` and return true.  It seems a bit weird to me to be doing different things depending on what else has been already created in the process.

Relatedly, what is the downside to calling `participant->create_topic()` for every new subscription or publisher that comes in?

clalancette: I figured this all out, and rewrote this bit in #669.  I then rebased this on top of that, so this should be ready to go.

clalancette:
@methylDragon @fujitatomoya @MiguelCompany I could use another review on this one, which has undergone significant changes since the last review and #669 .  Thanks in advance!

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_event_info.hpp
```diff
@@ -56,6 +58,8 @@ class EventListenerInterface
     return event_guard[event_type];
   }
 
+  virtual void update_inconsistent_topic(uint32_t total_count, uint32_t total_count_change) = 0;
```
methylDragon, line 61: I can't comment on the appropriate line, but I think `take_event`'s docstring is outdated

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
```diff
@@ -23,15 +24,74 @@
...
+{
+}
+
+eprosima::fastdds::dds::StatusCondition & RMWPublisherEvent::get_statuscondition() const
 {
   return publisher_info_->data_writer_->get_statuscondition();
 }
 
-bool PubListener::take_event(
+bool RMWPublisherEvent::take_event(
```
methylDragon, line 94: It might just be me but I find it confusing that `RMWPublisherEvent` would be able to take events (I think of the event as a struct of information that can be manipulated, as opposed to having any listening capabilities, which is what seems to be happening in the body of this method.)

I'm wondering if there's a better name for the class. Or it might just be a bad name for the method in the current version of the library, or I might just be confusing myself.

Perhaps `update_event`?

If changes are made addressing this comment, they should be rolled out to the subscriber and event info too.

methylDragon (approved):
No blocking comments, just a bunch of potential clarity changes.

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  std::lock_guard<std::mutex> lck(topic_name_to_topic_mutex_);
+  std::map<std::string, std::unique_ptr<UseCountTopic>>::const_iterator it =
+    topic_name_to_topic_.find(topic->get_name());
+
+  if (it != topic_name_to_topic_.end()) {
+    it->second->use_count--;
+    it->second->topic_listener->remove_event_listener(event_listener);
+    if (it->second->use_count <= 0) {
```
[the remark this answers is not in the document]
clalancette: It should never be the case, no.  This was merely me being cautious.  It really shouldn't be a performance concern either way, so I'm inclined to leave it as-is.  Let me know if you think differently.

MiguelCompany:
@clalancette I won't have time till next week, but I promise to give you a review

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -197,9 +197,9 @@ rmw_create_client(
       delete info->pub_listener_;
       delete info->listener_;
       rmw_fastrtps_shared_cpp::remove_topic_and_type(
-        participant_info, info->response_topic_, info->response_type_support_);
+        participant_info, nullptr, info->response_topic_, info->response_type_support_);
       rmw_fastrtps_shared_cpp::remove_topic_and_type(
-        participant_info, info->request_topic_, info->request_type_support_);
+        participant_info, nullptr, info->request_topic_, info->request_type_support_);
```
MiguelCompany, line 202: For consistency with publishers and subscriptions, I would move the calls to `remove_topic_and_type` before the deletion of listeners.

If this is taken, applies also to services, and also to the dynamic RMW

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -55,9 +57,30 @@ enum class publishing_mode_t
...
+  on_inconsistent_topic(
+    eprosima::fastdds::dds::Topic * topic,
+    eprosima::fastdds::dds::InconsistentTopicStatus status) override;
+
+  void add_event_listener(EventListenerInterface * event_listener);
+
+  void remove_event_listener(EventListenerInterface * event_listener);
+
+private:
+  std::set<EventListenerInterface *> event_listeners_;
```
MiguelCompany, line 79: Should this collection be protected against access from different threads?

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
```diff
@@ -0,0 +1,124 @@
...
+      delete it->second->topic_listener;
+
+      participant_->delete_topic(it->second->topic);
```
MiguelCompany, line 114: I would change the order of these two lines. Otherwise, the listener could be invoked after being deleted.

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
```diff
@@ -0,0 +1,124 @@
...
+    std::unique_ptr<UseCountTopic>>::const_iterator it = topic_name_to_topic_.find(topic_name);
+  if (it == topic_name_to_topic_.end()) {
+    // Not already in the map, we need to add it
+    topic = participant_->create_topic(topic_name, type_name, topic_qos);
+
+    auto uct = std::make_unique<UseCountTopic>();
+    uct->topic = topic;
+    uct->topic_listener = new CustomTopicListener(event_listener);
+    uct->use_count = 1;
+
```
MiguelCompany, line 84: Either:
1. create the `topic_listener` before the topic, and then call `create_topic(topic_name, type_name, topic_qos, topic_listener)`
2. Call `topic->set_listener(topic_listener)`

MiguelCompany (changes requested):

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
```diff
@@ -0,0 +1,124 @@
...
+  event_listeners_.erase(event_listener);
+}
+
+eprosima::fastdds::dds::Topic * CustomParticipantInfo::find_or_create_topic(
+  const std::string & topic_name,
+  const std::string & type_name,
+  const eprosima::fastdds::dds::TopicQos & topic_qos,
+  EventListenerInterface * event_listener)
+{
+  eprosima::fastdds::dds::Topic * topic = nullptr;
```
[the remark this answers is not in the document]
clalancette: Actually, we can't do that, as there are paths that come through here with a nullptr.  In particular, when creating a service server or a service client, we currently call this will nullptr.  We could change that, but at the moment there is no real reason to.  Maybe this is something to change in the future, but I'm going to leave this check out for now.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_event_info.hpp
clalancette: Oh, never mind,  I actually fixed it in a0919117f9db938c217dc1034165b035bcd4e0db .

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/utils.hpp
```diff
@@ -75,6 +75,7 @@ RMW_FASTRTPS_SHARED_CPP_PUBLIC
 void
 remove_topic_and_type(
   CustomParticipantInfo * participant_info,
+  EventListenerInterface * event_listener,
```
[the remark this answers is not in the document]
clalancette: Fixed in a0919117f9db938c217dc1034165b035bcd4e0db.

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -337,7 +337,12 @@ rmw_fastrtps_shared_cpp::destroy_participant(CustomParticipantInfo * participant
   // Remove topics
   eprosima::fastdds::dds::TypeSupport dummy_type;
   for (auto topic : topics_to_remove) {
-    remove_topic_and_type(participant_info, topic, dummy_type);
+    // Passing nullptr as the EventListenerInterface isn't ideal here;
+    // it means that for each individual topic we remove, we won't remove
+    // the corresponding TopicListener callback (if any).  However, this
+    // shouldn't matter as we are going to shortly deleting the entire
```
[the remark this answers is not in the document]
clalancette: I rewrote the whole thing in 72ab8a2f3dab994094aac3865b62877acc2a9c16.

rmw_fastrtps_cpp/src/rmw_client.cpp
clalancette: Good call.  Fixed in 222d3ce43fffe6191a7ee544bb21bce915f688e1.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
clalancette: Yes, good idea.  Fixed in 222d3ce43fffe6191a7ee544bb21bce915f688e1.

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
clalancette: Fixed in 222d3ce43fffe6191a7ee544bb21bce915f688e1.

rmw_fastrtps_shared_cpp/src/custom_participant_info.cpp
clalancette: Fixed in 222d3ce43fffe6191a7ee544bb21bce915f688e1.

rmw_fastrtps_shared_cpp/src/custom_publisher_info.cpp
clalancette: So what the `RMWPublisherEvent` class does is to collect data about events that happened to the publisher, and make them available to the RMW layer.  In this new refactor, this class doesn't directly register for any event callbacks; that is done via the `CustomDataWriterListener` and `CustomTopicListener`, and when the callbacks fire they call their saved handle to this class to update the data.  Then the layers about the RMW get notified of the event, and can come fetch the event via `take_event`.

So to me, the `take_event` makes sense; the upper layers are taking the data for this event.  The name of the class is more up-in-the-air; it really is a collection of publisher events collected for the RMW layer, hence the current name.  But I'm open to naming it something different, if you have a better idea.  Let me know what you think.

rmw_fastrtps_shared_cpp/src/participant.cpp
```diff
@@ -337,7 +337,12 @@ rmw_fastrtps_shared_cpp::destroy_participant(CustomParticipantInfo * participant
   // Remove topics
   eprosima::fastdds::dds::TypeSupport dummy_type;
   for (auto topic : topics_to_remove) {
-    remove_topic_and_type(participant_info, topic, dummy_type);
+    // Passing nullptr as the EventListenerInterface isn't ideal here;
```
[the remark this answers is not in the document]
clalancette: And done in ff96eaaa064122f746df47f6a84b1681eb9397a9.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -80,143 +119,94 @@ struct CustomSubscriberInfo : public CustomEventInfo
...
+    rmw_event_callback_t callback) override;
+
+  RMW_FASTRTPS_SHARED_CPP_PUBLIC
+  void update_inconsistent_topic(uint32_t total_count, uint32_t total_count_change) override;
+
+  void add_publisher(eprosima::fastrtps::rtps::GUID_t guid);
+
+  void remove_publisher(eprosima::fastrtps::rtps::GUID_t guid);
+
+  size_t publisher_count() const;
```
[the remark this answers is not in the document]
clalancette: > It's a little unintuitive to me that a class called RMWSubscriptionEvent (as opposed to SubListener) would be able to keep track of publishers.

Yeah, it's there strictly to have a count of unique publishers attached to this subscription (`RMWPublisherInfo` similarly has `add_subscription`/`remove_subscription`, for the same reason).

I'm going to go ahead and change the name of these to `track_unique_publisher`/`untrack_unique_publisher` to make this a little clearer.  I'll also add some documentation to make it clear why we are doing this.  And I'll also make the same change to `RMWPublisherEvent`.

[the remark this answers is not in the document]
clalancette: And this is now done in 817ddd940c17a798f0864fea335df69e2e1d21a8.

clalancette:
OK, I think I've addressed all comments except for one here.  @methylDragon and @MiguelCompany thanks for the reviews so far.  When you get a chance, could you please take another look?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -76,6 +76,7 @@ class CustomTopicListener final : public eprosima::fastdds::dds::TopicListener
   void remove_event_listener(EventListenerInterface * event_listener);
 
 private:
+  std::mutex event_listeners_mutex_;
```
MiguelCompany, line 79: nit: I guess we now should add `RCPPUTILS_TSA_GUARDED_BY` to `event_listeners_`

rmw_fastrtps_cpp/src/rmw_client.cpp
```diff
@@ -194,12 +194,12 @@ rmw_create_client(
...
-      delete info->listener_;
       rmw_fastrtps_shared_cpp::remove_topic_and_type(
-        participant_info, info->response_topic_, info->response_type_support_);
+        participant_info, nullptr, info->response_topic_, info->response_type_support_);
       rmw_fastrtps_shared_cpp::remove_topic_and_type(
-        participant_info, info->request_topic_, info->request_type_support_);
+        participant_info, nullptr, info->request_topic_, info->request_type_support_);
+      delete info->pub_listener_;
+      delete info->listener_;
       delete info;
```
MiguelCompany, line 203: Looking at these three consecutive `delete` calls makes me wonder: "Shall we use `unique_ptr` for the listeners so we don't have to worry about their deletion?"

This can be done on a follow-up, though.

MiguelCompany (changes requested):

rmw_fastrtps_cpp/src/rmw_client.cpp
clalancette: > Looking at these three consecutive delete calls makes me wonder: "Shall we use unique_ptr for the listeners so we don't have to worry about their deletion?"

Yeah, it's a good question.  The one problem there is that `std::make_unique` doesn't support the `nothrow` variant, which we currently use.  It looks like we could probably define something ourselves (see https://stackoverflow.com/a/57092496), but I think that is probably follow-up work.  So I'm not going to do this here and we can consider it separately.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
clalancette: Good call, fixed in 8e73176203e41ca355636163f54b5b6f6915345b.

MiguelCompany (approved):

iuhilnehc-ynos:
Sorry for the late review and it's not a blocking comment/question as well.

Can we add a new struct derived from `eprosima::fastdds::dds::Topic` to keep the `EventListenerInterface` as a member inside, and then use two `dynamic_cast` operations for checking the `new struct` and `Topic` by order in `remove_topic_and_type`?

something might be done as below,

add overloaded functions for `CustomParticipantInfo::find_or_create_topic` and `CustomParticipantInfo::delete_topic` for publisher and subscription

so we can get the following benefit

1. the client and service can take the original source code if they don't care about the event_listener.
2. there is no `rmw_fastrtps_shared_cpp::remove_topic_and_type( 'info->publisher_event_ / nullptr' ) with an extra parameter unrelated to the method name called.

clalancette:
@iuhilnehc-ynos I partially understand what you are saying, but not completely.

That said, I'm inclined to go with this PR as-is, and leave that for follow-up work.  In particular, until Fast-DDS can produce inconsistent topic events, it is somewhat hard to tell exactly how this will all shake out.

So what I'm going to do here is to run one more CI on the main issue, and then merge this as-is.  I'll then open an issue for your follow-up thoughts.  Thanks for taking a look.

clalancette:
And I've opened up #675 to track the improvement.

PR #655, opened 2023-01-03
Runtime Interface Reflection: rmw_fastrtps

methylDragon:
This PR is part of the runtime interface reflection subscription feature of REP-2011: https://github.com/ros2/ros2/issues/1374

# Description
This PR includes implementation of the new rmw APIs for supporting runtime interface reflection:
- Getting serialization support
- Converting from serialized data to dynamic data
- rmw_features labels

It also includes:
- Utilities for type name mangling
- An alternate FastDDS subscriber creation pathway supporting runtime types
  - (There's quite a bit of code duplication here, I'm not sure what the best approach is for whittling it down)

TODO:
- Implement take_dynamic_data


rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp
```diff
@@ -147,6 +147,22 @@ class RMWSubscriptionEvent final : public EventListenerInterface
...
+  //   const eprosima::fastrtps::types::TypeObject *,
+  //   eprosima::fastrtps::types::DynamicType_ptr dyn_type)) final
+  // {
+  // NOTE(methylDragon): The dynamic type deferred case is !! NOT SUPPORTED !!
+  //                     This is because currently subscriptions are required to have the type at
+  //                     construction to create the listener. Deferring it means that the listener
+  //                     construction will have to be deferred, and that would require logic changes
+  //                     elsewhere (e.g. to check for listener initialization status), which is
+  // }
+
```
wjwwood, line 165: Should this just be removed then? I think (as I've stated elsewhere) we should just not use `NOTE`. I think this should either be a TODO or not left in.

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -103,6 +129,351 @@ create_subscription(
...
+  info->type_support_ = fastdds_type;
+
+  // NOTE(methylDragon): I'm not sure if this is essential or not...
+  //                     It doesn't appear in the dynamic type example for FastDDS though
+  // if (!rmw_fastrtps_shared_cpp::register_type_object(type_support, type_name)) {
+  //   RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+  //     "failed to register type object with incompatible type %s",
+  //     type_name.c_str());
+  //   return nullptr;
+  // }
```
wjwwood, line 312: I think at this point we should remove snippets like this.

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -103,6 +129,351 @@ create_subscription(
...
+  const rosidl_message_type_support_t * type_support,
+  const char * topic_name,
+  const rmw_qos_profile_t * qos_policies,
+  const rmw_subscription_options_t * subscription_options,
+  bool keyed)
+{
+  // NOTE(methylDragon): The dynamic type deferred case is !! NOT SUPPORTED !!
+  //                     This is because it's difficult as-is to create a subscription without
+  //                     already having the type. Too much restructuring is needed elsewhere to
+  //                     support deferral...
```
wjwwood, line 163: Should we remove this or is this suppose to be accompanied with a check of some sort?

wjwwood (approved):
lgtm, I had a few comments, but they're more questions and clean up of comments than anything. I'm less worried about the polish of this pr since it's not part of the user facing API and also its functionality is covered by examples and tests above.

wjwwood (approved):
I think my comments should still be addressed, but since they're all about code comments, I think we should merge this and fix them up in a follow up. @methylDragon can you make an issue for that if you don't have time to make a pr?

PR #657, opened 2023-01-11
Delay lock on message callback setters

### LINKED ISSUE #646: Deadlock when calling get_unread_count() and destroying subscription
I've identified a deadlock when a subscription (having assigned an `on_new_message_callback`) is being destroyed, while is also getting a message from a publisher. I'm using the latest master branch of FastDDS.

Attached, an example to reproduce the issue and the resulting cleaned backtrace.

[FastDDS-example-and-backtrace.zip](https://github.com/ros2/rmw_fastrtps/files/10125748/FastDDS-example-and-backtrace.zip)

You can run the example with GDB, wait until it deadlocks, and `ctrl+c` to check where happened.

The issue seems to be in https://github.com/ros2/rmw_fastrtps/blob/ccea7a356900a16a2872670b67952c99a72ab65d/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp#L166

According to @jefferyyjhsu, one mutex belongs to `rmw_fastrtps` while the other to `fast-dds`.

This issue happens since this PR got merged: https://github.com/ros2/rmw_fastrtps/pull/633/files#diff-26246d1a82280de3805240670a0e83425979cd7598ac0eaad21db840ebab7d7dR118-R122

Besides this, I can also see that in `custom_service_info.hpp` , there is no check about if `on_new_request_cb_` exists before calling it:
https://github.com/ros2/rmw_fastrtps/blob/ccea7a356900a16a2872670b67952c99a72ab65d/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_service_info.hpp#L209-L220

Let me know if you need any more details.

MiguelCompany:
This should fix #646 by calling `get_unread_count` before taking the lock on the mutex `on_new_message_m_`.

It provides similar fixes for the callback setter on subscriptions, services, and clients.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_client_info.hpp
@@ -127,10 +127,10 @@ public:
     const void * user_data,
     rmw_event_callback_t callback)
   {
-    std::unique_lock<std::mutex> lock_mutex(on_new_response_m_);
-
     if (callback) {
       auto unread_responses = get_unread_responses();
+
+      std::lock_guard<std::mutex> lock_mutex(on_new_response_m_);
 
       if (0 < unread_responses) {
         callback(user_data, unread_responses);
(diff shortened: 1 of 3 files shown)


fujitatomoya (commented):
@MiguelCompany I think this fixes subscription problem, but the same fix should be applied to `on_new_response_m_` and `on_new_request_m_` as well?

MiguelCompany:
> @MiguelCompany I think this fixes subscription problem, but the same fix should be applied to `on_new_response_m_` and `on_new_request_m_` as well?

@fujitatomoya You are right. I've pushed the changes and also changed the PR description accordingly

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17950/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12494/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18598/)

MiguelCompany:
@fujitatomoya Do you think this can be merged?

fujitatomoya:
@mergifyio backport humble

PR #662, opened 2023-01-19
add network notify function

jakymiws:


jakymiws:
Related to 
https://github.com/ros2/rcl/pull/1028
https://github.com/ros2/rmw_implementation/pull/213
https://github.com/ros2/rclcpp/pull/2086
https://github.com/ros2/rmw/pull/344

rmw_fastrtps_cpp/src/rmw_node.cpp
```diff
@@ -121,4 +123,30 @@ rmw_node_get_graph_guard_condition(const rmw_node_t * node)
   return rmw_fastrtps_shared_cpp::__rmw_node_get_graph_guard_condition(
     node);
 }
+
+rmw_ret_t
+rmw_notify_participant_dynamic_network_interface(rmw_context_t * context)
+{
+  auto impl = static_cast<CustomParticipantInfo *>(context->impl->participant_info);
```
alsora, line 130: should we check if `impl` is a valid pointer?

PR #665, opened 2023-01-26
Capture `std::bad_alloc` on deserializeROSmessage.

MiguelCompany:
This should fix crashes coming from serialization mismatches that cannot be detected by Fast CDR

Related to https://github.com/ros2/ros2_documentation/issues/3288#issuecomment-1404876431

### DIFF
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -119,6 +119,11 @@ bool TypeSupport::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
+  } catch (...) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Unhandled exception deserializing message of type %s.",
+      getName());
+    return false;
   }
 
   return true;
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -119,9 +119,9 @@ bool TypeSupport::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
-  } catch (...) {
+  } catch (std::bad_alloc &) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-      "Unhandled exception deserializing message of type %s.",
+      "'Bad alloc' exception deserializing message of type %s.",
       getName());
     return false;
   }
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -119,7 +119,7 @@ bool TypeSupport::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
-  } catch (std::bad_alloc &) {
+  } catch (const std::bad_alloc &) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "'Bad alloc' exception deserializing message of type %s.",
       getName());
--- a/rmw_fastrtps_dynamic_cpp/include/rmw_fastrtps_dynamic_cpp/TypeSupport_impl.hpp
+++ b/rmw_fastrtps_dynamic_cpp/include/rmw_fastrtps_dynamic_cpp/TypeSupport_impl.hpp
@@ -983,9 +983,9 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
-  } catch (...) {
+  } catch (std::bad_alloc &) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-      "Unhandled exception deserializing message of type %s.",
+      "'Bad alloc' exception deserializing message of type %s.",
       getName());
     return false;
   }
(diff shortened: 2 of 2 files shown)


fujitatomoya:
- service client (cyclonedds)

cannot get the result from server.

```bash
root@tomoyafujita:~/ros2_ws/colcon_ws# RMW_IMPLEMENTATION=rmw_cyclonedds_cpp ros2 run demo_nodes_cpp add_two_ints_client
^C[INFO] [1674755622.413766599] [rclcpp]: signal_handler(signum=2)
[ERROR] [1674755622.414263025] [add_two_ints_client]: Interrupted while waiting for response. Exiting.
```

- service server (Fast-DDS)

in this sample, cannot observe the core dump or crash. but it is obvious that de-serialization works unexpectedly.

```bash
root@tomoyafujita:~/ros2_ws/colcon_ws# cat DEFAULT_FASTRTPS_PROFILES.xml 
<?xml version="1.0" encoding="UTF-8" ?>
<profiles xmlns="http://www.eprosima.com/XMLSchemas/fastRTPS_Profiles">

  <!-- Default publisher profile -->
  <data_writer profile_name="default publisher profile" is_default_profile="true">
    <qos>
      <data_sharing>
        <kind>AUTOMATIC</kind>
      </data_sharing>
    </qos>
    <historyMemoryPolicy>DYNAMIC</historyMemoryPolicy>
  </data_writer>

  <data_reader profile_name="default subscription profile" is_default_profile="true">
    <qos>
      <data_sharing>
        <kind>AUTOMATIC</kind>
      </data_sharing>
    </qos>
    <historyMemoryPolicy>DYNAMIC</historyMemoryPolicy>
  </data_reader>
</profiles>
root@tomoyafujita:~/ros2_ws/colcon_ws# export RMW_FASTRTPS_USE_QOS_FROM_XML=1
root@tomoyafujita:~/ros2_ws/colcon_ws# RMW_IMPLEMENTATION=rmw_fastrtps_cpp ros2 run demo_nodes_cpp add_two_ints_server 
[INFO] [1674755608.560241436] [add_two_ints_server]: Incoming request
a: -2268473885457029753 b: 1
^C[INFO] [1674755623.645151231] [rclcpp]: signal_handler(signum=2)
```

fujitatomoya (approved):
lgtm, this should avoid unexpected exceptions during de-serialization.

fujitatomoya:
@alsora @mauropasse if possible, would you mind trying this patch to see if it can avoid the program crash? (just sending request to the server can get the program crashed, for me that is the security issue...)

rmw_fastrtps_cpp/src/type_support_common.cpp
```diff
@@ -119,6 +119,11 @@ bool TypeSupport::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
+  } catch (...) {
```
clalancette, line 122: Generally, we would prefer to have specific exceptions listed here, rather than just catch everything.

What exceptions can this deserialization throw?

christophfroehlich:
This fixes #733 if I build this package from source in the humble container. Is there any chance that this gets merged and backported to humble?

rmw_fastrtps_cpp/src/type_support_common.cpp
clalancette: Looking at the original report, it looks like we should just catch `std::bad_alloc` for now.  @MiguelCompany if you make that change and rebase this PR onto the latest, I'll be happy to approve and run CI on it.

rmw_fastrtps_cpp/src/type_support_common.cpp
```diff
@@ -119,6 +119,11 @@ bool TypeSupport::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
+  } catch (std::bad_alloc &) {
```
clalancette, line 122:
```suggestion
  } catch (const std::bad_alloc &) {
```

(it probably makes no difference in performance here, but it is good practice in my opinion)

rmw_fastrtps_dynamic_cpp/include/rmw_fastrtps_dynamic_cpp/TypeSupport_impl.hpp
```diff
@@ -983,6 +983,11 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
       "Fast CDR exception deserializing message of type %s.",
       getName());
     return false;
+  } catch (std::bad_alloc &) {
```
clalancette, line 986:
```suggestion
  } catch (const std::bad_alloc &) {
```

clalancette (changes requested):
Two more minor changes, then we can run CI on this.  Thanks!

rmw_fastrtps_cpp/src/type_support_common.cpp
MiguelCompany: Sure. Done in 2e8c1e6ef959429b17412c7753b3bfbcf4936489

clalancette (approved):
Thanks!  Will run CI on this next.

christophfroehlich:
@clalancette would you mind backporting this to iron and humble as well? This would fix #733 then.

clalancette:
@Mergifyio backport humble iron

naorwaiss:
hi can i get some help 
i try to get topic at my ros2 iron from my jetson nano 20.04 that run ros2 foxy 
i get the same errror..

fujitatomoya:
@naorwaiss foxy is E.O.L https://docs.ros.org/en/rolling/Releases.html#list-of-distributions, can you try ROS Iron instead of `20.04 that run ros2 foxy`?

naorwaiss:
um i use jetson nano so i can run ubuntu 20.04 - this mean if i understand it .. that i can install only ros2 foxy because ros2 iron is available to ubuntu 22.04..

fujitatomoya:
I think that is the requirement for jetson nano platform? can you ask help for jetson community?

naorwaiss:
hi the installation of humble isnt work - can help me to make the bridge between ros2 iron and foxy ?

fujitatomoya:
> can help me to make the bridge between ros2 iron and foxy ?

cross-distribution communication is not officially supported.

PR #669, opened 2023-02-08
Rewrite how Topics are tracked in rmw_fastrtps_cpp.

clalancette:
Every DataReader and DataWriter in the system needs to have an associated Topic.  Ideally we would create these as one per publisher/subscriber, but Fast-DDS does not allow you to create multiple topics within the same DomainParticipant with the same topic name.  Instead, what we need to do is track ourselves whether a topic should be reused.

This was previously done with a combination of TopicHolder and cast_or_create_topic, but that had a couple of problems. First of all, a TopicHolder is really a SCOPE_EXIT by a different name.  Second, cast_or_create_topic worked, but really left the semantics of who owned a Topic up in the air.  If you created multiple topics with the same name, the first one would own it, and thus own the lifetime. The subsequent ones would reuse that.  But if you happened to remove the first one, then everything would probably crash.

Rewrite this whole thing to instead store the list of Topic pointers in the CustomParticipantInfo, which seems like a much more appropriate place for it.  When we go to add them, we first look if it already exists and if so, just increase the use_count.  If it doesn't exist, we create it with a use_count of 1.  On deletion, we do the opposite; decrease the use_count, and call delete_topic iff the use_count <= 0.  This data is wrapped in another class called UseCountTopic; we'll also be using this in the future to associate TopicListener with the Topic.

With this implementation, the semantics are much more clearly defined, and this should be able to handle deletion as well. It also happens to remove a bunch of code from the implementation, which is an added bonus.

Once we get this in (or something like it), I'll be able to redo the code in #654 to take advantage of this refactor.

Still a draft while I run CI on it.

@MiguelCompany I could especially use review from you.

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
@@ -16,8 +16,10 @@
 #define RMW_FASTRTPS_SHARED_CPP__CUSTOM_PARTICIPANT_INFO_HPP_
 
 #include <map>
+#include <memory>
 #include <mutex>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "fastdds/dds/domain/DomainParticipant.hpp"
@@ -53,10 +55,25 @@ enum class publishing_mode_t
   AUTO           // Use publishing mode set in XML file or Fast DDS default
 };
 
+typedef struct UseCountTopic
+{
+  eprosima::fastdds::dds::Topic * topic{nullptr};
+  size_t use_count{0};
+} UseCountTopic;
+
 typedef struct CustomParticipantInfo
 {
   eprosima::fastdds::dds::DomainParticipant * participant_{nullptr};
   ParticipantListener * listener_{nullptr};
+
+  std::mutex topic_name_to_topic_mutex_;
+  // As of 2023-02-07, Fast-DDS only allows one create_topic() with the same
+  // topic name per DomainParticipant.  Thus, we need to check if the topic
+  // was already created.  If it did, then we just increase the use_count
+  // that we are tracking, and return the existing topic.  If it
+  // didn't, then we create a new one and start tracking it.  Once all
+  // users of the topic are removed, we will delete the topic.
+  std::map<std::string, std::unique_ptr<UseCountTopic>> topic_name_to_topic_;
 
   eprosima::fastdds::dds::Publisher * publisher_{nullptr};
   eprosima::fastdds::dds::Subscriber * subscriber_{nullptr};
@@ -71,6 +88,54 @@ typedef struct CustomParticipantInfo
   // with the default configuration.
   bool leave_middleware_default_qos;
   publishing_mode_t publishing_mode;
+
+  eprosima::fastdds::dds::Topic * find_or_create_topic(
+    const std::string & topic_name,
+    const std::string & type_name,
+    const eprosima::fastdds::dds::TopicQos & topic_qos)
+  {
+    eprosima::fastdds::dds::Topic * topic = nullptr;
+
+    std::lock_guard<std::mutex> lck(topic_name_to_topic_mutex_);
+    std::map<std::string,
+      std::unique_ptr<UseCountTopic>>::const_iterator it = topic_name_to_topic_.find(topic_name);
+    if (it == topic_name_to_topic_.end()) {
+      // Not already in the map, we need to add it
+      topic = participant_->create_topic(topic_name, type_name, topic_qos);
+
+      auto uct = std::make_unique<UseCountTopic>();
+      uct->topic = topic;
+      uct->use_count = 1;
+
+      topic_name_to_topic_[topic_name] = std::move(uct);
+    } else {
+      // Already in the map, just increase the use count
+      it->second->use_count++;
+      topic = it->second->topic;
+    }
+
+    return topic;
+  }
+
+  void delete_topic(const eprosima::fastdds::dds::Topic * topic)
+  {
+    if (topic == nullptr) {
+      return;
+    }
+
+    std::lock_guard<std::mutex> lck(topic_name_to_topic_mutex_);
+    std::map<std::string, std::unique_ptr<UseCountTopic>>::const_iterator it =
+      topic_name_to_topic_.find(topic->get_name());
+
+    if (it != topic_name_to_topic_.end()) {
+      it->second->use_count--;
+      if (it->second->use_count <= 0) {
+        participant_->delete_topic(it->second->topic);
+
+        topic_name_to_topic_.erase(it);
+      }
+    }
+  }
 } CustomParticipantInfo;
 
 class ParticipantListener : public eprosima::fastdds::dds::DomainParticipantListener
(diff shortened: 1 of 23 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18030/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12555/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18674/)

methylDragon (approved):
I don't see any tests for CustomParticipantInfo that currently exist, so I don't think we need to test the new methods in this PR. Let's merge this PR to unblock the others!

LGTM on green CI (though I have one curiosity question)

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -71,6 +88,54 @@ typedef struct CustomParticipantInfo
...
+  {
+    if (topic == nullptr) {
+      return;
+    }
+
+    std::lock_guard<std::mutex> lck(topic_name_to_topic_mutex_);
+    std::map<std::string, std::unique_ptr<UseCountTopic>>::const_iterator it =
+      topic_name_to_topic_.find(topic->get_name());
+
+    if (it != topic_name_to_topic_.end()) {
```
methylDragon, line 130: Should we emit a warning if the topic was not found?

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -53,11 +55,26 @@ enum class publishing_mode_t
   AUTO           // Use publishing mode set in XML file or Fast DDS default
 };
 
+typedef struct UseCountTopic
+{
+  eprosima::fastdds::dds::Topic * topic{nullptr};
+  size_t use_count{0};
+} UseCountTopic;
+
 typedef struct CustomParticipantInfo
```
methylDragon, line 64: We should have a destructor that iterates through the `topic_name_to_topic_` map and calls `participant_->delete_topic()` on each, so we don't leak

methylDragon (changes requested):
I just noticed a missing destructor

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
```diff
@@ -53,11 +55,26 @@ enum class publishing_mode_t
...
   ParticipantListener * listener_{nullptr};
 
+  std::mutex topic_name_to_topic_mutex_;
+  // As of 2023-02-07, Fast-DDS only allows one create_topic() with the same
+  // topic name per DomainParticipant.  Thus, we need to check if the topic
+  // was already created.  If it did, then we just increase the use_count
+  // that we are tracking, and return the existing topic.  If it
+  // didn't, then we create a new one and start tracking it.  Once all
+  // users of the topic are removed, we will delete the topic.
+  std::map<std::string, std::unique_ptr<UseCountTopic>> topic_name_to_topic_;
```
methylDragon, line 76: We should have a destructor that iterates through the `topic_name_to_topic_` map and calls `participant_->delete_topic()` on each, so we don't leak

EduPonz:
Hi guys, I wanted to point out the existence of the [DomainParticipant::find_topic()](https://github.com/eProsima/Fast-DDS/blob/master/include/fastdds/dds/domain/DomainParticipant.hpp#L398) API. This API takes a topic name and returns a `Topic` corresponding with the name. If the topic had not yet been created, it waits for a max timeout duration until the topic is created by another thread. Important point is that all topics obtained by the means of `find_topic` must be deleted with `delete_topic`.

I can see that you already solve your need with some reference counting, but I wanted to leave this here in case it is of interest for you.

rmw_fastrtps_shared_cpp/src/utils.cpp
```diff
@@ -53,43 +53,6 @@ rmw_ret_t cast_error_dds_to_rmw(ReturnCode_t code)
...
-      topic_qos);
-
-    if (!topic_holder->topic) {
-      return false;
-    }
-
-    topic_holder->desc = topic_holder->topic;
-    topic_holder->should_be_deleted = true;
-  } else {
-    if (is_writer_topic) {
```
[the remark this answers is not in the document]
clalancette: Yeah, this is an excellent question, and one I had as well.  In short, I don't know the answer.  This is one of the reasons I wanted to rewrite this.

clalancette:
> Hi guys, I wanted to point out the existence of the [DomainParticipant::find_topic()](https://github.com/eProsima/Fast-DDS/blob/master/include/fastdds/dds/domain/DomainParticipant.hpp#L398) API. This API takes a topic name and returns a `Topic` corresponding with the name. If the topic had not yet been created, it waits for a max timeout duration until the topic is created by another thread. Important point is that all topics obtained by the means of `find_topic` must be deleted with `delete_topic`.

Thanks for the info.  I did briefly consider using `find_topic`, but the fact that I had to destroy it every time made me shy away from it.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
clalancette: Yeah, good point.  Added in 06ab8a9d1b305ea359560de2182106a53504d690.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp
clalancette: That's right, `destroy_participant` does this.  That is `destroy_participant` calls `remove_topic_and_type` on every type that is currently registered, and in turn `remove_topic_and_type` calls `CustomParticipantInfo::delete_topic` for every topic.  So we should be good here.

(I double-checked this by running under valgrind, and I didn't see any new leaks with this in place)

clalancette:
Another CI with the latest fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18031/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12556/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18675/)

rmw_fastrtps_shared_cpp/src/utils.cpp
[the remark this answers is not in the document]
methylDragon: :exploding_head: all's well that ends well :grimacing:

methylDragon (approved):

rmw_fastrtps_shared_cpp/src/utils.cpp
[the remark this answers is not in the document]
MiguelCompany: Just FYI, readers only need a TopicDescription to subscribe to, writers require it to be a Topic. In the case described here, the `topic_holder->desc` member (which is what the reader would use) was already assigned at the beginning of the method.

For writers, this was ensuring that the TopicDescription received as an input was a Topic.

MiguelCompany (approved):
LGTM. Such a nicely crafted refactor!

clalancette:
All right, CI is green and this is approved.  Going ahead and merging this one, thanks for looking!

PR #672, opened 2023-02-27
get valid subscription handle even if cft failed

iuhilnehc-ynos:


MiguelCompany (changes requested):
I suppose this also needs to be changed on rmw_fastrtps_dynamic_cpp

iuhilnehc-ynos:
`rmw_fastrtps_dynamic_cpp` does not currently support CFT, so there is no update for `rmw_fastrtps_dynamic_cpp`.

fujitatomoya:
@iuhilnehc-ynos can you rebase just in case and start CI for this?

fujitatomoya:
see https://github.com/ros2/demos/pull/579#issuecomment-3757278811

Issue #681, opened 2023-04-03
Receive repeated matched event


Barry-Xu-2018:
## Bug report

Get repeated matched event

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  -  source code
- Version or commit hash:
  - rolling
- DDS implementation:
  - fastdds (7123d353f4e63)
- Client library (if applicable):
  rclcpp

#### Steps to reproduce issue
```c++
#include <chrono>
#include <string>
#include <vector>

#include "rclcpp/rclcpp.hpp"

#include "std_msgs/msg/string.hpp"

using namespace std::chrono_literals;

class MatchedEventDetectNode : public rclcpp::Node
{
public:
  explicit MatchedEventDetectNode(
    const std::string & pub_topic_name)
  : Node("matched_event_detect_node")
  {
    rclcpp::PublisherOptions pub_options;
    pub_options.event_callbacks.matched_callback =
      [this](rclcpp::MatchedInfo & s) {
        RCLCPP_INFO(
          this->get_logger(),
          "Matched Event: %lu, %lu, %lu, %d", s.total_count, s.total_count_change, s.current_count, s.current_count_change);
      };

    pub_ = create_publisher<std_msgs::msg::String>(
      pub_topic_name, 10, pub_options);
  }

private:
  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr pub_;
  std::shared_ptr<std::promise<bool>> promise_;
};

class MultiSubNode : public rclcpp::Node
{
public:
  explicit MultiSubNode(const std::string & topic_name)
  : Node("multi_sub_node"),
    topic_name_(topic_name)
  {}

  rclcpp::Subscription<std_msgs::msg::String>::WeakPtr create_one_sub(void)
  {
    RCLCPP_INFO(this->get_logger(), "Create a new subscription.");
    auto sub = create_subscription<std_msgs::msg::String>(
      topic_name_,
      10,
      [](std_msgs::msg::String::ConstSharedPtr) {});

    subs_.emplace_back(sub);
    return sub;
  }

private:
  std::string topic_name_;
  std::vector<rclcpp::Subscription<std_msgs::msg::String>::SharedPtr> subs_;
};

int main(int argc, char ** argv)
{
  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
  rclcpp::init(argc, argv);

  std::string topic_name_for_detect_pub_matched_event = "pub_topic_matched_event_detect";

[...]

PR #682, opened 2023-04-03
Fix repeated matched event notification

Barry-Xu-2018:
Address #681

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
@@ -201,6 +201,17 @@ __rmw_wait(
         if (guard_condition.get_trigger_value()) {
           active = true;
           guard_condition.set_trigger_value(false);
+        } else {
+          switch (event->event_type) {
+            // Both a listener and a WaitSet for the same communication status, there are
+            // 2 triggers. Avoid repeated notification, filter it based trigger value.
+            case RMW_EVENT_SUBSCRIPTION_MATCHED:
+            case RMW_EVENT_PUBLICATION_MATCHED:
+              active = false;
+              break;
+            default:
+              break;
+          }
         }
       }
 


Barry-Xu-2018:
Although this correction is targeted towards matched events, I guess other events may also have the same issue.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -201,6 +201,17 @@ __rmw_wait(
         if (guard_condition.get_trigger_value()) {
           active = true;
           guard_condition.set_trigger_value(false);
+        } else {
+          switch (event->event_type) {
+            // Both a listener and a WaitSet for the same communication status, there are
+            // 2 triggers. Avoid repeated notification, filter it based trigger value.
```
fujitatomoya, line 207: Off topic from the original issue we are trying to solve, but if this should be applied to QoS events as well?

fujitatomoya: https://github.com/eProsima/Fast-DDS/issues/3395#issuecomment-1493706345 (that said that pure duplication of the same matched events.) would be also the solution, right? and it can address this behavior in the core before masking the event here.

fujitatomoya:
@MiguelCompany we would like you to review this PR, thanks in advance.

fujitatomoya:
Just quick heads-up, https://docs.ros.org/en/rolling/Releases/Release-Iron-Irwini.html#release-timeline this must be addressed in this week.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
Barry-Xu-2018: @fujitatomoya 

> Off topic from the original issue we are trying to solve, but if this should be applied to QoS events as well?

The processing logic of events in rmw_fastrtps is the same. So I guess QoS events has the same problem.
But need to further confirm it.   

> https://github.com/eProsima/Fast-DDS/issues/3395#issuecomment-1493706345 (that said that pure duplication of the same matched events.) would be also the solution, right? and it can address this behavior in the core before masking the event here.

I think this way is suitable for repeated event which has been returned from rmw_wait.   
Change here avoid repeated event returned from rmw_wait.

fujitatomoya:
@MiguelCompany @Barry-Xu-2018 

i left a couple of comments, https://github.com/ros2/rmw_fastrtps/pull/683. I was not clear that this can address the repeated issue precisely.

Barry-Xu-2018:
@MiguelCompany thanks for your explanation. I understand. 
I close this PR.

PR #685, opened 2023-04-07
Be more proactive about skipping work if possible

mjcarroll:
This makes `rmw_wait` more aggressive about how it skips work.

In the case that a condition is already met, this will skip attaching/detaching conditions from the underlying waitset and skip waiting.

In the case that a wait is required, it generally looks like this: 
image

and when a wait isn't needed: 

image

### DIFF
--- a/rmw_fastrtps_shared_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_shared_cpp/CMakeLists.txt
@@ -49,9 +49,6 @@ find_package(fastcdr REQUIRED CONFIG)
 find_package(fastrtps 2.3 REQUIRED CONFIG)
 find_package(FastRTPS 2.3 REQUIRED MODULE)
 
-find_package(Threads REQUIRED)
-find_package(tracy_vendor REQUIRED)
-
 find_package(rmw REQUIRED)
 
 add_library(rmw_fastrtps_shared_cpp
@@ -118,8 +115,6 @@ ament_target_dependencies(rmw_fastrtps_shared_cpp
   "rosidl_typesupport_introspection_c"
   "rosidl_typesupport_introspection_cpp"
   "tracetools"
-  "tracy_vendor"
-  "Threads"
 )
 
 # Causes the visibility macros to use dllexport rather than dllimport,
@@ -142,8 +137,6 @@ ament_export_dependencies(rmw_dds_common)
 ament_export_dependencies(rosidl_typesupport_introspection_c)
 ament_export_dependencies(rosidl_typesupport_introspection_cpp)
 ament_export_dependencies(tracetools)
-ament_export_dependencies(tracy_vendor)
-ament_export_dependencies(Threads)
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
--- a/rmw_fastrtps_shared_cpp/CMakeLists.txt
+++ b/rmw_fastrtps_shared_cpp/CMakeLists.txt
@@ -49,6 +49,9 @@ find_package(fastcdr REQUIRED CONFIG)
 find_package(fastrtps 2.3 REQUIRED CONFIG)
 find_package(FastRTPS 2.3 REQUIRED MODULE)
 
+find_package(Threads REQUIRED)
+find_package(tracy_vendor REQUIRED)
+
 find_package(rmw REQUIRED)
 
 add_library(rmw_fastrtps_shared_cpp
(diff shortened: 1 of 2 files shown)


mjcarroll:
Should skip wait has a relatively constant cost:

image


And postprocess

image

rmw_fastrtps_shared_cpp/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ find_package(fastcdr REQUIRED CONFIG)
 find_package(fastrtps 2.3 REQUIRED CONFIG)
 find_package(FastRTPS 2.3 REQUIRED MODULE)
 
+find_package(Threads REQUIRED)
+find_package(tracy_vendor REQUIRED)
```
clalancette, line 53: What's with the `tracy_vendor` here?  Do we need this to get this work in?

mjcarroll: No, there are some profiling points in this branch, I will remove for final merge if we determine this is a valid approach.

rmw_fastrtps_shared_cpp/src/rmw_wait.cpp
```diff
@@ -27,8 +27,85 @@
...
+      if (guard_condition->get_trigger_value())
+        return true;
+    }
+  }
+
+  if (events) {
+    for (size_t i = 0; i < events->event_count; ++i) {
+      auto event = static_cast<rmw_event_t *>(events->events[i]);
+      auto custom_event_info = static_cast<CustomEventInfo *>(event->data);
+      if (custom_event_info->get_listener()->get_statuscondition().get_trigger_value())
```
MiguelCompany, line 59: Note that events also have a guard condition:

```suggestion
      if (custom_event_info->get_listener()->get_statuscondition().get_trigger_value() ||
          custom_event_info->get_listener()->get_event_guard(event->event_type).get_trigger_value())
```

MiguelCompany (changes requested):

mjcarroll:
@MiguelCompany I am also looking at adding a method to the waitset to allow attaching/detaching multiple conditions.  I think this can save a bit of churn in locking/unlocking mutexes: https://github.com/mjcarroll/Fast-DDS/tree/mjcarroll/attach_multiple_conditions

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18491/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13039/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19216/)

clalancette (changes requested):
I've left just a few nits inline.  All of these are small enough that a "it compiles for me" comment is good enough so we don't have to restart CI.

I went through this with @mjcarroll in some detail.  In short, what this does is to change it so that we no longer call `wait_set->wait()` if there is already work waiting for us.  If that is the case, then we also don't need to attach/detach conditions to the wait_set if we already have work, and we can just go through and null out the entries that don't have work waiting.  If we do have work waiting for us, then we need to do the whole dance.

I will note that this is much faster in the best case, where the first guard condition we are waiting on is triggered.  In that case, we'll do much less work than before.  However, if there is no work for us to do, then this is actually somewhat slower than before since we need to iterate over the lists 3 times instead of 2.  I think this is an acceptable tradeoff for now, but it bears looking into further to see if we can reduce how expensive `get_first_untaken_info` is.

Anyway, with my nits fixed, I'm happy to approve this.  It looks like a net win for now.

mjcarroll:
> I will note that this is much faster in the best case, where the first guard condition we are waiting on is triggered. In that case, we'll do much less work than before. However, if there is no work for us to do, then this is actually somewhat slower than before since we need to iterate over the lists 3 times instead of 2. I think this is an acceptable tradeoff for now, but it bears looking into further to see if we can reduce how expensive get_first_untaken_info is.

Based on my testing with https://github.com/irobot-ros/ros2-performance, there is less tradeoff than you would think at first.  We do add a second set of iterations before the `wait` call, but in the case that we are actually waiting, that time would have been yielded to waiting on the waitset anyway.  The amount of work after the wait is over remains the same, so the event-to-callback latency is the same.

In the case that there is something already ready, we skip attaching the conditions, which are somewhat expensive (each requires a lock and unlock of a mutex) as well as skipping the `get_first_untaken_info` which is also expensive.

mjcarroll:
> All of these are small enough that a "it compiles for me" comment is good enough

It compiles for me. But for good measure, I'm throwing in a RHEL and Windows Debug build:

RHEL: [Build Status](https://ci.ros2.org/job/ci_linux-rhel/385/)
Windbg: [Build Status](https://ci.ros2.org/job/ci_windows/19221/)

clalancette (approved):

mjcarroll:
Alright, everything but Windows debug is green, so I'm calling this good.

PR #694, opened 2023-05-15
Add timestamp to rmw_publish tracepoint

cwecht:
Depends on https://github.com/ros2/ros2_tracing/pull/74
For some background: https://github.com/ros2/ros2_tracing/issues/73

Question: Has `write_w_timestamp` + `Time_t::now` any drawbacks if compared with using plain `write`?

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_publish.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_publish.cpp
@@ -15,6 +15,8 @@
 #include "fastcdr/Cdr.h"
 #include "fastcdr/FastBuffer.h"
 
+#include "fastdds/rtps/common/Time_t.h"
+
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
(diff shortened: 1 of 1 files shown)


MiguelCompany:
@cwecht Functionally, there are no drawbacks.
Semantically, there might be one issue if we want the tracepoint to be as close to the network/transport as possible (as stated in https://github.com/ros2/ros2_tracing/issues/73#issuecomment-1546273243)

Depending on the QoS settings, there might be a non-negligible blocking of the calling thread before the sample is added to the writer's history. Calling `write` will set `source_timestamp` to the time after that block

cwecht:
I think the proxmity of the tracepoint to the actual write call is not that important in this case, as we can not get any closer to the actual write operation right? For the tracing itself the timestamp passed here is not that important.
For a more precise tracing, adding tracepoints to the DDS implementation itself would be necessary, I suppose, right?

MiguelCompany:
> there is no differing behaviour to be expected, right?

Correct. No differing behaviour other than a different `source_timestamp` being notified on the subscription.

fujitatomoya (commented):
lgtm

fujitatomoya (approved):

PR #711, opened 2023-08-31
Improve node graph delivery by using a unique listening port

MiguelCompany:
Using a unique listening port for `ros_discovery_info` will make that topic have its own socket buffer, thus avoiding interference from other topics.

This could help mitigate service discovery issues when incoming network traffic on user topics is high.

### DIFF
--- a/rmw_fastrtps_cpp/src/init_rmw_context_impl.cpp
+++ b/rmw_fastrtps_cpp/src/init_rmw_context_impl.cpp
@@ -51,7 +51,8 @@ init_context_impl(
   // Avoid receiving graph updates from our own publication
   subscription_options.ignore_local_publications = true;
   // Improve graph discovery by using a unique listening port for its subscription
-  subscription_options.require_unique_network_flow_endpoints = RMW_UNIQUE_NETWORK_FLOW_ENDPOINTS_OPTIONALLY_REQUIRED;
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
creating DataReader for graph with optionally required unique network flow sounds good.

clalancette:
If I understand what this is doing correctly, this is going to create a unique network flow for all subscriptions in the context.  In turn, that means we are going to use more ports.  Do we know what the effect on the maximum number of subscriptions is going to be?  Or am I understanding this incorrectly?

MiguelCompany:
@clalancette This only affects the subscription to `ros_discovery_info` created when the context is initialized. All the other subscriptions will honor the options received in the call to `rmw_create_subscription`.

So, unless the user states it explicitly, only two ports will be used with this merged: one for `ros_discovery_info`, and one for everything else.

clalancette (approved):
Looks good to me with green CI.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20231/)

PR #718, opened 2023-10-04
add doc section for Zero Copy Data Sharing.

fujitatomoya:
CC: @MiguelCompany @EduPonz

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ However, `rmw_fastrtps` offers the possibility to further configure Fast DDS:
 * [Change publication mode](#change-publication-mode)
 * [Full QoS configuration](#full-qos-configuration)
 * [Change participant discovery options](#change-participant-discovery-options)
+* [Enable Zero Copy Data Sharing](#enable-zero-copy-data-sharing)
 
 ### Change publication mode
 
@@ -233,6 +234,39 @@ The following example configures Fast DDS to publish synchronously, and to have 
 ROS 2 allows controlling participant discovery with two environment variables: `ROS_AUTOMATIC_DISCOVERY_RANGE` and `ROS_STATIC_PEERS`.
 Full configuration of participant discovery can also be set with XML files; however, the ROS specific environment variables should be disabled to prevent them from interfering.
 Set `ROS_AUTOMATIC_DISCOVERY_RANGE` to the value `SYSTEM_DEFAULT` to disable both ROS specific environment variables.
+
+### Enable Zero Copy Data Sharing
+
+ROS 2 provides [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) that allows the user application to loan the message memory from the RMW implementation to eliminate the copy between the ROS 2 application and RMW implementation.
+And Fast DDS `rmw_fastrtps_cpp` provides [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) features to speed up the localhost communication.
+Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
+
+In default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is bounded type.
+
+To enable [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) with `rmw_fastrtps_cpp`, [Plain Data Data](https://en.wikipedia.org/wiki/Passive_data_structure) is the only requirement to `Iron Irwini` or later. For `Humble Hawksbill`, the following XML file needs to be applied to set Fast-DDS `data_sharing` is explicitly enabled. (see more details for https://github.com/ros2/rmw_fastrtps/pull/568)
+
+```xml
+<?xml version="1.0" encoding="UTF-8" ?>
+<profiles xmlns="http://www.eprosima.com/XMLSchemas/fastRTPS_Profiles">
+
+  <!-- Default publisher profile -->
+  <data_writer profile_name="default publisher profile" is_default_profile="true">
+    <qos>
+      <data_sharing>
+        <kind>AUTOMATIC</kind>
+      </data_sharing>
+    </qos>
+  </data_writer>
+
+  <data_reader profile_name="default subscription profile" is_default_profile="true">
+    <qos>
+      <data_sharing>
+        <kind>AUTOMATIC</kind>
+      </data_sharing>
+    </qos>
+  </data_reader>
+</profiles>
+```
 
 ## Quality Declaration files
 
--- a/README.md
+++ b/README.md
@@ -241,9 +241,10 @@ ROS 2 provides [Loaned Messages](https://design.ros2.org/articles/zero_copy.html
 And Fast DDS `rmw_fastrtps_cpp` provides [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) features to speed up the localhost communication.
 Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
 
-By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type(fixed sized data object).
+By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type (a fixed sized data object).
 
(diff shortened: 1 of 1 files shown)


fujitatomoya:
address https://github.com/ros2/rmw_fastrtps/issues/714

@Zard-C it would be really nice if you can review this. thanks in advance.

fujitatomoya (commented):
This includes only doc fix, so i do not think we need to run CI.

README.md
```diff
@@ -234,6 +235,39 @@ ROS 2 allows controlling participant discovery with two environment variables: `
...
+Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
+
+In default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is bounded type.
+
+To enable [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) with `rmw_fastrtps_cpp`, [Plain Data Data](https://en.wikipedia.org/wiki/Passive_data_structure) is the only requirement to `Iron Irwini` or later. For `Humble Hawksbill`, the following XML file needs to be applied to set Fast-DDS `data_sharing` is explicitly enabled. (see more details for https://github.com/ros2/rmw_fastrtps/pull/568)
```
Zard-C, line 246: It seems there's a small typo here, you might want `[Plain Old Data]`
Apart from this, the document is well-written! Thanks for your effort! 👍

README.md
```diff
@@ -234,6 +235,39 @@ ROS 2 allows controlling participant discovery with two environment variables: `
...
 
+### Enable Zero Copy Data Sharing
+
+ROS 2 provides [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) that allows the user application to loan the message memory from the RMW implementation to eliminate the copy between the ROS 2 application and RMW implementation.
+And Fast DDS `rmw_fastrtps_cpp` provides [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) features to speed up the localhost communication.
+Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
+
+In default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is bounded type.
```
clalancette, line 244:
```suggestion
By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type.
```

It might also be worthwhile describing what a "bounded type" means.

README.md
clalancette, line 246:
```suggestion
To enable [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) with `rmw_fastrtps_cpp`, [Plain Old Data](https://en.wikipedia.org/wiki/Passive_data_structure) is the only requirement to `Iron Irwini` or later. For `Humble Hawksbill`, the following XML file needs to be applied to set Fast-DDS `data_sharing` is explicitly enabled. (see more details for https://github.com/ros2/rmw_fastrtps/pull/568)
```

clalancette (changes requested):

README.md
fujitatomoya: addressed with https://github.com/ros2/rmw_fastrtps/pull/718/commits/e39c11c401a64da532661dfee111d1e5711a5ee3

README.md
fujitatomoya: https://github.com/ros2/rmw_fastrtps/pull/718/commits/e39c11c401a64da532661dfee111d1e5711a5ee3

fujitatomoya:
@Zard-C @clalancette could you take an another look? thank you very much for the review!

README.md
```diff
@@ -234,6 +235,39 @@ ROS 2 allows controlling participant discovery with two environment variables: `
...
 
+### Enable Zero Copy Data Sharing
+
+ROS 2 provides [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) that allows the user application to loan the message memory from the RMW implementation to eliminate the copy between the ROS 2 application and RMW implementation.
+And Fast DDS `rmw_fastrtps_cpp` provides [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) features to speed up the localhost communication.
+Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
+
+By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type(fixed sized data object).
```
clalancette, line 244:
```suggestion
By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type (a fixed sized data object).
```

README.md
```diff
@@ -234,6 +235,39 @@ ROS 2 allows controlling participant discovery with two environment variables: `
...
+
+To enable [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) with `rmw_fastrtps_cpp`, [Plain Old Data](https://en.wikipedia.org/wiki/Passive_data_structure) is the only requirement to `Iron Irwini` or later. For `Humble Hawksbill`, the following XML file needs to be applied to set Fast-DDS `data_sharing` is explicitly enabled. (see more details for https://github.com/ros2/rmw_fastrtps/pull/568)
```
clalancette, line 246:
```suggestion
To enable [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) with `rmw_fastrtps_cpp`, [Plain Old Data](https://en.wikipedia.org/wiki/Passive_data_structure) is the only requirement to `Iron Irwini` or later.
For `Humble Hawksbill`, the following XML file needs to be applied to explicitly enable Fast-DDS `data_sharing` (see more details in https://github.com/ros2/rmw_fastrtps/pull/568).
```

clalancette (changes requested):
Sorry, a couple more typos then I think this is good.

README.md
fujitatomoya: https://github.com/ros2/rmw_fastrtps/pull/718/commits/4ce89f3afa5da9097a99b6055b3215397d544665

fujitatomoya:
@clalancette no worries, thanks for the review. appreciate it. the comments are addressed.

Zard-C (approved):

README.md
```diff
@@ -234,6 +235,40 @@ ROS 2 allows controlling participant discovery with two environment variables: `
...
 
+### Enable Zero Copy Data Sharing
+
+ROS 2 provides [Loaned Messages](https://design.ros2.org/articles/zero_copy.html) that allows the user application to loan the message memory from the RMW implementation to eliminate the copy between the ROS 2 application and RMW implementation.
+And Fast DDS `rmw_fastrtps_cpp` provides [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) features to speed up the localhost communication.
+Taking advantage of these features all together, it provides significant performance improvement to ROS 2 application.
+
+By default, `rmw_fastrtps_cpp` tries to use [Shared Memory Transport](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html) and [Data-sharing delivery](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html) for localhost communication along with network communication if the message data type is a bounded type (a fixed sized data object).
```
EduPonz, line 244: It seems that I'm late, but this is not entirely true. By default `rmw_fastrtps` sets data-sharing to `OFF` [here](https://github.com/ros2/rmw_fastrtps/blob/rolling/rmw_fastrtps_cpp/src/publisher.cpp#L255) and [here](https://github.com/ros2/rmw_fastrtps/blob/rolling/rmw_fastrtps_dynamic_cpp/src/publisher.cpp#L261). To enable it, users need to:

1. Set the XML
2. `export RMW_FASTRTPS_USE_QOS_FROM_XML=1`.

I'll make a PR to adjust the text

fujitatomoya: @EduPonz really? i did the test to see the files under `/dev/shm` with current rolling. you mean the data did not go through the shared memory mapped files?

fujitatomoya: So i did some experimental tests on local environment. i think i was looking at the files for `shared memory transport` but `data sharing`. as you mentioned, if i enable `data sharing` via xml, i can see the file like `fast_datasharing_01.0f.60.db.97.c9.17.1a.00.00.00.00_0.0.14.3`.

thanks for catching up on this, really helped!

EduPonz: Yeah, it is a rather common misunderstanding, and at honest one as that, as the terms can very easily get mixed up. In short, Fast DDS has two shared memory based delivery mechanism:

1. One is what we call a full compliant transport, which is similar to a socket based transport in the sense that you have ports and encapsulation of the data that is copied from the data writers local history cache into the transport buffers. This can be used without any kind of type restrictions. Out of the box, Fast DDS participants use one instance of this transport to communicate with other participants that are detected to be on the same machine.
1. Then we have what we called data sharing delivery mechanism. It is still based on memory mapped files (same as the SHM), but in here, under some conditions, the data writer stores its history cache in the mmap file directly, and the data readers read the data directly from there, thus removing all the transport related copies. This approach however comes with data type limitations, mainly that data is required to be bounded, as the nmap segment needs to be able to store all the writer's history cache. Mind that this is not the case with SHM transport, as the serialized data can be fragmented if the payload is larger than the SHM buffer. If both a data writer and reader in the same host are data sharing compatible, meaning that their data types are bounded and they have the qos set to `on` or `auto`, then they will communicate over data sharing instead of the SHM transport.

Because of the previous, Zero Copy can only be achieved by combining sample loaning with data sharing, but not with SHM transport.

fujitatomoya:
@mergifyio backport iron humble

PR #726, opened 2023-10-12
Data sharing is required to enable loaned interface on subscription.

fujitatomoya:
this is draft PR to try to address,

- https://github.com/ros-planning/navigation2/pull/3863#discussion_r1356135469
- https://github.com/ros-planning/navigation2/pull/3876
- https://github.com/ros-planning/navigation2/pull/3875

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -44,6 +44,8 @@
 namespace rmw_fastrtps_shared_cpp
 {
 
+using DataSharingKind = eprosima::fastdds::dds::DataSharingKind;
+
 void
 _assign_message_info(
   const char * identifier,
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -562,7 +564,8 @@ __init_subscription_for_loans(
 {
   auto info = static_cast<CustomSubscriberInfo *>(subscription->data);
   const auto & qos = info->data_reader_->get_qos();
-  subscription->can_loan_messages = info->type_support_->is_plain();
+  bool has_data_sharing = DataSharingKind::OFF != qos.data_sharing().kind();
+  subscription->can_loan_messages = has_data_sharing && info->type_support_->is_plain();
```
fujitatomoya, line 568: I intentionally did not apply the same change to the publisher, since that is clear that user need to use `LoanedMessage` class explicitly, and it also provide performance improvement for intra and UDP transport w/o data sharing.

wjwwood: I think this misses the point that you may want to use messages loaned from the middleware even when you're not doing inter-process zero-copy (so called "DataSharing" with Fast-DDS).

You would also like to use loaned messages to avoid copies, which is why the rclcpp executor uses it by default.

wjwwood: * you may want *your subscription* to use messages loaned from the middleware

fujitatomoya: yeah that was original intention to enable loaned message w/o `DataSharing` in Fast-DDS. It still provides the performance improvement for subscription, after https://github.com/ros2/rcl/pull/1110 is merged, probably we do not need this one.

fujitatomoya:
I will go ahead to close this one, see https://github.com/ros-planning/navigation2/pull/3863#discussion_r1358087687 for more details.

and,

- https://github.com/ros2/rcl/pull/1110
- https://github.com/ros2/rclcpp/pull/2335

will do the work.

PR #746, opened 2024-02-29
Support Fast CDR v2

MiguelCompany:
This makes the necessary changes to allow building with Fast CDR v2

Part of ros2/ros2#1530

### DIFF
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -21,7 +21,7 @@
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
   <build_depend version_gte="2">fastcdr</build_depend>
-  <build_depend>fastrtps</build_depend>
+  <build_depend version_gte="2.13">fastrtps</build_depend>
   <build_depend>fastrtps_cmake_module</build_depend>
   <build_depend>rcpputils</build_depend>
   <build_depend>rcutils</build_depend>
@@ -37,7 +37,7 @@
   <build_depend>tracetools</build_depend>
 
   <build_export_depend version_gte="2">fastcdr</build_export_depend>
-  <build_export_depend>fastrtps</build_export_depend>
+  <build_export_depend version_gte="2.13">fastrtps</build_export_depend>
   <build_export_depend>fastrtps_cmake_module</build_export_depend>
   <build_export_depend>rcutils</build_export_depend>
   <build_export_depend>rmw_dds_common</build_export_depend>
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -20,7 +20,7 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
-  <build_depend>fastcdr</build_depend>
+  <build_depend version_gte="2">fastcdr</build_depend>
   <build_depend>fastrtps</build_depend>
   <build_depend>fastrtps_cmake_module</build_depend>
   <build_depend>rcpputils</build_depend>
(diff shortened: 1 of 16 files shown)


rmw_fastrtps_cpp/package.xml
```diff
@@ -20,8 +20,8 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
-  <build_depend>fastcdr</build_depend>
-  <build_depend>fastrtps</build_depend>
+  <build_depend version_gte="2">fastcdr</build_depend>
+  <build_depend version_gte="2.13">fastrtps</build_depend>
```
clalancette, line 24: Let's not add these constraints in here.  I know it is technically more correct, but we don't do this anywhere else.

(same goes for all of the uses below)

clalancette (changes requested):
Other than my one comment, this looks good to me.

rmw_fastrtps_cpp/package.xml
MiguelCompany: This is something we mentioned within the Production WG while discussing [this proposal](https://github.com/bpwilcox/ros2-production-working-group/issues/11)

Not doing it anywhere else doesn't mean it should not be done.
Although it seems that [rosdep will not enforce version checks](https://github.com/ros-infrastructure/rosdep/issues/325),  bloom will put the version requirements in the rules to build debian packages.

That said, I will add a commit removing version restrictions from all `package.xml` files, and add them in a follow-up PR so we don't block merging this one

MiguelCompany: 3d2f5f5

clalancette: > Although it seems that [rosdep will not enforce version checks](https://github.com/ros-infrastructure/rosdep/issues/325), bloom will put the version requirements in the rules to build debian packages.

It's true that bloom will put those version checks in, but there isn't really a scenario where it would ever be needed, especially in the core and in Rolling.  Releasing this package will cause all of the downstream packages to rebuild (which is most of the ROS ecosystem).  If any of that fails to build, then we won't have a viable distribution to "sync".  If it all succeeds in building, then we will sync the whole thing out together.  In that sense, it is "atomic" with the rest of the packages (at least up to `desktop`).  You can definitely construct scenarios where someone cherry-picks certain packages and tries to install them together, but anyone doing that on Rolling gets to keep all of the pieces, in my opinion.

Anyway, I agree with that this is a wider discussion we should have, so removing it from this PR is the correct thing for now.

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
```diff
@@ -78,11 +78,12 @@ bool TypeSupport::serialize(
         eprosima::fastcdr::FastBuffer fastbuffer(  // Object that manages the raw buffer
           reinterpret_cast<char *>(payload->data), payload->max_size);
         eprosima::fastcdr::Cdr ser(  // Object that serializes the data
-          fastbuffer, eprosima::fastcdr::Cdr::DEFAULT_ENDIAN, eprosima::fastcdr::Cdr::DDS_CDR);
+          fastbuffer, eprosima::fastcdr::Cdr::DEFAULT_ENDIAN, eprosima::fastcdr::CdrVersion::XCDRv1);
```
clalancette, line 81: CI is complaining that this line is too long (needs to be <= 100 characters): https://ci.ros2.org/job/ci_linux-aarch64/14925/testReport/junit/(root)/projectroot/cpplint/

clalancette (changes requested):
One linter change needed here.

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
MiguelCompany: 9f0296e

clalancette (approved):

PR #755, opened 2024-05-02
Altering liveliness lease_duration

CursedRock17:
This pull request is supposed to solve an issue noted in [this `RCUTILS` pr](https://github.com/ros2/rcutils/pull/460#issuecomment-2080348553) in which a [qos line](https://github.com/ros2/rmw_fastrtps/blob/c48406a5f87fd7ebe5a576fbf627b6c9efcc786e/rmw_fastrtps_shared_cpp/src/qos.cpp#L145) undergoes an incorrect conversion. The original intent was to pass an `int64_t` and sacrifice some accuracy, but I don't believe it's possible due to our limited [constructors for `Duration_t`](https://github.com/eProsima/Fast-DDS/blob/eaa23dbdea6ff2fa8259452be368ac8bcb273cc5/src/cpp/rtps/common/Time_t.cpp#L67-#L86). Instead, I converted to a nanosecond value, then operated on that value by making it 2/3s the original. Then, converting that value to a sec (double) which should fill the `long double` argument. We shouldn't need to operate on the sec (double), after having converted, because while our new version will have less accuracy, `RCUTILS_NS_TO_S` is undergoing changes to become more accurate.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -141,8 +141,8 @@ bool fill_entity_qos_from_profile(
     // Docs suggest setting no higher than 0.7 * lease_duration, choosing 2/3 to give safe buffer.
     // See doc at https://github.com/eProsima/Fast-RTPS/blob/
     //   a8691a40be6b8460b01edde36ad8563170a3a35a/include/fastrtps/qos/QosPolicies.h#L223-L232
-    double period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2.0 / 3.0;
-    double period_in_s = RCUTILS_NS_TO_S(period_in_ns);
+    int64_t period_in_ns = entity_qos.liveliness().lease_duration.to_ns();
+    double period_in_s = RCUITLS_NS_TO_S(period_in_s) * 2.0 / 3.0
     entity_qos.liveliness().announcement_period = eprosima::fastrtps::Duration_t(period_in_s);
   }
 
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -142,7 +142,7 @@ bool fill_entity_qos_from_profile(
     // See doc at https://github.com/eProsima/Fast-RTPS/blob/
     //   a8691a40be6b8460b01edde36ad8563170a3a35a/include/fastrtps/qos/QosPolicies.h#L223-L232
     int64_t period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2 / 3;
-    double period_in_s = RCUITLS_NS_TO_S(period_in_s);
+    double period_in_s = RCUTILS_NS_TO_S(period_in_ns);
     entity_qos.liveliness().announcement_period = eprosima::fastrtps::Duration_t(period_in_s);
   }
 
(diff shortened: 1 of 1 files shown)


rmw_fastrtps_shared_cpp/src/qos.cpp
```diff
@@ -141,8 +141,8 @@ bool fill_entity_qos_from_profile(
     // Docs suggest setting no higher than 0.7 * lease_duration, choosing 2/3 to give safe buffer.
     // See doc at https://github.com/eProsima/Fast-RTPS/blob/
     //   a8691a40be6b8460b01edde36ad8563170a3a35a/include/fastrtps/qos/QosPolicies.h#L223-L232
-    double period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2.0 / 3.0;
-    double period_in_s = RCUTILS_NS_TO_S(period_in_ns);
+    int64_t period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2 / 3;
+    double period_in_s = RCUITLS_NS_TO_S(period_in_s);
```
clalancette, line 145:
```suggestion
    double period_in_s = RCUTILS_NS_TO_S(period_in_ns);
```

clalancette (changes requested):

fujitatomoya (commented):
nit: title would be `Altering liveliness lease_duration`? `announcement period` sounds really different qos setting as domain participant, and that is not what we are changing here. (https://fast-dds.docs.eprosima.com/en/latest/fastdds/discovery/general_disc_settings.html#discovery-lease-announ)

sloretz:
@clalancette friendly ping :waffle:

rmw_fastrtps_shared_cpp/src/qos.cpp
```diff
@@ -141,7 +141,7 @@ bool fill_entity_qos_from_profile(
     // Docs suggest setting no higher than 0.7 * lease_duration, choosing 2/3 to give safe buffer.
     // See doc at https://github.com/eProsima/Fast-RTPS/blob/
     //   a8691a40be6b8460b01edde36ad8563170a3a35a/include/fastrtps/qos/QosPolicies.h#L223-L232
-    double period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2.0 / 3.0;
+    int64_t period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2 / 3;
     double period_in_s = RCUTILS_NS_TO_S(period_in_ns);
```
clalancette, line 145: Just coming back to look at this again, and I'm realizing that these two lines together are wrong, at least, with the current implementation of RCUTILS_NS_TO_S.

The problem is in switching `period_in_ns` to an `int64_t`.  Because `RCUTILS_NS_TO_S` currently uses longs in the divisor, then everything is an integer.  So if you have less than 1 second (1000000000 ns), then we use integer division here and always end up with zero.  That's likely why `period_in_ns` was a double to begin with, since everything will be promoted to a double.

If we put in https://github.com/ros2/rcutils/pull/460 along with this one, then I guess that particular issue goes away.  But this is making me much more nervous about https://github.com/ros2/rcutils/pull/460 in general.  While this macro isn't used very heavily, I think that will change the semantics enough that it is concerning.  I'll leave a comment over there.

As far as this PR goes, I actually think that what is here is fine, at least for now.  So I'm actually going to suggest that we close this PR out.

clalancette (changes requested):

fujitatomoya:
@CursedRock17 https://github.com/ros2/rmw_fastrtps/pull/755#discussion_r1670636641 makes sense to me. do we close this one out?

CursedRock17:
Apologies, I wanted to go through solutions in `rcutils` before I came back here to clear up this pull request. Yeah we can go ahead and close this out. Though perhaps, way in the future, we may have to revisit if the MACRO behavior ends up changing.

PR #776, opened 2024-09-06
Changes to build against Fast DDS 3.0

MiguelCompany:
This PR brings in the necessary changes to build against Fast DDS 3.0.x

It has been reviewed by the eProsima Team in eProsima#32

Part of ros2/ros2#1598


fujitatomoya (commented):
@MiguelCompany DCO is missing, can you check that?

rmw_fastrtps_cpp/CMakeLists.txt
```diff
@@ -38,10 +38,8 @@ find_package(rmw_dds_common REQUIRED)
 find_package(rmw_fastrtps_shared_cpp REQUIRED)
 find_package(tracetools REQUIRED)
 
-find_package(fastrtps_cmake_module REQUIRED)
```
clalancette, line 41: Do we need to keep `fastrtps_cmake_module` around at all, or should we look at removing that package entirely?

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
```diff
@@ -234,263 +250,234 @@ _create_type_name(const MembersType * members)
   std::string message_namespace(members->message_namespace_);
   std::string message_name(members->message_name_);
   if (!message_namespace.empty()) {
+    // Find and replace C namespace separator with C++, in case this is using C typesupport
+    message_namespace = rcpputils::find_and_replace(message_namespace, "__", "::");
```
clalancette, line 254: Can you explain a little bit more of why this is needed now?

clalancette (changes requested):
I've left two minor questions.

Otherwise, this tentatively looks good to me, but I'd like to see full CI, including Windows Debug and a clang run.

rmw_fastrtps_cpp/CMakeLists.txt
richiware: This might be removed.

rmw_fastrtps_shared_cpp/src/TypeSupport_impl.cpp
richiware: This is needed to ensure same type coming from Python or C++ is the same type at DDS X-Types level. Without this a Python application will register a different DDS-XTypes TypeObject and maybe won't communicate with a C++ application.

This rename is already done in other places like [MessageTypeSupport_impl](https://github.com/ros2/rmw_fastrtps/blob/rolling/rmw_fastrtps_dynamic_cpp/src/MessageTypeSupport_impl.hpp#L47).

rmw_fastrtps_cpp/CMakeLists.txt
MiguelCompany: See https://github.com/ros2/rosidl_typesupport_fastrtps/pull/120 for this

bshantam97:
@MiguelCompany Hello, when will the FastDDS3 RMW Changes be merged with ROS2?

clalancette (changes requested):
This generally looks good to me, but since we are removing `fastrtps_cmake_module` we should also update the QUALITY_DECLARATION.md files to remove any mention of it.

clalancette:
CI for this is in https://github.com/ros2/ros2/pull/1598#issuecomment-2607298187

PR #781, opened 2024-10-01
Use client's reader guid for service introspection event gid.

fujitatomoya:
part of https://github.com/ros2/rmw/issues/357

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/rmw_get_gid_for_client.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_get_gid_for_client.cpp
@@ -38,7 +38,10 @@ __rmw_get_gid_for_client(
   RMW_CHECK_ARGUMENT_FOR_NULL(gid, RMW_RET_INVALID_ARGUMENT);
 
   const auto * info = static_cast<const CustomClientInfo *>(client->data);
-  copy_from_fastrtps_guid_to_byte_array(info->writer_guid_, gid->data);
+  // Use client's reader guid instead of writer guid for service event,
+  // because service server uses client's reader guid for the event.
+  // Service event message requires gid must be unique during transaction.
+  copy_from_fastrtps_guid_to_byte_array(info->reader_guid_, gid->data);
   gid->implementation_identifier = identifier;
   return RMW_RET_OK;
 }


fujitatomoya (commented):
@MiguelCompany what do you think about this change?

I am almost sure that using client's reader guid instead of writer guid here in the client for the service introspection event, will not affect any other behavior in rmw_fastrtps.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21928/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16195/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1442/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22654/)

MiguelCompany:
This is ok... if only using `rmw_fastrtps_cpp`.

Should we consider the case of the server and the client using different vendors?
What would the expectation be in that case?

I am ok with merging this, and addressing the cross-vendor case in a follow-up, but I would like @clalancette opinion first.

fujitatomoya:
atm, i think that is okay not to consider cross-vendor communication and current code does not seem to care about that either. (https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Different-Middleware-Vendors.html#cross-vendor-communication) i understand we necessarily do not want to break that compatibility with this., but rosbag2 service problem

> without this uniqueness capability during a single request-reponse pair, it cannot tell which request is mapped to which response, so that rosbag2 cannot know which message needs to play.

is much bigger problem happening right now...

fujitatomoya:
@clalancette @wjwwood what do you think?

MiguelCompany (approved):

MiguelCompany:
> i think that is okay not to consider cross-vendor communication

In fact, we would first need to have service interoperability between all the RMWs, which is currently not the case.

Approved, and CI green, so I feel free to merge

clalancette:
Did we run CI on RHEL Debug for this?  That was the case that caused all of the problems before.

clalancette:
It looks like not, here is a run.  If this fails, we'll have to revert:

* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1454/)

clalancette:
All right, all looks happy with this.  We'll see what happens overnight.

PR #799, opened 2025-01-21
Fix some overwritten errors in rmw_fastrtps.

clalancette:
There are 2 different kinds of errors fixed in here:

1.  When trying to get the topic QoS settings, the function get_topic_qos() always sets an error when it fails.  Thus the higher level functions do not also need to set an error.

2.  When rmw_serialize() calls rmw_serialized_message_resize(), it may fail deep in rcutils calls.  Those are very low-level error messages, and not that useful, so instead we clear those out and report an error at this rmw_fastrtps layer.

### DIFF
--- a/rmw_fastrtps_dynamic_cpp/src/rmw_serialize.cpp
+++ b/rmw_fastrtps_dynamic_cpp/src/rmw_serialize.cpp
@@ -45,6 +45,7 @@ rmw_serialize(
   auto data_length = tss->getEstimatedSerializedSize(ros_message, ts->data);
   if (serialized_message->buffer_capacity < data_length) {
     if (rmw_serialized_message_resize(serialized_message, data_length) != RMW_RET_OK) {
+      rmw_reset_error();
       RMW_SET_ERROR_MSG("unable to dynamically resize serialized message");
       type_registry.return_message_type_support(ts);
       return RMW_RET_ERROR;
--- a/rmw_fastrtps_cpp/src/publisher.cpp
+++ b/rmw_fastrtps_cpp/src/publisher.cpp
@@ -217,7 +217,7 @@ rmw_fastrtps_cpp::create_publisher(
   // Create and register Topic
   eprosima::fastdds::dds::TopicQos topic_qos = dds_participant->get_default_topic_qos();
   if (!get_topic_qos(*qos_policies, topic_qos)) {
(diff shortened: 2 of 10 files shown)


clalancette:
Pulls: https://github.com/ros2/rmw_fastrtps/pull/799
Gist: https://gist.githubusercontent.com/clalancette/94df29ee7118d04d0de56b5f3fb1daed/raw/a0261734cfe49f98105399bc0f0f37da735e43bd/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15077
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22398/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16656/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1927/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23188/)

rmw_fastrtps_dynamic_cpp/src/rmw_serialize.cpp
```diff
@@ -45,6 +45,7 @@ rmw_serialize(
   auto data_length = tss->getEstimatedSerializedSize(ros_message, ts->data);
   if (serialized_message->buffer_capacity < data_length) {
     if (rmw_serialized_message_resize(serialized_message, data_length) != RMW_RET_OK) {
+      rmw_reset_error();
       RMW_SET_ERROR_MSG("unable to dynamically resize serialized message");
```
fujitatomoya, line 49: or maybe we can do the following?

```suggestion
      RMW_SET_ERROR_MSG_AND_APPEND_PREV_ERROR("unable to dynamically resize serialized message");
```

clalancette: In this case, I'm not sure it is a good idea.  `rmw_serialized_message_resize` is defined to be the same as `rcutils_uint8_array_resize`.  At this level, getting an error message that a `uint8_array` resize failed is just confusing, in my opinion.  So in this case, I think it is better to clear out the previous error and set the new one instead.

fujitatomoya (approved):
lgtm with green CI

PR #853, opened 2025-11-19
Remove vscode settings.json file.

clalancette:
## Description

It should not be checked in.

### Is this user-facing behavior change?

No.

### Did you use Generative AI?

No.

### Additional Information

PR #855, opened 2025-11-19
Add VS Code folder to .gitignore to prevent #853

SuperJappie08:
## Description

Add the `.vscode`, `build`, `install` and `log` folders to `.gitignore` to prevent it from ever being committed.

Prevents #853 from reoccurring.

No

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+# Workspace folders
+build/
+install/
+log/
+
+# VS Code
+.vscode/


ahcorde (approved):

.gitignore
```diff
@@ -0,0 +1,7 @@
+# Workspace folders
+build/
+install/
+log/
+
```
fujitatomoya, line 5: @SuperJappie08 

why do we need this? this is not the workspace, so even after build these folders are not gonna be created in source repo?

fujitatomoya: the same question goes to,

https://github.com/ros2/rmw_cyclonedds/pull/548
https://github.com/ros2/rmw_implementation/pull/269
https://github.com/ros2/rmw/pull/410
https://github.com/ros2/rmw_fastrtps/pull/853

SuperJappie08: @fujitatomoya 
It's not a workspace, but running colcon in the wrong folder (without a specific plugins) creates those folders.

It's a sort of beginner mistake, which most contributors to the RMW layers will never make.

But I was adding a `.gitignore` anyway, so I thought it couldn't hurt.

fujitatomoya: i do not really like this approach, because

- for the consistency, we need to add `.gitignore` for each maintained repository under ros2 organization. (maintenance burden)
- `build`, `install` and `log` directories are accidently created by user. i do not think those are responsibility for `.gitignore` in each repository. if that is, other tools could also create some folders in accident.
- repository could have `install` and `log` directory on purpose by the name (could be naming conflicts)

with that, i would suggest to add `.vscode/` to user's `.gitignore` in the ROS 2 development documentation? what do you think? those PRs are already merged, but i really do not think we need to add `.gitignore` to all repositories...

MiguelCompany: I agree with this. Adding `.vscode/` is arbitrary, since other users might be using other tools (for instance, I develop on Windows and use Visual Studio 2022)

peci1: I agree adding this doesn't make sense. Some of the accidents might be better handled by guiding users in the beginner docs/tutorials to create a global gitignore depending on the tools they use.

peci1: I've submitted a docs issue for this: https://github.com/ros2/ros2_documentation/issues/6040 .

SuperJappie08: I was unaware of the existence of userprofile gitignores, which for usersetup-related files is indeed a much nicer solution.

If there is a need for it, feel free to revert this commit.

@ahcorde, you might be interested in some of this information.

ahcorde: This is a complex situation, because some of the ROS 2 core packages already have this `.gitignore` files. 

We can bring this topic to the PMC and reach an agreement to be consistent across all ROS 2 core packages and document which is the right thing to do.

peci1: That would be great. I think a part of this problem could also be watched by CI. E.g. a CI script might want to check for some well known directories like those in this gitignore, and if it finds them, fail the CI pipeline.

MiguelCompany: That would be a nice pre-commit hook

PR #866, opened 2026-03-05
Add thread-local TypeSupport cache to rmw_serialize/deserialize

mjcarroll:
This PR adds a thread_local MessageTypeSupport_cpp and TypeSupportCache to rmw_serialize and rmw_deserialize. It eliminates per-call hash-map lookup and mutex overhead, significantly improving performance for small and high-frequency messages.

Generated with the help of Claude Sonnet 4.6


MiguelCompany (changes requested):
Very nice optimization!

Sadly, DCO needs to be fixed

fujitatomoya (approved):

PR #867, opened 2026-03-16
Add support for rosidl::Buffer-aware per-endpoint pub/sub

nvcyc:
## Description

This pull request adds full `rosidl::Buffer` support to `rmw_fastrtps_cpp`, enabling per-endpoint DDS DataWriters/DataReaders for zero-copy buffer transport between compatible backends. Backend discovery uses DDS `user_data` QoS to advertise and detect supported backends.

This pull request modifies the following packages:

### `rmw_fastrtps_shared_cpp` -- shared types and QoS

- **`custom_participant_info.hpp`**: Added `BufferDiscoveryCallback` type and `buffer_discovery_cb_` field for participant-level buffer discovery.
- **`custom_publisher_info.hpp`**: Added `BufferPublisherEndpoint` struct and buffer-related fields (`is_buffer_aware_`, `backend_aux_info_`, `local_endpoint_info_`, `buffer_mutex_`, `buffer_endpoints_`, `pending_buffer_endpoints_`, `buffer_alive_flag_`, `participant_`, `dds_publisher_`).
- **`custom_subscriber_info.hpp`**: Added `BufferSubscriptionEndpoint` struct (with `owns_topic` flag for single-process topic reuse) and buffer-related fields. Added `RMWSubscriptionEvent::notify_buffer_data_available()`.
- **`qos.hpp`/`qos.cpp`**: Added optional `buffer_backends` parameter to QoS functions; added `encode_buffer_backends_for_user_data()` / `parse_buffer_backends_from_user_data()` (format: `BUFBE:name=aux;name2=aux2`).
- **`rmw_wait.cpp`**: Extended WaitSet to also check `buffer_data_guard_`; wakes subscription when buffer data is available.

### `rmw_fastrtps_cpp` -- buffer-aware per-endpoint pub/sub

- **Backend lifecycle**: Calls `initialize_buffer_backends()` / `shutdown_buffer_backends()` during `rmw_init`/`rmw_shutdown`.
- **`buffer_endpoint_registry`**: Per-process singleton mapping topic names to discovery callbacks; publishers register subscriber-discovery callbacks and vice versa. Maintains known endpoints so late-registering callbacks see already-discovered peers.
- **Publisher creation**: Detects buffer-aware types via `callbacks->has_buffer_fields`; encodes backend names in DDS `user_data` with `"cpu"` always present. Registers subscriber-discovery callbacks that create per-subscriber DataWriters on unique topics (`{topic}/_buf/{pub_gid}_{sub_gid}`) with reliable QoS, synchronous publish mode, and data sharing off. Discovery callbacks use `buffer_alive_flag_` to prevent use-after-free.
- **Subscription creation**: Parses `acceptable_buffer_backends` (`NULL`/empty/`"cpu"`: CPU-only; `"any"`: all; specific names: filtered). Creates per-publisher DataReaders with `BufferDataReaderListener` that triggers `buffer_data_guard_`. **Topic reuse**: In single-process scenarios, uses `lookup_topicdescription()` to reuse existing topics (`owns_topic = false`).
- **Buffer-aware publish** (`rmw_publish.cpp`): Iterates per-subscriber endpoints, calls `cdr_serialize_with_endpoint()`. Conditionally publishes to the main DataWriter only when non-buffer-aware subscribers exist, avoiding unnecessary CPU conversion.
- **Buffer-aware take** (`rmw_take.cpp`): Iterates per-publisher DataReaders, takes raw CDR bytes and calls `cdr_deserialize_with_endpoint()`. Falls back to main DataReader when no buffer endpoint data is available. **Guard condition re-arm**: Re-sets `buffer_data_guard_` when remaining data exists.
- **Lock-ordering safety**: Both publisher and subscriber discovery use a three-phase pattern matching `rmw_zenoh_cpp`: (1) acquire `buffer_mutex_` to check duplicates and mark pending; (2) call DDS create/delete APIs without holding the lock; (3) re-acquire to store endpoints. Destroy functions move endpoints to local lists under lock and clean up DDS resources after releasing.

This pull request does not change existing `rmw_fastrtps` behavior for standard messages. For messages with `uint8[]` fields, the per-endpoint transport is transparent -- publishers and subscribers share backend info via DDS `user_data`, and CPU fallback ensures correctness when backends are incompatible.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with creating an initial prototype version of the changes contained in this PR.

### Additional Information

This PR is part of the broader ROS 2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).

### DIFF
--- a/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
+++ b/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
@@ -31,12 +31,6 @@ bool gid_equal(const rmw_gid_t & a, const rmw_gid_t & b)
 }
 }  // namespace
 
-BufferEndpointRegistry & BufferEndpointRegistry::get_instance()
-{
-  static BufferEndpointRegistry instance;
-  return instance;
-}
-
 void BufferEndpointRegistry::register_subscriber_discovery_callback(
   const std::string & topic_name,
   const rmw_gid_t & publisher_gid,
--- a/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
+++ b/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
@@ -15,21 +15,13 @@
 #include "buffer_endpoint_registry.hpp"
 
 #include <algorithm>
-#include <cstring>
 #include <vector>
 
 #include "rcutils/logging_macros.h"
+#include "rmw/rmw.h"
 
 namespace rmw_fastrtps_cpp
 {
-
-namespace
-{
-bool gid_equal(const rmw_gid_t & a, const rmw_gid_t & b)
-{
-  return std::memcmp(a.data, b.data, RMW_GID_STORAGE_SIZE) == 0;
-}
-}  // namespace
 
 void BufferEndpointRegistry::register_subscriber_discovery_callback(
   const std::string & topic_name,
@@ -89,7 +81,17 @@ void BufferEndpointRegistry::unregister_callbacks(const rmw_gid_t & gid)
         entries.erase(
           std::remove_if(
             entries.begin(), entries.end(),
-            [&gid](const CallbackEntry & e) {return gid_equal(e.registrant_gid, gid);}),
+            [&gid](const CallbackEntry & e) {
+              bool equal = false;
+              rmw_ret_t ret = rmw_compare_gids_equal(&e.registrant_gid, &gid, &equal);
+              if (RMW_RET_OK != ret) {
+                RCUTILS_LOG_ERROR_NAMED(
+                  "rmw_fastrtps_cpp",
+                  "BufferEndpointRegistry: rmw_compare_gids_equal failed in unregister_callbacks");
+                return false;
+              }
+              return equal;
+            }),
           entries.end());
       }
     };
@@ -105,7 +107,15 @@ void BufferEndpointRegistry::notify_subscriber_discovered(const BufferEndpointIn
     std::lock_guard<std::mutex> lock(mutex_);
 
     for (const auto & existing : known_subscribers_) {
-      if (gid_equal(existing.gid, info.gid)) {
+      bool equal = false;
+      rmw_ret_t ret = rmw_compare_gids_equal(&existing.gid, &info.gid, &equal);
+      if (RMW_RET_OK != ret) {
+        RCUTILS_LOG_ERROR_NAMED(
+          "rmw_fastrtps_cpp",
+          "BufferEndpointRegistry: rmw_compare_gids_equal failed in notify_subscriber_discovered");
+        continue;
+      }
+      if (equal) {
         RCUTILS_LOG_DEBUG_NAMED(
           "rmw_fastrtps_cpp",
           "BufferEndpointRegistry: subscriber on '%s' already known, skipping",
@@ -144,7 +154,15 @@ void BufferEndpointRegistry::notify_publisher_discovered(const BufferEndpointInf
     std::lock_guard<std::mutex> lock(mutex_);
 
     for (const auto & existing : known_publishers_) {
-      if (gid_equal(existing.gid, info.gid)) {
+      bool equal = false;
+      rmw_ret_t ret = rmw_compare_gids_equal(&existing.gid, &info.gid, &equal);
+      if (RMW_RET_OK != ret) {
+        RCUTILS_LOG_ERROR_NAMED(
+          "rmw_fastrtps_cpp",
+          "BufferEndpointRegistry: rmw_compare_gids_equal failed in notify_publisher_discovered");
+        continue;
+      }
+      if (equal) {
         RCUTILS_LOG_DEBUG_NAMED(
           "rmw_fastrtps_cpp",
           "BufferEndpointRegistry: publisher on '%s' already known, skipping",
--- a/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
+++ b/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
@@ -95,7 +95,7 @@ void BufferEndpointRegistry::unregister_callbacks(const rmw_gid_t & gid)
         entries.erase(
           std::remove_if(
             entries.begin(), entries.end(),
-            [&gid](const CallbackEntry & e) { return gid_equal(e.registrant_gid, gid); }),
+            [&gid](const CallbackEntry & e) {return gid_equal(e.registrant_gid, gid);}),
           entries.end());
       }
     };
--- a/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
+++ b/rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
@@ -1,0 +1,185 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "buffer_endpoint_registry.hpp"
+
+#include <algorithm>
+#include <cstring>
+#include <vector>
+
+#include "rcutils/logging_macros.h"
+
+namespace rmw_fastrtps_cpp
+{
+
+namespace
+{
+bool gid_equal(const rmw_gid_t & a, const rmw_gid_t & b)
+{
+  return std::memcmp(a.data, b.data, RMW_GID_STORAGE_SIZE) == 0;
+}
+}  // namespace
+
+BufferEndpointRegistry & BufferEndpointRegistry::get_instance()
+{
+  static BufferEndpointRegistry instance;
+  return instance;
+}
+
+void BufferEndpointRegistry::register_subscriber_discovery_callback(
+  const std::string & topic_name,
+  const rmw_gid_t & publisher_gid,
+  BufferEndpointDiscoveryCallback callback)
+{
+  // Collect matching known endpoints while holding the lock, then fire outside.
+  std::vector<BufferEndpointInfo> to_fire;
+  BufferEndpointDiscoveryCallback cb_copy;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    subscriber_callbacks_[topic_name].push_back({publisher_gid, callback});
+    cb_copy = subscriber_callbacks_[topic_name].back().callback;
+
+    for (const auto & info : known_subscribers_) {
+      if (info.topic_name == topic_name) {
+        to_fire.push_back(info);
+      }
+    }
+  }
+
+  for (const auto & info : to_fire) {
+    cb_copy(info);
+  }
+}
+
+void BufferEndpointRegistry::register_publisher_discovery_callback(
+  const std::string & topic_name,
+  const rmw_gid_t & subscriber_gid,
+  BufferEndpointDiscoveryCallback callback)
+{
+  std::vector<BufferEndpointInfo> to_fire;
+  BufferEndpointDiscoveryCallback cb_copy;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    publisher_callbacks_[topic_name].push_back({subscriber_gid, callback});
+    cb_copy = publisher_callbacks_[topic_name].back().callback;
+
+    for (const auto & info : known_publishers_) {
+      if (info.topic_name == topic_name) {
+        to_fire.push_back(info);
+      }
+    }
+  }
+
+  for (const auto & info : to_fire) {
+    cb_copy(info);
+  }
+}
+
+void BufferEndpointRegistry::unregister_callbacks(const rmw_gid_t & gid)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  auto remove_from = [&gid](auto & map) {
+      for (auto & [topic, entries] : map) {
+        entries.erase(
+          std::remove_if(
+            entries.begin(), entries.end(),
+            [&gid](const CallbackEntry & e) { return gid_equal(e.registrant_gid, gid); }),
+          entries.end());
+      }
+    };
+
+  remove_from(subscriber_callbacks_);
+  remove_from(publisher_callbacks_);
+}
+
+void BufferEndpointRegistry::notify_subscriber_discovered(const BufferEndpointInfo & info)
+{
+  std::vector<BufferEndpointDiscoveryCallback> callbacks;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    for (const auto & existing : known_subscribers_) {
+      if (gid_equal(existing.gid, info.gid)) {
+        RCUTILS_LOG_DEBUG_NAMED(
+          "rmw_fastrtps_cpp",
+          "BufferEndpointRegistry: subscriber on '%s' already known, skipping",
+          info.topic_name.c_str());
+        return;
+      }
+    }
+    known_subscribers_.push_back(info);
+
+    auto it = subscriber_callbacks_.find(info.topic_name);
+    if (it == subscriber_callbacks_.end()) {
+      RCUTILS_LOG_DEBUG_NAMED(
+        "rmw_fastrtps_cpp",
+        "BufferEndpointRegistry: no publisher callbacks registered for topic '%s'",
+        info.topic_name.c_str());
+      return;
+    }
+    for (const auto & entry : it->second) {
+      callbacks.push_back(entry.callback);
+    }
+  }
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_fastrtps_cpp",
+    "BufferEndpointRegistry: firing %zu publisher callback(s) for discovered subscriber on '%s'",
+    callbacks.size(), info.topic_name.c_str());
+  for (const auto & cb : callbacks) {
+    cb(info);
+  }
+}
+
+void BufferEndpointRegistry::notify_publisher_discovered(const BufferEndpointInfo & info)
+{
+  std::vector<BufferEndpointDiscoveryCallback> callbacks;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    for (const auto & existing : known_publishers_) {
+      if (gid_equal(existing.gid, info.gid)) {
+        RCUTILS_LOG_DEBUG_NAMED(
+          "rmw_fastrtps_cpp",
+          "BufferEndpointRegistry: publisher on '%s' already known, skipping",
+          info.topic_name.c_str());
+        return;
+      }
+    }
+    known_publishers_.push_back(info);
+
+    auto it = publisher_callbacks_.find(info.topic_name);
+    if (it == publisher_callbacks_.end()) {
+      RCUTILS_LOG_DEBUG_NAMED(
+        "rmw_fastrtps_cpp",
+        "BufferEndpointRegistry: no subscriber callbacks registered for topic '%s'",
+        info.topic_name.c_str());
+      return;
+    }
+    for (const auto & entry : it->second) {
+      callbacks.push_back(entry.callback);
+    }
+  }
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_fastrtps_cpp",
+    "BufferEndpointRegistry: firing %zu subscriber callback(s) for discovered publisher on '%s'",
+    callbacks.size(), info.topic_name.c_str());
+  for (const auto & cb : callbacks) {
+    cb(info);
+  }
+}
+
+}  // namespace rmw_fastrtps_cpp
--- a/rmw_fastrtps_cpp/src/rmw_init.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_init.cpp
@@ -45,7 +45,7 @@
 #include "rmw_fastrtps_cpp/publisher.hpp"
 #include "rmw_fastrtps_cpp/subscription.hpp"
 
-#include "rosidl_buffer_backend_registry/buffer_backend_loader.hpp"
+#include "buffer_backend_loader.hpp"
 
 extern "C"
 {
@@ -119,7 +119,7 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
   }
 
   try {
-    rosidl_buffer_backend_registry::initialize_buffer_backends();
+    rmw_fastrtps_cpp::initialize_buffer_backends();
   } catch (const std::exception &) {
     // Non-fatal: buffer backends are optional.
   }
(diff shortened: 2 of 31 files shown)


rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+namespace rmw_fastrtps_cpp
+{
+
+namespace
+{
+bool gid_equal(const rmw_gid_t & a, const rmw_gid_t & b)
+{
+  return std::memcmp(a.data, b.data, RMW_GID_STORAGE_SIZE) == 0;
+}
```
hidmic: @nvcyc there's [`rmw_compare_gids_equal()`](https://github.com/ros2/rmw_fastrtps/blob/0fda8bbaa0d075222c8cdc5812c1c1fa94b1d83a/rmw_fastrtps_cpp/src/rmw_compare_gids_equal.cpp#L26) already

rmw_fastrtps_cpp/src/rmw_publish.cpp
```diff
@@ -12,17 +12,101 @@
...
+    return has_non_buffer_subscribers;
+  }
+
+  eprosima::fastdds::dds::Time_t stamp;
+  eprosima::fastdds::dds::Time_t::now(stamp);
+  TRACETOOLS_TRACEPOINT(rmw_publish, publisher, ros_message, stamp.to_ns());
+
+  for (const auto & endpoint : info->buffer_endpoints_) {
+    uint32_t serialized_size = callbacks->get_serialized_size(ros_message);
+    size_t buffer_size = serialized_size + 4096;
```
hidmic: @nvcyc mind to explain those extra 4KB?

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +132,120 @@ rmw_create_publisher(
...
+        auto gid_to_hex = [](const rmw_gid_t & gid, size_t bytes = 8) -> std::string {
+            static const char hex_chars[] = "0123456789abcdef";
+            std::string result;
+            result.reserve(bytes * 2);
+            for (size_t i = 0; i < bytes && i < RMW_GID_STORAGE_SIZE; ++i) {
+              result += hex_chars[(gid.data[i] >> 4) & 0xF];
+              result += hex_chars[gid.data[i] & 0xF];
+            }
+            return result;
+          };
```
hidmic: @nvcyc nit: considering making this a free function.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +132,120 @@ rmw_create_publisher(
...
+        endpoint->backend_aux_info = sub_info.backend_aux_info;
+
+        endpoint->subscriber_endpoint_info = rmw_get_zero_initialized_topic_endpoint_info();
+        endpoint->subscriber_endpoint_info.endpoint_type = RMW_ENDPOINT_SUBSCRIPTION;
+        std::memcpy(
+          endpoint->subscriber_endpoint_info.endpoint_gid,
+          sub_info.gid.data, RMW_GID_STORAGE_SIZE);
+
+        eprosima::fastdds::dds::TopicQos topic_qos =
+          info->participant_->get_default_topic_qos();
```
hidmic: @nvcyc meta: we should notify the user somehow about the actual QoS profile used for buffer aware pub/sub, or fallback to CPU only, or something.

rmw_fastrtps_cpp/src/rmw_take.cpp
```diff
@@ -14,13 +14,112 @@
...
+    rmw_fastrtps_shared_cpp::SerializedData data;
+    data.type = rmw_fastrtps_shared_cpp::FASTDDS_SERIALIZED_DATA_TYPE_CDR_BUFFER;
+    data.data = &receive_buffer;
+    data.impl = nullptr;
+
+    eprosima::fastdds::dds::StackAllocatedSequence<void *, 1> data_values;
+    const_cast<void **>(data_values.buffer())[0] = &data;
+    eprosima::fastdds::dds::SampleInfoSeq info_seq{1};
+
+    auto ret_code = endpoint->data_reader->take(data_values, info_seq, 1);
```
hidmic: @nvcyc meta: how do we enforce that we don't starve some endpoints? This implicitly prioritizes endpoints that were discovered earlier than others.

rmw_fastrtps_cpp/src/subscription.cpp
```diff
@@ -655,9 +660,82 @@ __create_subscription(
...
+  if (has_buffer_fields) {
+    auto all_backends =
+      rosidl_buffer_backend_registry::BufferBackendRegistry::get_instance().get_all_aux_info();
+
+    // Parse acceptable_buffer_backends option (comma-separated) to filter
+    std::vector<std::string> requested_list;
+    if (subscription_options->acceptable_buffer_backends &&
+      strlen(subscription_options->acceptable_buffer_backends) > 0)
+    {
+      std::string acceptable(subscription_options->acceptable_buffer_backends);
```
hidmic: @nvcyc there's [`rcpputils::split()`](https://github.com/ros2/rcpputils/blob/rolling/include/rcpputils/split.hpp) for this.

rmw_fastrtps_shared_cpp/src/qos.cpp
```diff
@@ -150,13 +151,75 @@ bool fill_entity_qos_from_profile(
...
+static const size_t BUFFER_BACKEND_SENTINEL_LEN = 6;
+
+std::string
+encode_buffer_backends_for_user_data(
+  const std::unordered_map<std::string, std::string> & backends)
+{
+  if (backends.empty()) {
+    return {};
+  }
+  std::string result = BUFFER_BACKEND_SENTINEL;
```
hidmic: @nvcyc fyi: there's `std::stringstream`.

hidmic (commented):
First pass. Another PR that needs many 👀 .

rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
```diff
@@ -0,0 +1,185 @@
...
+    for (const auto & entry : it->second) {
+      callbacks.push_back(entry.callback);
+    }
+  }
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_fastrtps_cpp",
+    "BufferEndpointRegistry: firing %zu publisher callback(s) for discovered subscriber on '%s'",
+    callbacks.size(), info.topic_name.c_str());
+  for (const auto & cb : callbacks) {
```
mjcarroll, line 145: This looks safe from deadlock risk.

rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
```diff
@@ -0,0 +1,185 @@
...
+    for (const auto & entry : it->second) {
+      callbacks.push_back(entry.callback);
+    }
+  }
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_fastrtps_cpp",
+    "BufferEndpointRegistry: firing %zu subscriber callback(s) for discovered publisher on '%s'",
+    callbacks.size(), info.topic_name.c_str());
+  for (const auto & cb : callbacks) {
```
mjcarroll, line 192: This one looks safe too.

rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
hidmic: This still stands.

rmw_fastrtps_cpp/src/rmw_publish.cpp
hidmic: Still needs an explanation.

hidmic (commented):
Looks like things are in flight. @MiguelCompany @richiware we need your 👀  here.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +132,73 @@ rmw_create_publisher(
...
+            {
+              return;
+            }
+          }
+          for (const auto & p : state->pending) {
+            if (p.unique_topic == unique_topic) {
+              return;
+            }
+          }
+
```
nvcyc, line 228: Should call `rosidl_buffer_backend_registry::BufferBackendRegistry::get_instance().call notify_endpoint_discovered()` around here.

rmw_fastrtps_cpp/src/rmw_subscription.cpp
```diff
@@ -123,6 +130,80 @@ rmw_create_subscription(
...
+            {
+              return;
+            }
+          }
+          for (const auto & p : state->pending) {
+            if (p.unique_topic == unique_topic) {
+              return;
+            }
+          }
+
```
nvcyc, line 168: Should call `rosidl_buffer_backend_registry::BufferBackendRegistry::get_instance().call notify_endpoint_discovered()` around here.

rmw_fastrtps_cpp/src/rmw_publish.cpp
```diff
@@ -12,17 +12,171 @@
...
+  create_pending_buffer_writers(info);
+
+  auto callbacks = static_cast<const message_type_support_callbacks_t *>(info->type_support_impl_);
+  auto & state = *info->buffer_state_;
+
+  std::lock_guard<std::mutex> lock(state.mutex);
+
+  size_t total_matched = info->publisher_event_->subscription_count();
+  size_t buffer_count = state.endpoints.size() + state.pending.size();
+  bool has_non_buffer_subscribers = (total_matched > buffer_count);
```
nvcyc: Return directly if `has_non_buffer_subscribers` is non-zero as a fallback when legacy RMWs appear in any peer nodes. That is, every subscriber will get CPU-based messages from their main data reader if there is one or more subscriber that is based on legacy (older version) RMWs that do not support buffer backend protocol (and only work with main data reader/writer).

rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
hidmic: Still applies.

rmw_fastrtps_cpp/src/rmw_init.cpp
```diff
@@ -110,12 +114,36 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
...
+    context->impl->buffer_serialization_context = nullptr;
+    RMW_SET_ERROR_MSG("failed to allocate buffer endpoint registry");
+    return RMW_RET_BAD_ALLOC;
+  }
+  context->impl->buffer_endpoint_registry = buffer_endpoint_registry;
+
+  try {
+    rmw_fastrtps_cpp::initialize_buffer_backends(*buffer_context);
+  } catch (const std::exception &) {
+    // Non-fatal: buffer backends are optional.
```
hidmic, line 145: @nvcyc nit: should this log something instead of swallowing all errors?

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +137,102 @@ rmw_create_publisher(
...
+          auto gid_to_hex = [](const rmw_gid_t & gid, size_t bytes = 8) -> std::string {
+            static const char hex_chars[] = "0123456789abcdef";
+            std::string result;
+            result.reserve(bytes * 2);
+            for (size_t i = 0; i < bytes && i < RMW_GID_STORAGE_SIZE; ++i) {
+              result += hex_chars[(gid.data[i] >> 4) & 0xF];
+              result += hex_chars[gid.data[i] & 0xF];
+            }
+            return result;
+          };
```
hidmic: @nvcyc nit: this could be a free function (isn't it already?)

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +137,102 @@ rmw_create_publisher(
...
+          std::string unique_topic = base_topic +
+          "/_buf/" + pub_hex + "_" + sub_hex;
+
+          {
+            std::lock_guard<std::mutex> lock(state->mutex);
+            if (!state->alive.load()) {
+              return;
+            }
+            for (const auto & ep : state->endpoints) {
+              if (std::memcmp(ep->target_subscriber_gid.data, sub_info.gid.data,
```
hidmic: @nvcyc nit: GID comparison? I thought we had some function for this above.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
```diff
@@ -126,6 +137,102 @@ rmw_create_publisher(
...
+                existing_endpoints.push_back(ep->subscriber_endpoint_info);
+              }
+              for (const auto & p : state->pending) {
+                existing_endpoints.push_back(p.subscriber_endpoint_info);
+              }
+
+              std::unordered_map<std::string,
+              std::vector<std::set<uint32_t>>> backend_endpoint_groups;
+              (void)rosidl_buffer_backend_registry::notify_endpoint_discovered(
+              backend_context->backend_instances,
```
hidmic: @nvcyc nit^2: formatting is peculiar. Is `uncrustify` happy with this indentation?

rmw_fastrtps_cpp/src/rmw_take.cpp
hidmic: What was the conclusion of this? Is this safe?

rmw_fastrtps_cpp/src/subscription.cpp
hidmic: Still applies.

hidmic (commented):
Second (or third?) pass. What about transient local buffer aware pub/sub? What is the expectation there?

rmw_fastrtps_cpp/src/buffer_endpoint_registry.cpp
yuanknv: Thanks for the catch, replaced `gid_equal()` calls with existing `rmw_compare_gids_equal() `

rmw_fastrtps_cpp/src/subscription.cpp
yuanknv: Done. Replaced the manual tokenizer with `rcpputils::split`

rmw_fastrtps_cpp/src/rmw_init.cpp
yuanknv: Thanks for the catch, added `RCUTILS_LOG_INFO_NAMED ` with the exception message

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Fixed, replaced `memcpy `with `rmw_compare_gids_equal`

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Thanks for the catch. Re-ran `uncrustify `to fix the formatting. All clean now

rmw_fastrtps_shared_cpp/src/qos.cpp
yuanknv: Done. Replaced manual string concatenation with `std::ostringstream` and the manual tokenizer with `std::istringstream`

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Thanks for the catch. Fixed the per-subscriber/per-publisher buffer endpoint TopicQos to inherit from the main topic instead of using bare defaults. Since they are always identical to the main pub/sub, added DEBUG-level logging to avoid flooding the user's console. Also added DEBUG log messages for buffer-aware DataReader/DataWriter creation, including QoS details (reliability, durability, history) and the mode used (cpu-only vs accelerated).

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Thanks for the catch. Extracted `gid_to_hex ` into a free function inguid_utils.hpp.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Done, use the extracted `gid_to_hex ` function.

rmw_fastrtps_cpp/src/rmw_take.cpp
yuanknv: Thanks for pointing it out, it's a valid concern. Per discussion with @nvcyc , we replaced the per-publisher DataReaders on the subscriber side with a single shared DataReader. All buffer-aware publishers now write to a common per-subscriber topic (/_buf/<sub_hex>). This avoids the starvation issue and reduces the number of DataReaders used.

rmw_fastrtps_cpp/src/rmw_publish.cpp
yuanknv: Thanks for the catch. Replaced the `+4096` with `+4`. The 4 bytes account for the CDR encapsulation header  which `get_serialized_size `does not include.

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: added

rmw_fastrtps_cpp/src/rmw_subscription.cpp
yuanknv: added

ahcorde:
@mjcarroll @hidmic @MiguelCompany do you mind to take a look ? rmw freeze it's next monday (6th of April)

rmw_fastrtps_cpp/src/rmw_publisher.cpp
hidmic: Awesome, but I think that the caveats around QoS for hardware accelerated data paths should be glaring. 

As far as I can see, buffer backend implementation, acceptable backends per P2P connection, and the presence of legacy RMWs will determine to what extent QoS imposed constraints are fulfilled (particularly durability). E.g. a transient local publisher will lose its properties unless all subscribers accept CPU buffers and there are no legacy RMWs laying around, and here I'm assuming the buffer backend can handle a descriptor that was published 2 hours ago.

Consider disabling buffer aware transport for QoS profiles we can't handle at the moment. Log a big warning when that's the case. That'll make sure what's working will continue working and users will know they are leaving performance on the table.

hidmic (approved):
Other than the QoS comment, I think this is correct (but don't quote me on it). Pending green CI, DCO, yada yada. Also, I think system tests are super important. We may want to add a few more if the uint8[] type is not well represented there.

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/guid_utils.hpp
```diff
@@ -86,6 +89,19 @@ struct hash_fastdds_guid
   }
 };
 
+inline std::string
+gid_to_hex(const rmw_gid_t & gid, size_t bytes = 8)
```
MiguelCompany: rmw_gid_t is 16 bytes long, and there is a constant with that value.

```suggestion
gid_to_hex(const rmw_gid_t & gid, size_t bytes = RMW_GID_STORAGE_SIZE)
```

MiguelCompany (changes requested):

nvcyc:
Started by user [CY Chen](https://ci.ros2.org/user/nvcyc)
Running as SYSTEM
Building on the built-in node in workspace /var/lib/jenkins/workspace/ci_launcher
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28506/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21369/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8364/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27370/)

MiguelCompany (approved):

nvcyc:
Triggered manually with custom ros2.repos that contain all PRs: https://raw.githubusercontent.com/nvcyc/ros2/refs/heads/rolling-native-buffer-prs/ros2.repos

Started by user [CY Chen](https://ci.ros2.org/user/nvcyc)
Running as SYSTEM
Building on the built-in node in workspace /var/lib/jenkins/workspace/ci_launcher
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21383/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8393/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27388/)

rmw_fastrtps_cpp/src/rmw_publisher.cpp
yuanknv: Thanks for the feedback—that's a great point. I agree with the approach you're suggesting. The plan is to add a QoS compatibility check at publisher/subscriber creation time in the RMW layer. When the QoS profile includes settings we can't properly support (starting with transient local durability), we'll disable the buffer-aware path for that entity and fall back to standard DDS transport, logging a clear warning so users know. 

Since we're approaching code freeze, we will track this and implement the feature in a follow-up PR

nvcyc:
Pulls: ros2/rosidl#943, ros2/rosidl_typesupport_fastrtps#144, ros2/rosidl_runtime_py#39, ros2/rosidl_python#250, ros2/rosidl_core#14, ros2/rmw_fastrtps#867
Gist: https://gist.githubusercontent.com/nvcyc/2b81ce4a59e71ec57d7d52f520fab590/raw/a28b02c5a17f2e12f08af57ffeb8479e1e399f98/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rosidl_buffer rosidl_buffer_py rosidl_runtime_c rosidl_typesupport_introspection_c rosidl_typesupport_introspection_tests rosidl_generator_cpp rosidl_runtime_cpp rosidl_typesupport_introspection_cpp rosidl_typesupport_fastrtps_c rosidl_typesupport_fastrtps_cpp rosidl_runtime_py rosidl_generator_py rosidl_core_runtime rmw_fastrtps_cpp rmw_fastrtps_shared_cpp
TEST args:  --packages-above rosidl_buffer rosidl_buffer_py rosidl_runtime_c rosidl_typesupport_introspection_c rosidl_typesupport_introspection_tests rosidl_generator_cpp rosidl_runtime_cpp rosidl_typesupport_introspection_cpp rosidl_typesupport_fastrtps_c rosidl_typesupport_fastrtps_cpp rosidl_runtime_py rosidl_generator_py rosidl_core_runtime rmw_fastrtps_cpp rmw_fastrtps_shared_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18835
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28542/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21400/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8400/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27397/)

PR #877, opened 2026-04-14
Avoid intermediate memcpy in _take_serialized_message

facontidavide:
The previous implementation of **_take_serialized_message** filled a local `eprosima::fastcdr::FastBuffer` during the `DataReader::take()` call, then memcpy'd the payload from that `FastBuffer` into the user's `rmw_serialized_message_t`. For large serialized messages (e.g. uncompressed image frames) the second memcpy is the dominant cost of the take path.

Add a new FASTDDS_SERIALIZED_DATA_TYPE_RMW_SERIALIZED_MESSAGE enum variant to SerializedDataType, and teach `TypeSupport::deserialize` to write the CDR payload straight into the user's rmw_serialized_message_t, resizing it via rmw_serialized_message_resize when the capacity is insufficient.

Measured on ROS 2 Humble (the equivalent bool-flag backport), FastDDS SHM transport, 1920x1080 BGR8 images (6.22 MB / frame), 80 measured frames, perf stat on the subscriber process, mean of 3 runs:

     before (stock)      169 ms
     after  (this patch) 138 ms   (-18%)

Wire format and external behaviour are unchanged. The existing FASTDDS_SERIALIZED_DATA_TYPE_CDR_BUFFER path is preserved for other callers (publisher serialize, services).

### DIFF
--- a/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/TypeSupport.hpp
+++ b/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/TypeSupport.hpp
@@ -41,7 +41,12 @@ enum SerializedDataType
 {
   FASTDDS_SERIALIZED_DATA_TYPE_CDR_BUFFER,
   FASTDDS_SERIALIZED_DATA_TYPE_DYNAMIC_MESSAGE,
-  FASTDDS_SERIALIZED_DATA_TYPE_ROS_MESSAGE
+  FASTDDS_SERIALIZED_DATA_TYPE_ROS_MESSAGE,
+  // `data` points to a `rmw_serialized_message_t` (aka
+  // `rcutils_uint8_array_t`). The CDR payload is written directly into that
+  // buffer on deserialize, avoiding an intermediate FastBuffer and the
+  // follow-up memcpy in `_take_serialized_message`.
+  FASTDDS_SERIALIZED_DATA_TYPE_RMW_SERIALIZED_MESSAGE
 };
 
 // Publishers write method will receive a pointer to this struct
--- a/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
+++ b/rmw_fastrtps_shared_cpp/src/rmw_take.cpp
@@ -310,12 +310,15 @@ _take_serialized_message(
   auto info = static_cast<CustomSubscriberInfo *>(subscription->data);
   RCUTILS_CHECK_FOR_NULL_WITH_MSG(info, "custom subscriber info is null", return RMW_RET_ERROR);
 
-  eprosima::fastcdr::FastBuffer buffer;
-
+  // Point the deserialize path at the user's rmw_serialized_message_t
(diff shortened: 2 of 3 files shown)


rmw_fastrtps_shared_cpp/src/rmw_take.cpp
```diff
@@ -339,15 +342,10 @@ _take_serialized_message(
           continue;
         }
       }
-      auto buffer_size = static_cast<size_t>(buffer.getBufferSize());
-      if (serialized_message->buffer_capacity < buffer_size) {
-        auto ret = rmw_serialized_message_resize(serialized_message, buffer_size);
-        if (ret != RMW_RET_OK) {
-          return ret;           // Error message already set
```
fujitatomoya, line 346: before, it returns corresponding error code that is propagated for the failure here? but the new code conceals the reason because it just returns false without any information with it. i think this is minor but we would want to consider to add `RMW_SET_ERROR_MSG("failed to resize serialized message")` in FASTDDS_SERIALIZED_DATA_TYPE_RMW_SERIALIZED_MESSAGE case statement?

fujitatomoya:
@MiguelCompany can you or someone else take a look at this?

rmw_fastrtps_shared_cpp/src/rmw_take.cpp
MiguelCompany: Either that, or pass in `data.data` a pointer to a custom structure that has both the `serialized_message` and a `rmw_ret_t` field that is filled in the new case of the type support implementation

MiguelCompany (changes requested):
Agree with @fujitatomoya that the return code from `rmw_serialized_message_resize` shall be preserved.

Also please change the PR description to follow the standard PR template

facontidavide:
will do

rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/TypeSupport.hpp
```diff
@@ -41,7 +41,12 @@ enum SerializedDataType
 {
   FASTDDS_SERIALIZED_DATA_TYPE_CDR_BUFFER,
   FASTDDS_SERIALIZED_DATA_TYPE_DYNAMIC_MESSAGE,
-  FASTDDS_SERIALIZED_DATA_TYPE_ROS_MESSAGE
+  FASTDDS_SERIALIZED_DATA_TYPE_ROS_MESSAGE,
+  // `data` points to a `rmw_serialized_message_t` (aka
+  // `rcutils_uint8_array_t`). The CDR payload is written directly into that
+  // buffer on deserialize, avoiding an intermediate FastBuffer and the
+  // follow-up memcpy in `_take_serialized_message`.
```
facontidavide, line 48: is this comment too verbose or useful to keep?

MiguelCompany: I would put similar comments in the new structure I suggested in https://github.com/ros2/rmw_fastrtps/pull/877#discussion_r3085643222

PR #879, opened 2026-04-30
Fix UB in accessing the keys

clalancette:
## Description

This fixes two different problems with respect to keys:

* Make sure to do key.resize()
    
Doing a key.reserve() allocates memory, but still
leaves the size at 0.  Thus an access with the []
operator is UB.  This came up on Ubuntu 26.04 because
it is more strict about UB.
    
Fix it by using resize() instead, which does increase
the size and thus the [] operator is no longer UB.

* rmw_fastrtps_dynamic_cpp: Fix a UB in key support.
    
The problem here is that we were always accessing 16 bytes
of data from the key buffer, even when it was shorter than
that.  What we do instead is to always set the ihandle->value
to all zeros, then only copy in the number of key_buffer
bytes we have.  This matches what rmw_fastrtps_cpp does,
and fixes a crash in tests on Ubuntu 26.04

This should be merged at approximately the same time as https://github.com/ros2/rmw_fastrtps/pull/880 , https://github.com/ros2/rmw_fastrtps/pull/881 , and https://github.com/ros2/system_tests/pull/592

No.

Yes, Claude Opus 4.7

### Additional Information

This should be backported to Lyrical and Kilted.

### DIFF
--- a/rmw_fastrtps_cpp/src/type_support_common.cpp
+++ b/rmw_fastrtps_cpp/src/type_support_common.cpp
@@ -179,7 +179,7 @@ bool TypeSupport::get_key_hash_from_ros_message(
 
   const size_t max_serialized_key_length = 16;
 
-  auto ser_length = ser.get_serialized_data_length();
+  const auto ser_length = ser.get_serialized_data_length();
 
   // check for md5
   if (force_md5 || key_max_serialized_size_ > max_serialized_key_length) {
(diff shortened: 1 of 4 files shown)


MiguelCompany (approved):
LGTM

clalancette:
Here is CI with this and all of the other PRs listed here:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28895/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21695/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8758/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27770/)

fujitatomoya (approved):
lgtm with green CI.

clalancette:
This one has ACKs and passes CI, so I'm going to merge it and backport it.

clalancette:
@Mergifyio backport lyrical kilted

PR #880, opened 2026-04-30
Change the buffer-aware BUFBE: -> bufbe.

clalancette:
## Description

The buffer-aware work appends a BUFBE: sentinel to the publisher's user_data QoS field.  But parse_key_value does not support colons in keys, so all type hashes for buffers using this would be INVALID.

Fix this by dropping the colon from the sentinel, which makes it pass parse_key_value and keep type hashes working.

Also add in a regression test here.

This should be merged at approximately the same time as https://github.com/ros2/rmw_fastrtps/pull/879 , https://github.com/ros2/rmw_fastrtps/pull/881 , and https://github.com/ros2/system_tests/pull/592

No.

Yes, Claude Opus 4.7

### Additional Information

This should be backported to Lyrical.

### DIFF
--- a/rmw_fastrtps_shared_cpp/src/qos.cpp
+++ b/rmw_fastrtps_shared_cpp/src/qos.cpp
@@ -30,6 +30,7 @@
 #include "fastdds/dds/topic/qos/TopicQos.hpp"
 
 #include "rmw/error_handling.h"
+#include "rmw/impl/cpp/key_value.hpp"
 
 #include "rmw_dds_common/qos.hpp"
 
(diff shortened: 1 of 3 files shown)


clalancette:
Here is CI with this and all of the other PRs listed here:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28895/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21695/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8758/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27770/)

fujitatomoya (commented):
the PR description says "no user-facing behavior change" and "should be backported to Lyrical." If anyone has used buffer-aware code based on the unpatched form (are we expecting this fix is going to be merged before release?), this rename is a wire break for them? i guess that is probably not a great idea?

clalancette:
> the PR description says "no user-facing behavior change" and "should be backported to Lyrical." If anyone has used buffer-aware code based on the unpatched form (are we expecting this fix is going to be merged before release?), this rename is a wire break for them? i guess that is probably not a great idea?

That's a good point, but in my opinion, since Lyrical isn't released yet, this is fine to do.  Once Lyrical ships, then I think it would be a breaking change and we'd have to figure something else out.  But I'm open to other opinions.

MiguelCompany (approved):
LGTM

clalancette:
Here's another build with the updates in place:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28923/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21722/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8785/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27817/)

clalancette:
See https://github.com/ros2/system_tests/pull/592#issuecomment-4382327661

clalancette:
@Mergifyio backport lyrical

PR #904, opened 2026-08-11
Reuse FastBuffer managed buffer for each endpoint.

### LINKED ISSUE #903: Remove an extra memset for buffer_data for cdr serialization
### Description

When I was experimenting my custom rosidl::Buffer backend with `rmw_fastrtps`, I realized that the publish/subscribe timimg gaps seems to be somewhat proportional to its buffer size, while the buffer sits in shared memory, despite that the actual wire message sizes are almost the same.

### Motivation

I'm trying to experiment a custom `rosidl::Buffer` backend with shared memory fashion in Lyrical distribution. When that rosidl buffer path is active, the actual wire message is almost constant, regardless of its actual buffer size, since they are managed in different way like `cuda_buffer_backend_msgs` in https://github.com/ros2/rosidl_buffer_backends.

### Design / Implementation Considerations

`rmw_publisher` allocates temporary buffer with `std::vector<uint8_t> buffer_data(buffer_size)`, and immediately pass it to fast_buffer as follows. The buffer is for writing through Cdr serialization, hence those memory areas can be uninitialized.

```c++
    uint32_t serialized_size = callbacks->get_serialized_size(ros_message);
    size_t buffer_size = serialized_size + 4;  // +4 for CDR encapsulation header
    std::vector<uint8_t> buffer_data(buffer_size);

    eprosima::fastcdr::FastBuffer fast_buffer(
      reinterpret_cast<char *>(buffer_data.data()), buffer_size);
    eprosima::fastcdr::Cdr ser(
      fast_buffer, eprosima::fastcdr::Cdr::DEFAULT_ENDIAN,
      eprosima::fastcdr::CdrVersion::XCDRv1);
```

Since the current implementation allocates with std::vector, it is unavoidable to leave the buffer uninitialized. This seems to introduce an extra memset in its initialization.

## Possible workaround

One way to leave the buffer uninitialized is to use `unique_ptr<uint8_t[]>` instead of std::vector, like

```diff
diff --git a/rmw_fastrtps_cpp/src/rmw_publish.cpp b/rmw_fastrtps_cpp/src/rmw_publish.cpp
index 8cf7574..3570e83 100644
--- a/rmw_fastrtps_cpp/src/rmw_publish.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_publish.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
 #include <vector>
 
 #include "fastcdr/Cdr.h"
@@ -174,10 +175,10 @@ publish_to_buffer_endpoints(
   for (const auto & endpoint : state.endpoints) {
     uint32_t serialized_size = callbacks->get_serialized_size(ros_message);
     size_t buffer_size = serialized_size + 4;  // +4 for CDR encapsulation header
-    std::vector<uint8_t> buffer_data(buffer_size);
+    std::unique_ptr<uint8_t[]> buffer_data(new uint8_t[buffer_size]);
 
     eprosima::fastcdr::FastBuffer fast_buffer(
-      reinterpret_cast<char *>(buffer_data.data()), buffer_size);
+      reinterpret_cast<char *>(buffer_data.get()), buffer_size);
     eprosima::fastcdr::Cdr ser(
       fast_buffer, eprosima::fastcdr::Cdr::DEFAULT_ENDIAN,
       eprosima::fastcdr::CdrVersion::XCDRv1);
```

### Additional Information

Two additional information:

1. allocation buffer size.

Ideally, we can make the buffer size small since rosidl::Buffer's descriptor is usually smaller than its actual payload. Currently, the buffer size is calculated based on the `rosidl::Buffer` size, but what to be sent is its descriptor only. 

2. Effect to the ordinal publish path

My original motivation was to improve `rosidl::Buffer` specialized path performance, but it looks that the above patch also contribute reduce overheads, especially for large message. Note that when the buffer size is smaller than several MiB, the gap was almost negligible.

3. Effect on the CUDA backend buffer

For CUDA backend, users may send large size of data, so it would be nice to avoid those extra allocation/initialization in the publisher side.

dskkato:
## Description

Previously, heap allocation happend for each endpoint with std::vector<uint8_t>. This introduced extra memset via its constructor.

For recent ROS 2, where rosidl::Buffer was introduced, sometimes the actual payload size and message size may not match because only buffer descriptors may be sent. In such a case, now `get_serialized_size()` seems to be unreliable, and hard to pre-allocate the buffer in that path.

Fixes #903

Possibly yes.

Previously, it is guaranteed that the heap allocation only happens once per endpoint, through externally managed buffer (allocated with `std::vector<uint8_t>`). Now the number of allocations depends on combinations of buffer's strategies, and Cdr's usage. So, in small message cases, this patch will introduce an extra overhead of buffer reallocation.

For rosidl::Buffer usage, this patch will dramatically reduce extra allocation and zero-initialization when only the optimized message paths exist. 

I used CodeX to investigate the cause analysis, and FastBuffer/Cdr behavior.

### Additional Information

This patch is trying to resolve two issues at same time:

1. Reducing extra memory allocation when publishing rosidl::Buffer via optimized buf (carries only a descriptor) 
2. Removing extra memset caused by `std::vector<uint8_t>` constructor.

What I originally found is the second point, so if we wish more conservative way, it is possible to reserve the `fast_buffer` as previouslly. Still, I believe it is better to reuse that buffer accross endpoints, in my perspective...

Since this is my first PR to `rmw_fastrtps` repository, let me know if I missed anything.

### DIFF
--- a/rmw_fastrtps_cpp/src/rmw_publish.cpp
+++ b/rmw_fastrtps_cpp/src/rmw_publish.cpp
@@ -171,13 +171,19 @@ publish_to_buffer_endpoints(
   }
 
   // Publish to per-subscriber peer-to-peer endpoints for non-CPU subscribers.
-  for (const auto & endpoint : state.endpoints) {
-    uint32_t serialized_size = callbacks->get_serialized_size(ros_message);
-    size_t buffer_size = serialized_size + 4;  // +4 for CDR encapsulation header
-    std::vector<uint8_t> buffer_data(buffer_size);
 
-    eprosima::fastcdr::FastBuffer fast_buffer(
-      reinterpret_cast<char *>(buffer_data.data()), buffer_size);
+  // Intentionally reuse an internally managed FastBuffer across endpoints. Each Cdr
+  // starts at the beginning and owns its cursor, and `write_w_timestamp()` copies
+  // the data before returning, so reuse is safe.
+  //
+  // Preallocating with `get_serialized_size()` is not recommended: for
+  // `rosidl::Buffer`, it may report the full CPU fallback size even when only a
+  // small descriptor is sent.
+  //
+  // The buffer grows automatically and lazily without zero-initialization.
+  eprosima::fastcdr::FastBuffer fast_buffer;
+
+  for (const auto & endpoint : state.endpoints) {
     eprosima::fastcdr::Cdr ser(
       fast_buffer, eprosima::fastcdr::Cdr::DEFAULT_ENDIAN,
       eprosima::fastcdr::CdrVersion::XCDRv1);


dskkato:
I shared a result of this patch here for a reference.

https://discourse.openrobotics.org/t/has-a-cpu-shared-memory-backend-for-rosidl-buffer-been-explored/57240/7

nvcyc (approved):
Thanks for the PR!
The fix looks good to me.

I also recently found the same allocation overhead issue from some benchmarks and I've verified that this PR does eliminate the unnecessary memory allocation overhead that grows with the underlying data size.

fujitatomoya (approved):
lgtm

fujitatomoya:
Pulls: ros2/rmw_fastrtps#904
Gist: https://gist.githubusercontent.com/fujitatomoya/d341bef8572ed8c44b839b419d386a18/raw/9283f9e0da31b323575cde057103abdaaa5b05a3/ros2.repos
BUILD args:  --packages-above-and-dependencies rmw_fastrtps_cpp
TEST args:  --packages-above rmw_fastrtps_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20164
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30235/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22929/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10045/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29097/)

fujitatomoya:
@MiguelCompany what do you think about this? i went ahead to start the CI.

nvcyc:
Note that this PR should also be backported to Lyrical once merged.

hidmic (approved):
LGTM, though FWIW I think that eventually the validity of `get_serialized_size()` should be restored. Backends know what the descriptor size is, and the ROS message is available in this scope, so there is plenty of opportunity to bubble up that information.

dskkato:
Thanks, I agree. Note that the current buffer size comes from [`get_buffer_serialized_size`](https://github.com/ros2/rosidl_typesupport_fastrtps/blob/b71af87201de38a54a3cef8e552ca089ccba08fc/rosidl_typesupport_fastrtps_cpp/include/rosidl_typesupport_fastrtps_cpp/buffer_serialization.hpp#L81-L124)

The current `get_serialized_size()` path cannot determine the exact size for the buffer-aware path because it does not receive the endpoint information or the buffer serialization context. Depending on the endpoint, serialization may use either the backend descriptor or the CPU fallback.

I think this should be addressed in a follow-up by adding an endpoint-aware size calculation through the above `rosidl_typesupport_fastrtps` and the generated type support callbacks. It should use the same backend selection as `cdr_serialize_with_endpoint()`, returning the descriptor size when the backend is accepted and the CPU size otherwise.

This PR uses the managed `FastBuffer` so that allocation does not depend on the currently conservative size estimate. I will track the endpoint-aware `get_serialized_size()` improvement separately.

fujitatomoya:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23053/)

fujitatomoya:
@dskkato i just restarted aarch64 CI. when it returs green, this is good to merge.

MiguelCompany (approved):
LGTM, but looking forward to the follow-up adding an endpoint-aware size calculation

fujitatomoya:
@Mergifyio backport lyrical

dskkato:
I looked into whether we could restore the previous behavior of `get_serialized_size()`, but found that it is not straightforward.

The difficulty is that the actual wire representation for an endpoint is not determined until [`create_descriptor_with_endpoint()` ](https://github.com/ros2/rosidl/blob/790b342041e7d2f8419a1f596083b94c438ba318/rosidl_buffer_backend/include/rosidl_buffer_backend/buffer_backend.hpp#L67-L79) is called. If it returns a descriptor, the descriptor path is used; otherwise serialization falls back to the CPU representation.

[`create_descriptor_with_endpoint()`](https://github.com/ros2/rosidl/blob/790b342041e7d2f8419a1f596083b94c438ba318/rosidl_buffer_backend/include/rosidl_buffer_backend/buffer_backend.hpp#L67-L79) is not just a capability query. The API allows it to perform endpoint-specific setup or bookkeeping, such as IPC handle setup or memory pinning, and does not require repeated calls to be side-effect-free or idempotent (though it should be so). Therefore, an endpoint-aware size calculation cannot safely call it once to determine the descriptor size and then call it again during serialization.

A separate side-effect-free capability/size query could make endpoint-aware size calculation possible, but it would need a consistency contract with descriptor creation and still handle runtime descriptor-creation failures.

PR #906, opened 2026-08-13
Use fixed version for rmw_fastrtps_shared_cpp dependency

### LINKED ISSUE #905: Lyrical: 9.4.9 changed rmw_fastrtps_shared_cpp ABI (CustomSubscriberInfo); unversioned deb deps allow a mixed 9.4.9/9.4.8 install that crashes all clients
## Bug report

**Required Info:**

- Operating System: Ubuntu 26.04 (Resolute), official `ros:lyrical-ros-core` / `ros:lyrical-ros-base` Docker images
- Installation type: binaries (packages.ros.org)
- Version or commit hash: `ros-lyrical-rmw-fastrtps-cpp` 9.4.9-1resolute.20260730 mixed with `ros-lyrical-rmw-fastrtps-shared-cpp` 9.4.8-1resolute.20260605
- DDS implementation: Fast DDS (default `rmw_fastrtps_cpp`)
- Client library (if applicable): reproduced with both rclcpp (32.0.x) and rclpy
- Architecture: arm64 confirmed (SIGBUS in rclpy, SIGSEGV in rclcpp during lifecycle activation); amd64 untested, but the mechanism is arch-independent

## Summary

The Lyrical 9.4.9 patch release changed the layout of `CustomSubscriberInfo` in a `rmw_fastrtps_shared_cpp` public header (#902 removed the `buffer_data_guard_` member, shifting the offsets of every member after it). `rmw_fastrtps_cpp` 9.4.9 is compiled against the new layout, but the deb only declares an **unversioned** dependency on `ros-lyrical-rmw-fastrtps-shared-cpp`, so apt is free to install 9.4.9 `rmw_fastrtps_cpp` on top of a 9.4.8 `rmw_fastrtps_shared_cpp` that is already present. The resulting mixed-ABI install crashes every Fast DDS client at startup: `ros2` CLI (rclpy) dies with SIGBUS, rclcpp lifecycle nodes segfault during activation.

This currently happens with the *default* docker + apt workflow: all `ros:lyrical-*` Docker Hub images were last built 2026-08-04 (before the 2026-08-07 sync) and ship the 9.4.8 pair, so any Dockerfile that does `apt-get install ros-lyrical-<pkg>` for a package that pulls in `rmw-fastrtps-cpp` — or lists `ros-lyrical-rmw-fastrtps-cpp` explicitly, without also naming `rmw-fastrtps-shared-cpp` — silently produces the broken mix.

## Steps to reproduce issue

```bash
docker run --rm ros:lyrical-ros-core bash -c '
  apt-get update &&
  apt-get install -y ros-lyrical-rmw-fastrtps-cpp &&
  dpkg -l | grep rmw-fastrtps &&
  source /opt/ros/lyrical/setup.bash &&
  ros2 topic list --no-daemon'
```

`dpkg -l` shows the mix apt produced:

```
ros-lyrical-rmw-fastrtps-cpp         9.4.9-1resolute.20260730.175440
ros-lyrical-rmw-fastrtps-shared-cpp  9.4.8-1resolute.20260605.200401
```

## Expected behavior

Either of:
- patch releases into a released (LTS) distro keep the `rmw_fastrtps_shared_cpp` ABI stable, or
- the generated debs carry a versioned dependency so `rmw-fastrtps-cpp` and `rmw-fastrtps-shared-cpp` upgrade in lock-step.

## Actual behavior

`ros2 topic list --no-daemon` dies with `Bus error` (SIGBUS) on arm64. An rclcpp `LifecycleNode` under the same install segfaults (SIGSEGV, exit -11) during lifecycle activation.

## Additional information

- A **coherent** 9.4.9 install is fine: upgrading `ros-lyrical-rmw-fastrtps-shared-cpp` to 9.4.9 in the broken container immediately fixes both crash modes, as does a full `apt-get dist-upgrade`. So this is not a code regression in 9.4.9 itself — it is the ABI change + unversioned dependency combination.
- Bisected from a real deployment: images rebuilt after the 2026-08-07 sync crashed while identical images built from the 2026-06 syncs worked, with the diff pointing at the rmw_fastrtps 9.4.8→9.4.9 bump.
- Workaround for anyone hitting this: run `apt-get dist-upgrade -y` after `apt-get update` in Dockerfiles based on `ros:lyrical-*` images (or explicitly install a matching `ros-lyrical-rmw-fastrtps-shared-cpp`).

MiguelCompany:
## Description

Package `rmw_fastrtps_shared_cpp` exists only to contain code that is shared by the other two packages in this repository.
It is not meant to be used directly by a user, and we sometimes allow ABI breaks within a distribution, since we assume that the three packages will be updated at the same time.

This makes the coupling explicit by requiring the versions to be in sync.

Fixes #905

No.

### Additional Information

This targets lyrical in order to fix #905.
We shall do similar PRs for other branches.

### DIFF
--- a/rmw_fastrtps_cpp/package.xml
+++ b/rmw_fastrtps_cpp/package.xml
@@ -17,13 +17,14 @@
 
   <buildtool_depend>ament_cmake_ros_core</buildtool_depend>
 
+  <depend version_eq="9.4.9">rmw_fastrtps_shared_cpp</depend>
+
   <build_depend>fastcdr</build_depend>
   <build_depend>fastdds</build_depend>
   <build_depend>rcpputils</build_depend>
   <build_depend>rcutils</build_depend>
   <build_depend>rmw</build_depend>
   <build_depend>rmw_dds_common</build_depend>
-  <build_depend>rmw_fastrtps_shared_cpp</build_depend>
   <build_depend>rosidl_dynamic_typesupport</build_depend>
   <build_depend>rosidl_dynamic_typesupport_fastrtps</build_depend>
   <build_depend>rosidl_runtime_c</build_depend>
@@ -37,7 +38,6 @@
   <build_export_depend>fastdds</build_export_depend>
   <build_export_depend>rcutils</build_export_depend>
   <build_export_depend>rmw_dds_common</build_export_depend>
-  <build_export_depend>rmw_fastrtps_shared_cpp</build_export_depend>
   <build_export_depend>rmw</build_export_depend>
   <build_export_depend>rosidl_dynamic_typesupport</build_export_depend>
   <build_export_depend>rosidl_runtime_c</build_export_depend>
@@ -49,7 +49,6 @@
   <exec_depend>rcpputils</exec_depend>
   <exec_depend>rcutils</exec_depend>
   <exec_depend>rmw</exec_depend>
-  <exec_depend>rmw_fastrtps_shared_cpp</exec_depend>
   <exec_depend>rosidl_dynamic_typesupport</exec_depend>
   <exec_depend>rosidl_dynamic_typesupport_fastrtps</exec_depend>
   <exec_depend>tracetools</exec_depend>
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
i think this is easiest way to address the situation right away.

